Issue #125, opened 2025-06-03
Add retry logic


Adam-Hammo:
At the moment the batch enhancement process could fail on a single network blip at multiple points. We should add various levels of retry logic (with backoffs, jitter etc), in a way that's easily reusable in other parts of the repo.

Candidates:
- Blob storage upload/download/signing
- Robot communication
- Tasks as a whole

mootpointer:
@Adam-Hammo Have we resolved these concerns with the switch to polling robots (#270)?

Adam-Hammo:
Yeah, I think we can close this. Robot communication and tasks are both more resilient now, and the point about blobs would be further mitigated by #318 if/when we get to it.

PR #134, opened 2025-06-06
Search/repo es integration

Adam-Hammo:
Summary:
- Implements three interactions with Elasticsearch:
  - `index_references()` - bulk generator-based method of indexing. Called after batch imports and batch enhancement results.
  - `index_reference()` - single simple indexing of a reference. Called after single enhancement results.
  - `POST /reference/index/rebuild` - deletes, recreates, repopulates the index in elasticsearch. I see this as a ghetto alembic and/or data repair while we're iterating. Obviously not a long-term solution to elasticsearch management.
- Mimicked the SQL persistence setup. This is stubbish as elasticsearch doesn't have transactions but creates a reasonable perimeter for us to implement some [retry logic](https://github.com/destiny-evidence/destiny-repository/issues/125) or other data controls in the future.
- Used Elasticsearch DSL to define the index - similar to the SQLAlchemy setup. Not fully convinced we'll be able to go all the way here, but we can pivot to manage indexes elsewhere and still use DSL AsyncDocuments for our in/out. Works to the extent of Andrew's index for annotation aggregations.


danielbreves (approved):
LGTM!

PR #139, opened 2025-06-16
98 send newly imported records to a set of robots

Adam-Hammo:
EQWD implementation of default enhancements. TL;DR creates a batch enhancement request (identical to how an actor would) on all imported references based on a boolean attribute on the robot registration.

Suitable for a non-dependent set of enhancements to be automatically requested on import of any reference.

The import process and the enhancement process as two entirely different entities, implying:
- we consider an import to be successful even if the auto enhancements failed
- in the period between an import and a batch enhancement, the references will be stored and viewable in our persistence in that "incomplete" state

### DIFF
--- a/app/domain/references/tasks.py
+++ b/app/domain/references/tasks.py
@@ -186,7 +186,7 @@ async def request_default_enhancements(
     """
     Request default enhancements for a set of references.
 
-    Technically a task distributor, not a task - may live in a higher layer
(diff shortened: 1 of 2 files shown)


Adam-Hammo:
Leaving this open in case we need it, but I expect #141 to supersede this next week.

Adam-Hammo:
Superseded by https://github.com/destiny-evidence/destiny-repository/pull/164

PR #150, opened 2025-06-18
QOL real data fixes

Adam-Hammo:


app/domain/references/services/ingestion_service.py
```diff
@@ -38,21 +37,10 @@ async def fetch_collided_identifiers(
...
-                        identifier.other_identifier_name,
-                    )
-                )
-            except NotFoundError:
-                pass
-            else:
-                collided.append(existing_identifier)
-        return collided
+        if not identifiers:
+            return []
```
danielbreves: Looks like `get_by_identifiers` already does this?

PR #153, opened 2025-06-19
Handle transient issues more resiliently

Adam-Hammo:
Adds an auto-retry function to import batches on transient issues (inconsistent data being uploaded in parallel, network problems), up to a configurable number of retries (default 3).


app/domain/imports/service.py
```diff
@@ -156,17 +160,28 @@ async def process_import_batch_file(
...
+        except IntegrityError as exc:
+            # This handles the case where files loaded in parallel cause a conflict at
+            # the persistence layer.
+            logger.warning(
+                "Integrity error processing batch, likely caused by inconsistent state"
+                " being loaded in parallel. Will retry.",
+                extra={"import_batch_id": import_batch.id, "error": str(exc)},
+            )
+            return ImportBatchStatus.RETRYING
+        except httpx.NetworkError as exc:
```
Adam-Hammo, line 172: We can get a bit fancier with this later: https://github.com/destiny-evidence/destiny-repository/issues/125

app/domain/imports/service.py
```diff
@@ -156,17 +160,28 @@ async def process_import_batch_file(
...
+            # This handles retryable network errors like connection refused,
+            # connection reset by peer, timeouts, etc.
+            logger.warning(
+                "Network error processing batch. Will retry.",
```
jack-o-the-green, line 176: This will be an incorrect log line on the third attempt I believe - perhaps a `retries remaining` in the log?

Adam-Hammo: Nice catch!

PR #154, opened 2025-06-19
Prepare destiny-sdk for publishing

jack-o-the-green:
* Updated SDK README
* Moved SDK test into the SDK 
* Ensured all necessary SDK dependencies are in the SDK .toml

### DIFF
--- a/libs/sdk/README.md
+++ b/libs/sdk/README.md
@@ -67,3 +67,13 @@ poetry publish --repository testpypi
 poetry config pypi-token.pypi [YOUR_PYPI_TOKEN]
 poetry publish
 ```
+
+### Versioning
+
+Follow the [semver](https://semver.org/) guidelines for versioning, tldr;
+
+Given a version number `MAJOR.MINOR.PATCH`, increment the:
+
+- `MAJOR` version when you make incompatible API change
+- `MINOR` version when you add functionality in a backward compatible manner
+- `PATCH` version when you make backward compatible bug fixes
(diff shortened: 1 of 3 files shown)


libs/sdk/README.md
```diff
@@ -14,16 +44,26 @@ or replace the dependency in `pyproject.toml` with
...
 
-```toml
-destiny-sdk = {git = "ssh://git@github.com:destiny-evidence/destiny-repository.git", subdirectory = "libs/sdk"}
+```sh
+poetry config repositories.testpypi https://test.pypi.org/legacy/
+poetry config pypi-token.testpypi [YOUR_TESTPYPI_TOKEN]
+poetry publish --repository testpypi
+```
+
+## Publishing
```
Adam-Hammo, line 64: Should we add any advice here about versioning?

Adam-Hammo (commented):
I think there might be a couple more touch points eg `sdk.rst` and the top-level README

libs/sdk/README.md
jack-o-the-green: I'm not sure _we_ have advice about versioning at this stage - could link to the regular semver best practices https://semver.org/?

jack-o-the-green:
> I think there might be a couple more touch points eg `sdk.rst` and the top-level README

Ooh good catch, I forgot about the other sdk docs

libs/sdk/README.md
jack-o-the-green: Have added in the semver advice

PR #155, opened 2025-06-19
let uow skip committing if session already rolled back, catch correct exception

Adam-Hammo:
If we catch an integrity error or similar, SQLAlchemy rolls back the session, but the UOW handler still tries to commit it - until now!


jack-o-the-green (approved):
LGTM

PR #156, opened 2025-06-19
Hotfix/terminal retrying state

Adam-Hammo:
- Set state to Failed when retries are all out.
- Move retries to task queue - better implementation and also hopefully more resilient to transient failure with a break in between runs.
- Catch and retry on deadlock.


jack-o-the-green (approved):
LGTM - ship it

PR #159, opened 2025-06-20
test serializability on all domain models

Adam-Hammo:
Quick serverside checking of SDK->domain conversions to catch edge cases that might cause serialising problems. Also see https://covidence.slack.com/archives/C0810ARPCRG/p1750382080141059.

### DIFF
--- a/app/domain/imports/models/models.py
+++ b/app/domain/imports/models/models.py
@@ -154,9 +154,12 @@ The number of references expected to be included in this import.
     async def from_sdk(cls, data: destiny_sdk.imports.ImportRecordIn) -> Self:
         """Create an ImportRecord from the SDK input model."""
         try:
-            return cls.model_validate(data.model_dump())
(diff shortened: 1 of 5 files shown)


app/domain/imports/models/models.py
```diff
@@ -154,9 +154,12 @@ class ImportRecord(DomainBaseModel, SQLAttributeMixin):
     async def from_sdk(cls, data: destiny_sdk.imports.ImportRecordIn) -> Self:
```
danielbreves, line 154: I wonder if we could implement a mixin for these?

Adam-Hammo: Yeah I did think about that - an abstract mixin would be nice but pydantic doesn't play so nice with abstract classes. I think the actual function bodies, even though they're relatively similar at the moment, will possible diverge over time.

PR #168, opened 2025-06-27
Resilience improvements

Adam-Hammo:
- Update `batch_enhancement_status` enum (alembic doesn't automatically detect enum changes)
- Pin `aiohttp` version per https://github.com/Azure/azure-sdk-for-python/issues/41363 - wasn't actually causing a problem (yet) but logs were very noisy
- Add automatic retries on timeouts to Elasticsearch. This happened to 3/70 batches of 10,000 documents, so I don't expect we need to increase the timeout interval itself.
- Attempt to improve performance of first DB call after replica boot, I was getting sporadic failures every now and then. Will verify this one straight after merging.


jack-o-the-green (approved):
Looking good, would like to know how much that cold boot change helps, it looks like it should completely fix the issue.

Adam-Hammo:
> Looking good, would like to know how much that cold boot change helps, it looks like it should completely fix the issue.

Anecdotally, first call post-boot was much better.

Issue #122, opened 2025-06-03
Consider a distinct SDK <> Domain layer


Adam-Hammo:
Currently we decouple the SDK and domain models with `to_sdk()` and `from_sdk()` methods on the domain models. 
This works to a point but as we increase complexity it necessitates that the domain models require a fair amount of SDK knowledge. For instance, SDK models with URLs (eg `robots.BatchRobotRequest`) now require knowledge on how to convert a domain model's file pointer to a SDK model's signed url.

We should implement something similar to an [anti-corruption layer](https://learn.microsoft.com/en-us/azure/architecture/patterns/anti-corruption-layer) that removes this concern from the domain models. It should be used at the router level.

Adam-Hammo:
Noting from https://github.com/destiny-evidence/destiny-repository/pull/188#discussion_r2203583006 that we'll standardise the thrown exceptions here as well.

PR #188, opened 2025-07-09
added ACL for robot domain

Adam-Hammo:
Implementation of https://github.com/destiny-evidence/destiny-repository/issues/122 in the robot domain.

It seems relatively pointless here, but is the simplest example to sort out how we want this pattern to look.

### DIFF
--- a/app/domain/robots/services/anti_corruption_service.py
+++ b/app/domain/robots/services/anti_corruption_service.py
@@ -1,0 +1,45 @@
+"""Anti-corruption service for robots domain."""
+
+import destiny_sdk
+from pydantic import ValidationError
+
+from app.core.exceptions import SDKToDomainError
+from app.domain.robots.models.models import Robot
+from app.domain.service import AntiCorruptionService
+
+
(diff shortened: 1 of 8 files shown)


app/domain/robots/services/anti_corruption_service.py
```diff
@@ -0,0 +1,45 @@
...
+            return robot
+
+    def robot_to_sdk(self, robot: Robot) -> destiny_sdk.robots.Robot:
+        """Convert the robot to a Robot SDK model."""
+        try:
+            model = robot.model_dump()
+            model.pop("client_secret", None)
+            return destiny_sdk.robots.Robot.model_validate(model)
+        except ValidationError as exception:
+            raise SDKToDomainError(errors=exception.errors()) from exception
```
jack-o-the-green, line 33: This was probably my fault from a while back, but the `_to_sdk()` functions should raise a DomainToSDKError not a SDKToDomainError

Adam-Hammo: Good call, will include that across the board in the ACL changes.

PR #189, opened 2025-07-09
Add a DenseEmbeddings enhancement to the SDK

mootpointer:
This allows for multiple embeddings to be created across a single enhancement, supporting document chunking or embedding fields separately.

### DIFF
--- a/libs/sdk/src/destiny_sdk/enhancements.py
+++ b/libs/sdk/src/destiny_sdk/enhancements.py
@@ -352,7 +352,7 @@ class DenseEmbeddingsEnhancement(BaseModel):
     enhancement_type: Literal[EnhancementType.DENSE_EMBEDDINGS] = (
         EnhancementType.DENSE_EMBEDDINGS
     )
-    embedding: list[DenseEmbedding] = Field(
+    embeddings: list[DenseEmbedding] = Field(
         description="A list of embeddings for the reference."
     )
     model_version: str = Field(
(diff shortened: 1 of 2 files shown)


libs/sdk/src/destiny_sdk/enhancements.py
```diff
@@ -303,12 +304,69 @@ class LocationEnhancement(BaseModel):
...
+        description="""
+An arbitrary identifier for the embedding to collect like embeddings across references
+""",
+    )
+    description: str | None = Field(
+        default=None,
+        description="""
+A description of the embedding, such as the part of the reference it represents.
+""",
+    )
```
Adam-Hammo, line 336: I'm not super familiar with embeddings but it feels like there's a level of structure we can add here. Something like a `source: [title,abstract,text,text_section]` enum and a `details: str`field.
Also it's inevitable but I think I'd love to avoid overloading `"identifier"` in the repo while we can.

libs/sdk/src/destiny_sdk/enhancements.py
```diff
@@ -303,12 +304,69 @@ class LocationEnhancement(BaseModel):
...
+    references. For example, if you have an embedding for the title of a reference,
+    you might use `title_embedding` as the identifier. You do not need to include model
+    details in the identifier, as these are captured in the `model_version` field of the
+    enhancement.
+    """
+
+    embedding: list[float] = Field(
+        description="A list of floats representing the dense embedding."
+    )
+    num_dimensions: int = Field(
```
Adam-Hammo, line 323: Could this just be
```
@property
def num_dimensions(self) -> int:
    return len(self.embedding)
```

Adam-Hammo: Otherwise we should maybe add a model validator

Issue #180, opened 2025-07-08
Database backups and recovery process


jack-o-the-green:
As we start to ingest data into destiny repository, it becomes important of us to preserve that data in the event of something going wrong with our production database. We need to set up backups and any other necessary resilience to allow us to restore our database from a recent back up in the instance of data corruption or other issues.

Scope 
* Set up database backups for at least the staging and production deployments.
* Document the process for restoring the database from a backup. This should include any elastic search concerns.

PR #196, opened 2025-07-15
#180 Database backups and recovery process

danielbreves:
See https://github.com/destiny-evidence/destiny-repository/issues/180


infra/app/main.tf
```diff
@@ -410,11 +434,23 @@ resource "elasticstack_elasticsearch_security_api_key" "read_only" {
...
+resource "elasticstack_elasticsearch_snapshot_lifecycle" "snapshots" {
+  name = "snapshot-policy"
+
+  # Every 30 minutes for production, once a day at 01:30 AM otherwise
+  schedule   = var.environment == "prod" ? "0 */30 * * * ?" : "0 30 1 * * ?"
+  repository = "found-snapshots" # Default Elastic Cloud repository
+
+  expire_after = "30d"
+  min_count    = 30
+  max_count    = 60
```
Adam-Hammo, line 455: What's the cost to keeping these snapshots? I feel we should up this if not too exorbitant. If for eg something happened on a Friday that we didn't pick up on until a Monday, the snapshots relevant to pre-incident recovery would be long gone.

infra/app/main.tf
```diff
@@ -239,6 +239,14 @@ resource "azurerm_postgresql_flexible_server" "this" {
   administrator_login           = var.admin_login
   administrator_password        = var.admin_password
   zone                          = "1"
+  backup_retention_days         = var.environment == "prod" ? 35 : 7
+
+  dynamic "high_availability" {
+    for_each = var.environment == "prod" ? [1] : []
```
Adam-Hammo, line 245: I love terraform.

infra/app/main.tf
```diff
@@ -253,6 +261,14 @@ resource "azurerm_postgresql_flexible_server" "this" {
 
   depends_on = [azurerm_private_dns_zone_virtual_network_link.db]
   tags       = local.minimum_resource_tags
+
+  # avoid migrating back to the primary availability zone after failover
```
Adam-Hammo, line 265: Domain Q: would we eventually migrate back manually, or just treat the failed over zone as the new primary?

Adam-Hammo (commented):
For the avoidance of doubt - the README describes the current setup and we don't need any infra changes, correct?

infra/app/main.tf
danielbreves: Thanks for catching this. It was left over template code.

danielbreves:
> For the avoidance of doubt - the README describes the current database setup and we don't need any infra changes, correct?

@Adam-Hammo the README describes the setup with the infra changes in this PR.

infra/app/main.tf
danielbreves: On that topic though, I just noticed we can enable zone redundancy for the container app: https://github.com/futureevidence/terraform-azure-container-app/pull/9. The only problem is that it forces the container app to be recreated, which could cause a bit of pain I think.

PR #204, opened 2025-07-16
Remove redundant async on translation methods

Adam-Hammo:
Sidecar PR to #192 #193 #194.

The above PRs remove `async` from translation functions unless there's an obvious performance improvement (network call). This PR performs that same change across translation functions not touched by the anti-corruption layer.


jack-o-the-green (approved):
LGTM, nice 👌

PR #207, opened 2025-07-21
Enhancement requests v1

Adam-Hammo:
Splits out enhancement requests to new top-resource router(s).

I'm also considering splitting out the SDK models similarly - right now they're all thrown under `robots`. 

E2E tests should work after updating SDK version on the toy robot - holding off this until I'm sure of scope of SDK changs in the underlying epic.


app/domain/references/routes.py
```diff
@@ -147,13 +161,30 @@ async def robot_auth(
     )(request)
 
 
-router = APIRouter(prefix="/references", tags=["references"])
-robot_router = APIRouter(
-    prefix="/robot", tags=["robots"], dependencies=[Depends(robot_auth)]
+reference_router = APIRouter(prefix="/references", tags=["references"])
```
jack-o-the-green, line 164: I think the `reference.writer` auth is no longer being used, so this router can depend on the `reference_reader_auth` and we can remove the `reference.writer` scope entirely.

jack-o-the-green (approved):
LGTM - one bit of suggested clean-up

PR #208, opened 2025-07-21
implemented HybridAuth AuthMethod

Adam-Hammo:
Adds a new HybridAuth AuthMethod (with some small refactoring to generalise our repo-side AuthMethods).

This can be used on endpoints which require both robot HMAC and managed oauth authentication.


tests/core/test_hybrid_auth.py
Adam-Hammo, line 1: I'll refactor this to test a simpler endpoint, the data setup is a bit too intensive and domain-driven for a functional test like this.

app/core/auth.py
```diff
@@ -405,7 +408,11 @@ def __init__(self, get_client_secret: Callable[[UUID], Awaitable[str]]) -> None:
         """
         self.get_secret = get_client_secret
 
-    async def __call__(self, request: Request) -> bool:
+    async def __call__(
+        self,
+        request: Request,
+        credentials: HTTPAuthorizationCredentials | None = None,  # noqa: ARG002
```
jack-o-the-green, line 414: Are we passing credentials here but not using them - and then the HMAC auth will get `None` every time?

jack-o-the-green (approved):
LGTM

PR #210, opened 2025-07-22
added changelog documentation

Adam-Hammo:
Looks like this at `/redoc` (we should preference this as docs endpoint going forward):

image

### DIFF
--- a/app/api/root.py
+++ b/app/api/root.py
@@ -1,10 +1,12 @@
 """Root API router for the repository API."""
 
+import pathlib
 from collections.abc import Callable
 from contextlib import AbstractAsyncContextManager
 
 from fastapi import APIRouter, FastAPI
 from fastapi.middleware import Middleware
+from fastapi.middleware.cors import CORSMiddleware
 
 from app.api.exception_handlers import (
     es_malformed_exception_handler,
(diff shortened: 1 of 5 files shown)


app/api/root.py
```diff
@@ -48,8 +50,30 @@ def register_api(
...
+        description=pathlib.Path(
+            pathlib.Path(__file__).parent, "description.md"
+        ).read_text(encoding="utf-8"),
+        version="1.0.0",
         lifespan=lifespan,
-        middleware=[Middleware(LoggerMiddleware)],
+        middleware=[
+            Middleware(LoggerMiddleware),
+            Middleware(
+                CORSMiddleware,
```
Adam-Hammo, line 67: I did this CORS in an attempt to embed nicely in the other docs. CSS won today but figured I'd leave in docs-friendly CORS?

app/api/root.py
```diff
@@ -48,8 +50,30 @@ def register_api(
...
+                allow_origins=[
+                    "https://destiny-evidence.github.io",
+                    "null",
```
jack-o-the-green, line 70: What does having "null" here do? Couldn't find that one in the docs

Adam-Hammo: Which does raise the question - could someone then CORS the API from their own local file? I might just remove for now.

Adam-Hammo: Great callout!

PR #224, opened 2025-07-28
8 instrument the application with open telemetry

Adam-Hammo:
- Implements open telemetry infrastructure with a local explorer
- Adds custom taskiq spans to correlate app and task work
- Adds user context to FastAPI spans 

Pointed at detached branch for now as some of those refactors are incorporated, will collapse that in the coming days.


app/core/logger.py
```diff
@@ -38,6 +39,24 @@ def exception(self, message: str, **kwargs: dict) -> None:
         self.logger.exception(message, **kwargs)
 
 
+class TraceContextProcessor:
```
Adam-Hammo, line 42: This is just a first pass - I'll comprehensively address logging in another PR.

app/core/telemetry/taskiq.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+
+    def __init__(self) -> None:
+        """Initialize the middleware."""
+        super().__init__()
+        self._current_span: Span | None = None
+        self._token: context.Token[Context] | None = None
+
+    @staticmethod
+    async def kiq(
```
danielbreves, line 42: Any reason this needs to be within the middleware?

danielbreves: I'd lean towards making it a helper function outside the class, just to keep it to a single responsibility.

Adam-Hammo: Done!

app/tasks.py
```diff
@@ -18,10 +21,19 @@
...
+
 elif settings.env == "test":
     broker = InMemoryBroker()
 
+# Add OpenTelemetry middleware to all brokers
+if settings.otel_config and settings.otel_enabled:
+    configure_otel(
+        settings.otel_config, settings.app_name, settings.app_version, settings.env
+    )
+    broker.add_middlewares(TaskiqTracingMiddleware())
```
danielbreves, line 35: Will this potentially make it share the same instance for multiple workers? Just wondering if we need to worry about concurrency here.

Adam-Hammo: Great point - each worker will have its own instance but there's definitely a concurrency concern where multiple tasks run at once on the same worker. I'll throw some contextvars in and test it.

danielbreves (approved):

PR #233, opened 2025-07-30
added e2e profile to github action command

Adam-Hammo:


Adam-Hammo:
Going to address in #226 instead

PR #234, opened 2025-07-30
217: Use api scopes instead of roles for auth client

danielbreves:
According to https://learn.microsoft.com/en-us/entra/identity-platform/scenario-protected-web-api-app-registration:

> App roles cannot be consented to by a user (as they're used by an application that calls the web API on behalf of itself). A tenant administrator will need to consent to client applications of your web API exposing app roles.

An example of this is [Covidence authenticating against Study Data Service](https://github.com/covidence/study-data-service?tab=readme-ov-file#authentication).

So it seems we should use scopes If the client app calls APIs on behalf of a user. If we need different scopes for different types of users I think we may need different `azuread_application_registration`.

~DISCLAIMER: I haven't tested this, so it's all theory at this stage.~ It works! 🎉


infra/app/auth.tf
```diff
@@ -1,105 +1,132 @@
...
   display_name         = "Administrator"
   role_id              = random_uuid.administrator_role.result
   value                = "administrator"
 }
 
 resource "azuread_application_app_role" "importer" {
-  application_id       = azuread_application_registration.destiny_repository.id
-  allowed_member_types = ["User", "Application"]
+  application_id       = azuread_application.destiny_repository.id
+  allowed_member_types = ["Application"]
```
danielbreves, line 85: I may need to confirm how openalex_incremental_updater uses this.

app/api/auth.py
```diff
@@ -41,15 +41,23 @@
 security = HTTPBearer(auto_error=False)
 
 
-class AuthScopes(StrEnum):
-    """Enum describing the available auth scopes that we understand."""
+class AuthRoles(StrEnum):
```
jack-o-the-green, line 44: `nitpick`: I believe the majority of our enum names aren't pluralized - e.g. `ImportBatchStatus`, so this an below would ideally be `AuthRole` and `AuthScope`

app/api/auth.py
```diff
@@ -327,21 +338,32 @@ async def _get_microsoft_keys(self) -> Any:  # noqa: ANN401
...
-    def _require_scope(
-        self, required_scope: StrEnum, verified_claims: dict[str, Any]
-    ) -> bool:
-        if verified_claims.get("roles"):
-            for scope in verified_claims["roles"]:
-                if scope.lower() == required_scope.value.lower():
-                    return True
+    def _require_scope_or_role(self, verified_claims: dict[str, Any]) -> bool:
+        # Delegated (user) token: check scopes
+        if self.scope and verified_claims.get("scp"):
```
jack-o-the-green, line 343: Question for my understanding - is it possible that a jwt would contain _both_ scopes and roles? If that could be the case, I think we'd need some additional error handling around this, as it might not be appropriate to raise an `AuthError` on a missing scope if the equivalent role was present.

infra/app/auth.tf
```diff
@@ -1,102 +1,129 @@
...
+      admin_consent_display_name = "Robot Writer as user"
+      id                         = random_uuid.robot_writer_scope.result
+      type                       = "User"
+      value                      = "robot.writer.all"
+      user_consent_description   = "Allow you to register robots and rotate robot client secrets"
+      user_consent_display_name  = "Robot Writer"
+    }
+  }
+
+  lifecycle {
```
jack-o-the-green, line 73: It's possible to specify the `app_roles` and `identifier_uris` as part of this resource, which would mean we don't need to have the lifecycle ignore changes here. 

Had previously not had the api blocks so the decomposed resources felt a bit more elegant, but the api permission scopes don't appear to be available as decomposed resources, so I'd be in favour of combining everything now.

app/api/auth.py
danielbreves: The short answer is yes, it's technically possible for a jwt to contain both scopes and roles, but I don't think we need to support that.

The `scp` claim [is only included for user tokens](https://learn.microsoft.com/en-us/entra/identity-platform/access-token-claims-reference) and which case we want to verify scopes, not roles. I think the only case we could receive both roles and scopes is if the app registration client had requested both scopes and roles and the user was an admin (since roles require admin consent), but I don't think we'd want to do that.

jack-o-the-green (approved):
LGTM

PR #238, opened 2025-08-01
Instrument/custom

Adam-Hammo:
I found it hard to get an intuition about where to draw lines here - open to opinions, might be something we feel out over time. In general I focused on:
- breaking up large traces, we now have no spans with more than about 10 children (with the exception of O(n) looping)
  - for instance, giving each line on an ingestion its own trace
- adding context - IDs, line numbers for eg
  - did these in an automated fashion where I could, eg in the `persistence.add()` method
  - I didn't want to go too overboard here and add redundant information, so these are pretty lightweight for now. If in an investigation you need to consult the DB using the ID, fine - so long as we're not using that important information before we've saved it to the DB!
  - The sky is the limit and the skeleton is here to add context as we see fit.
- adding exception information

The example we went over on Thursday now looks like: https://ui.honeycomb.io/destiny-evidence/environments/test/datasets/destiny-worker-e2e/result/s4JAszDPZE3/trace/yjTHthSAZUS?fields[]=s_name&fields[]=s_serviceName&span=bfcdb92ca55f7881

I've also added the option to filter out the sqlalchemy auto-instrumented traces (will filter by default). This will reduce our trace load by an order of magnitude, and in my opinion doesn't cost us much visibility. The example with this filtering looks like this: https://ui.honeycomb.io/destiny-evidence/environments/test/datasets/destiny-worker-e2e/result/8UKR76MrHZs/trace/FNVBYGrZMVY?fields[]=s_name&fields[]=s_serviceName&span=d9aa5b4188c1e4af


jack-o-the-green (approved):
LGTM - this looks reasonable to me! I expect we'll make changes as we find we need them.

PR #243, opened 2025-08-06
Migrating from poetry to UV

jack-o-the-green:
Updated from `poetry` to `uv`

Switched over to using trusted publishers for PYPI so we don't need PYPI tokens in our github environments anymore.


Adam-Hammo (approved):
LGTM!

PR #244, opened 2025-08-07
Instrument/alerting

Adam-Hammo:
Adds a simple alerter with slack integration. The alerter runs on all datasets (eg robots also).

This will be too aggressive to start, but we can use it to refine both the application tracing and the trigger over time. Then we can consider pivoting to PagerDuty or similar.

Alert conditions:
- Errored span
  - There's an implementation detail here: if an error is raised and we catch it in a higher span, the raising span will still trigger this alert. For eg, a FastAPI exception handler will not suppress a honeycomb alert from firing. For now, I think that visibility is nice anyway as it can help us guide users.
- Error or critical level log
- 5XX status code

I've put a couple of test alerts in the `#testiny` slack channel.

NB - we can also manage boards via terraform, but that seems like more effort than desired for now. They're two clicks to duplicate across environments and will take some time for us to cement.

### DIFF
--- a/app/core/exceptions.py
+++ b/app/core/exceptions.py
@@ -307,8 +307,8 @@ class SDKTranslationError(DestinyRepositoryError):
             errors (str): A sequence of errors, likely copied from ValidationError
 
         """
-        super().__init__(str(errors))
         self.errors = errors
+        super().__init__(str(self))
 
     def __str__(self) -> str:
         """Convert pydantic exception errors to string."""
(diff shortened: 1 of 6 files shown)


jack-o-the-green (commented):
Might be missing something here, but I'm not seeing how this limits to production only this seems like it would run an apply on staging when merged?

jack-o-the-green (approved):
LGTM

PR #247, opened 2025-08-11
Added OpenAlexIdentifier validator for removing url prefix

jack-o-the-green:
Model now strips out the openalex url when creating the model. 

Also removed an unused function that looks like it was an intermediate step to the DOIIdentifier validator


libs/sdk/src/destiny_sdk/identifiers.py
```diff
@@ -75,6 +68,16 @@ class OpenAlexIdentifier(BaseModel):
         ExternalIdentifierType.OPEN_ALEX, description="The type of identifier used."
     )
 
+    @field_validator("identifier", mode="before")
+    @classmethod
+    def remove_open_alex_url(cls, value: str) -> str:
+        """Remove the OpenAlex URL if it exists."""
+        return (
+            value.removeprefix("http://openalex.org.")
```
Adam-Hammo, line 76: should this be a `/` not a `.` at the end?

jack-o-the-green: Yes! I don't know how that happened 🤦

jack-o-the-green: All fixed

Adam-Hammo (approved):
LGTM

PR #248, opened 2025-08-11
Fix infra var name for log_endpoint

jack-o-the-green:
Quick fix to rename the `logs_endpoint` variable to `log_endpoint` so that it aligns with the expected variable name in the application config.py

### DIFF
--- a/infra/app/variables.tf
+++ b/infra/app/variables.tf
@@ -231,7 +231,7 @@ variable "honeycombio_meter_endpoint" {
   default     = "https://api.honeycomb.io/v1/metrics"
 }
 
-variable "honeycombio_logs_endpoint" {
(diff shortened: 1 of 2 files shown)


infra/app/variables.tf
```diff
@@ -231,7 +231,7 @@ variable "honeycombio_meter_endpoint" {
   default     = "https://api.honeycomb.io/v1/metrics"
 }
 
-variable "honeycombio_logs_endpoint" {
+variable "honeycombio_log_endpoint" {
   description = "Meter endpoint for Honeycomb.io"
```
Adam-Hammo, line 235:
```suggestion
  description = "Logging endpoint for Honeycomb.io"
```

Adam-Hammo (approved):

PR #254, opened 2025-08-13
generalise metrics datasets across api and worker

Adam-Hammo:
Follows best practices for non-trace data: https://docs.honeycomb.io/get-started/best-practices/organizing-data/#general-datasets.

Also added a correlation ID for each instance. I wasn't able to hit the [IMDS](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux) for some reason, so mapping to the actual Azure instance is manual, but didn't feel worth the rabbit hole.

### DIFF
--- a/app/core/telemetry/otel.py
+++ b/app/core/telemetry/otel.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import uuid
 from typing import TYPE_CHECKING
 
 from opentelemetry import metrics, trace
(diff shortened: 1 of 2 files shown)


app/core/telemetry/otel.py
```diff
@@ -49,12 +50,15 @@ def configure_otel(
     if config.api_key:
         headers["x-honeycomb-team"] = config.api_key
 
+    service_instance_id = str(uuid.uuid4())
```
jack-o-the-green, line 53: Just confirming we want the instance id to change every time we restart the service? I feel that does make sense, since it's a new instance.

Adam-Hammo: Yeah - mainly for workers, means we can compare the metrics of a single worker with a trace on it instead of looking in aggregate.

jack-o-the-green (approved):
LGTM

Issue #265, opened 2025-08-19
Fold single enhancement flow into batch enhancement flow and simplify robots


jack-o-the-green:
### Background
Currently we are supporting requests for an enhancement against a specific reference as a flow entirely independent of the flow for enhancements against a group of references. This doubles the work for changes to our enhancement flows and requires robots to implement additional endpoints.

It also further complicates our data management by handling the single enhancement requests via a different database table and model. 

We want to consolidate these flows so that a single enhancement against a specific reference is serviced by triggering an enhancement request where the list of reference ids contains a single element, and is then processed through the batch enhancement implementation 

### Proposed API Changes
`POST /v1/enhancement-requests/single-requests/` -> removed 
`POST /v1/enhancement-requests/batch-requests/` -> `POST /v1/enhancement-requests/`

`GET /v1/enhancement-requests/single-requests/{request_id}/` -> removed
`GET /v1/enhancement-requests/batch-requests/{request_id}/` -> `GET /v1/enhancement-requests/{request_id}/`

`POST /v1/enhancement-requests/single-requests/{request_id}/results/` -> removed
`POST /v1/enhancement-requests/batch-requests/{request_id}/results/` -> `POST /v1/enhancement-requests/{request_id}/results/`

The `batch-request` resource essentially becomes the `enhancement-request` resource and the `single-request` one goes away

### Technical scope
- Remove the single enhancement workflow including but not limited to 
  - Single enhancement endpoints
  - Unnecessary models  
  - Combining database tables
  - Related documentation 
  - posting to the single enhancement endpoint from the sdk client
- Updating existing robots to remove the endpoints for servicing single enhancements 

This will likely result in backwards incompatible sdk changes, which should be considered when rolling out.

jack-o-the-green:
The destiny repository pieces of this have been merged and deployed. We've kept `POST /v1/enhancement-requests/batch-requests/{request_id}/results/` on as a deprecated endpoint so robots work with either version of the sdk. 

PR to update the toy robot is here https://github.com/destiny-evidence/toy-robot/pull/18

Remaining work is to update the following robots 
- domain inclusion robot
- fetch abstracts robot 
- taxonomy classification robot 

Then we can remove the endpoint here https://github.com/destiny-evidence/destiny-repository/blob/main/app/domain/references/routes.py#L321

Issue #181, opened 2025-07-08
Replace Init Containers with Container App Jobs for database migrations


jack-o-the-green:
**Replace init containers with container app jobs**
We've landed on Container App Jobs as an appropriate solution to replace init containers for database migrations. We have already implemented one for managing elasticsearch https://github.com/destiny-evidence/destiny-repository/blob/main/infra/app/main.tf#L687.

This has the added benefit of allowing us to run easy database downgrades in the instance a migration fails, which was previously tricky. And with a single job replica, allows us to avoid prior race conditions.

As part of this work we may want to make local terraform modules for each container app job, as we're expecting to have at least three, and `main.tf` is getting quite large. https://developer.hashicorp.com/terraform/tutorials/modules/module-create

We should be able to trigger a container app job from github actions when deploying, though without official actions we'll need to use CLI commands.
We'll also need to confirm we can update the container image without needing to re-specify the majority of the container manifest (given they appear more like container apps, we're confident this is the case).

Once this is implemented we can clean up the environment based image tags from our github actions (for example https://github.com/destiny-evidence/destiny-repository/blob/main/.github/workflows/promote.yml#L12) which will simplify things nicely!

**Background**
We need to look into an alternative solution for running database migrations and any set-up required prior to deploying a new version of destiny repository, as it has become apparent the init containers are unsuitable. 

Issues we're facing with the init containers 
* Run once per replica instead of once per deployment, making them unsuitable for tasks (like database migrations) which need to happen against a single instance of some infrastructure e.g. database, elasticsearch.
* Init containers do not support managed identity authentication, meaning we are prevented from implementing authentication/security best practices
* Init containers cannot have their images updated without re-specifying the entire container manifest (otherwise it will do thinks like wipe environment variables. This has introduced undesirable complexity into our deployment process where we haveto continually update pre-specified image tags to have the init container pick up the right image.

We would like a replacement for init containers that resolves each of the above issues.

jack-o-the-green:
Note that this is preventing us from doing data migrations using alembic, for example to combine the `enhancement_request` and `batch_enhancement_request` tables as part of #265. Because the migrations will run once per replica, we would expect race conditions or collisions between the data.

Adam-Hammo:
Note also they can be problematic for expensive operations, eg managing indexes on existing tables, as the duration makes race conditions more likely.

jack-o-the-green:
[...]

PR #271, opened 2025-08-21
Consolidate enhancement workflows

jack-o-the-green:
### What's changed?
Have removed the single enhancement workflow as per #265 

Also updated our process for adding enhancements to optionally verify derived_from enhancements are for the same parent reference. 

### Database migration complications
Database migration will need to happen in two parts due to #181. Prior to merging this PR we will need to the following in staging (and then again in prod before promoting) 

```
BEGIN;

INSERT INTO batch_enhancement_request(id, reference_ids, robot_id, request_status, enhancement_parameters, error, created_at, updated_at, source)
SELECT id, array_agg(reference_id), robot_id, request_status::text::batch_enhancement_request_status, enhancement_parameters, error, created_at, updated_at, source
FROM enhancement_request
GROUP BY id;

DELETE from enhancement_request;

COMMIT;
```
Have tested this out locally with our data seed, and will also give it a spin in development before merging.


app/domain/references/service.py
```diff
@@ -376,67 +251,6 @@ async def get_batch_enhancement_request(
...
-    ) -> EnhancementRequest:
-        """Mark an enhancement request as failed and supply error message."""
-        return await self.sql_uow.enhancement_requests.update_by_pk(
-            pk=enhancement_request_id,
-            request_status=EnhancementRequestStatus.FAILED,
-            error=error,
-        )
-
     @sql_unit_of_work
     async def collect_and_dispatch_references_for_batch_enhancement(
```
danielbreves, line 255: Do we need more test coverage for failure scenarios in this method perhaps?

jack-o-the-green: I think so - I will add some in the PR I'm stacking on top of this one to rename the batch enhancements!

jack-o-the-green:
Moving this out to a detached epic branch

PR #274, opened 2025-08-25
Add database seed data for local development

jack-o-the-green:
To aid in the testing of #265, but standalone work. 

Add some database seed data to facilitate the testing of db migrations locally. Will register the toy robot to it's expected local url so it can be used. Primarily generated from leftover data from our E2E tests with a few changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,3 +231,17 @@ docker compose down -v \
 && docker compose -f docker-compose.yml -f docker-compose.e2e.yml -f docker-compose.observable.yml --profile e2e up -d --force-recreate \
 && docker compose -f docker-compose.yml -f docker-compose.e2e.yml -f docker-compose.observable.yml logs -f --tail=0 e2e app worker
 ```
+
+### Seeding the database
+
+```sh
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -70,6 +70,22 @@ chmod +x ./.minio/seed_fileserver.sh
...
+```sh
+PGPASSWORD=localpass psql -U localuser -h 0.0.0.0 -p 5432 -d destiny_dev -f ".db_seed/local.sql"
+```
+
+##### Updating the database seed
+
+After applying a migration to the database, you may need to update the data seed. Apply the migrstion locally, run the following command, and then commit the result.
+
+```sh
+PGPASSWORD=localpass pg_dump --data-only --exclude-table=alembic_version -U localuser -h 0.0.0.0 -p 5432 destiny_dev > .db_seed/local.sql
```
jack-o-the-green, line 86: We need to exclude the `alembic_version` table as alembic takes care of creating this itself and we are simply inserting the db data after running `alembic update head`

Issue #260, opened 2025-08-18
Add de-duplication service


Adam-Hammo:
For now this is a "black box" to be completed, with interfaces as described in https://miro.com/app/board/uXjVJVIl8IQ=/.

This should live as a sub-service in the reference domain.

This card should also include the creation of the `reference_duplicate_decision` table.

PR #275, opened 2025-08-25
Duplicate Detection Structure

Adam-Hammo:
Partially implements #260, creating a relatively abstract service and data layer for interfacing with ingestion processes.

There's more to do on the dedup service layer, but the data layer got complex and low-level so I won't pollute this PR any further. The next PR will flesh that out more.

Various notes to highlight:
- All data is tied together with the `reference_duplicate_decision` table, and de-duplication itself actually occurs when projecting. 
  - Very open question about how we'd want Elasticsearch to look. Taking opinions!
- A canonical reference does not own its duplicates (more precisely, duplicates & decisions are not part of the Reference aggregate), but it can view and use them to generate its own enhancements. I think this gives us the most levers to both project however we want (particularly into elasticsearch), and to evaluate/correct/evolve duplicate mappings.
- The actual interface (API, SDK) is unchanged. The repository abstracts away any duplicate complexity via the projection. This has the side-effect of not requiring any changes from the robots to expose duplicate reference's enhancements to them.
  - Of course this will probably change, but we can wait until we have a concrete use case to add a new projection.

Also adds a little `Projection` pattern :)

### DIFF
--- a/app/domain/base.py
+++ b/app/domain/base.py
@@ -70,7 +70,7 @@ class ProjectedBaseModel(BaseModel):
     """
     Base model for all projected models to inherit from.
 
-    Projected models are models derived from a domain model. They duplicate or process
+    Projected models are models derived from a domain model. They derive or process
     data in helpful ways and may also be persisted.
     """
 
--- a/app/domain/references/repository.py
+++ b/app/domain/references/repository.py
@@ -1,11 +1,12 @@
 """Repositories for references and associated models."""
 
+import math
 from abc import ABC
 from collections.abc import Sequence
 from uuid import UUID
 
 from elasticsearch import AsyncElasticsearch
-from elasticsearch.dsl import AsyncMultiSearch, AsyncSearch, Q, Response
+from elasticsearch.dsl import AsyncMultiSearch, AsyncSearch, Q
 from opentelemetry import trace
 from sqlalchemy import or_, select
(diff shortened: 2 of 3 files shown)


app/domain/references/services/deduplication_service.py
```diff
@@ -0,0 +1,106 @@
...
+from app.domain.service import GenericService
+from app.persistence.es.uow import AsyncESUnitOfWork
+from app.persistence.sql.uow import AsyncSqlUnitOfWork
+
+logger = get_logger(__name__)
+settings = get_settings()
+tracer = trace.get_tracer(__name__)
+
+
+class DeduplicationService(GenericService[ReferenceAntiCorruptionService]):
```
Adam-Hammo, line 31: This is partly a stub and is not tested/called anywhere. Next PR will flesh it out, this one is big enough.

tests/unit/domain/references/models/test_models.py
```diff
@@ -139,353 +133,3 @@ async def test_enhancement_unserializable_failure(
                 ],
             )
         )
-
-
-class TestReferenceMerge:
```
Adam-Hammo, line 144: Not relevant anymore as we are no longer merging, just projecting. The method itself will be pivoted to a "contentwise compare" instead.

Adam-Hammo:
Closing, will be superseded.

PR #276, opened 2025-08-26
265 consolidate enhancement workflows

jack-o-the-green:
### What's changed?
Have removed the single enhancement workflow as per #265 

Also updated our process for adding enhancements to optionally verify derived_from enhancements are for the same parent reference. 

### Database migration complications
Database migration will need to happen in two parts due to #181. Prior to merging this PR we will need to the following in staging (and then again in prod before promoting). Obviously this is not ideal and definitely highlights the problem with the init containers. 

```
BEGIN;

INSERT INTO batch_enhancement_request(id, reference_ids, robot_id, request_status, enhancement_parameters, error, created_at, updated_at, source)
SELECT id, array_agg(reference_id), robot_id, request_status::text::batch_enhancement_request_status, enhancement_parameters, error, created_at, updated_at, source
FROM enhancement_request
GROUP BY id;

DELETE from enhancement_request;

COMMIT;
```
Have tested this out locally with our data seed, and will also give it a spin in development before merging.


app/core/exceptions.py
```diff
@@ -246,22 +246,6 @@ def __init__(self, detail: str, *args: object) -> None:
         super().__init__(detail, *args)
 
 
-class WrongReferenceError(InvalidPayloadError):
```
Adam-Hammo, line 249: Just an FYI, I'm reviving this in my PR as there's a case I'd like to throw it in `Reference.merge()`. It might be easiest to reinstate here so we don't get git headaches?

jack-o-the-green: Oh nice, yeah I'll put that back

Issue #217, opened 2025-07-23
Investigate implementing user management


danielbreves:
We would like to enable a lightweight mechanism to allow trusted internal users to consume our API safely. It would be great to investigate a potential solution and break down into different tasks.

See https://covidence.slack.com/archives/C0810ARPCRG/p1753053429459649

PR #279, opened 2025-08-26
217: External directory

danielbreves:
🚧 #217 

Setup an external Entra directory for authentication and authorization.

### DIFF
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -261,7 +261,7 @@ class AzureJwtAuth(AuthMethod):
 
         """
         self.tenant_id = tenant_id
-        self.api_audience = f"api://{application_id}"
+        self.api_audience = application_id
         self.scope = scope
         self.role = role
         self.cache: TTLCache = TTLCache(maxsize=1, ttl=cache_ttl)
@@ -305,7 +305,7 @@ class AzureJwtAuth(AuthMethod):
                     rsa_key,
                     algorithms=["RS256"],
                     audience=self.api_audience,
-                    issuer=f"https://sts.windows.net/{self.tenant_id}/",
+                    issuer=f"https://login.microsoftonline.com/{self.tenant_id}/v2.0",
                 )
             except exceptions.ExpiredSignatureError as exc:
                 raise AuthError(
--- a/infra/app/auth_v2.tf
+++ b/infra/app/auth_v2.tf
@@ -24,6 +24,8 @@ resource "azuread_application" "external_directory_destiny_repository" {
   sign_in_audience = "AzureADMyOrg"
 
   api {
+    requested_access_token_version = "2"
+
     oauth2_permission_scope {
       admin_consent_description  = "Allow the app to administer the system as the signed-in user"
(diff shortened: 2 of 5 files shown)


app/api/auth.py
```diff
@@ -305,7 +305,7 @@ async def verify_token(self, token: str) -> dict[str, Any]:
                     rsa_key,
                     algorithms=["RS256"],
                     audience=self.api_audience,
-                    issuer=f"https://sts.windows.net/{self.tenant_id}/",
+                    issuer=f"https://{self.tenant_id}.ciamlogin.com/{self.tenant_id}/v2.0",
```
danielbreves, line 308: This is very poorly documented, but External tenants need this format for auth to work: https://learn.microsoft.com/en-us/entra/identity-platform/security-tokens#token-endpoints-and-issuers

infra/app/auth_v2.tf
```diff
@@ -0,0 +1,257 @@
...
+  provider             = azuread.external_directory
+  application_id       = azuread_application.external_directory_destiny_repository.id
+  allowed_member_types = ["Application"]
+  description          = "Can request enhancements"
+  display_name         = "Enhancement Request Writer"
+  role_id              = random_uuid.external_directory_enhancement_request_writer_role.result
+  value                = "enhancement_request.writer"
+}
+
+resource "azuread_application_app_role" "external_directory_robot_writer" {
```
Adam-Hammo: I believe this was initially omitted deliberately as apps don't need to register robots. Temporarily re-included to ensure a smooth flow as `robot.writer` is actively being used for testing but we should re-remove.

libs/sdk/pyproject.toml
```diff
@@ -34,7 +34,7 @@ description = "A software development kit (sdk) to support interaction with the
 license = "Apache-2.0"
 name = "destiny_sdk"
 readme = "README.md"
-requires-python = "~=3.12"
+requires-python = ">=3.12, <4"
```
danielbreves: Fixes a warning: `requires-python ~=3.12 will be interpreted as >=3.12, <4`

app/api/auth.py
```diff
@@ -302,21 +303,29 @@ async def verify_token(self, token: str) -> dict[str, Any]:
...
         if rsa_key:
             try:
+                expected_issuer = f"{self.login_url}/v2.0"
                 payload = jwt.decode(
                     token,
                     rsa_key,
                     algorithms=["RS256"],
                     audience=self.api_audience,
-                    issuer=f"https://sts.windows.net/{self.tenant_id}/",
+                    issuer=expected_issuer,
```
danielbreves: Issuer and audience are changing even with `var.external_directory_enabled = false` because of [requested_access_token_version = 2](https://github.com/destiny-evidence/destiny-repository/pull/279/changes#diff-d43436ee691123b4a1be741d202aded4e2c23e8863cae12a26e19b115ff0ffb0R24). This will make the transition to the External directory a bit simpler.

PR #286, opened 2025-09-05
285 decouple import result tasks

Adam-Hammo:
Mostly backwards-compatible implementation of per-reference task distribution for imports.

Breaking changes:
- Removal of callbacks
- (not technically breaking but a new behaviour) robot automations firing off one reference at a time. Not currently in a production workflow, temporary until robot polling is live.

Ready for review, just a couple of peripherals to add: 
- [x] more testing on the import_record task
- [x] updated sequence diagram in the docs

### DIFF
--- a/app/domain/imports/models/sql.py
+++ b/app/domain/imports/models/sql.py
@@ -31,6 +31,7 @@ from app.domain.imports.models.models import (
 from app.domain.imports.models.models import (
     ImportResult as DomainImportResult,
 )
+from app.persistence.sql.generics import GenericSQLPreloadableType
 from app.persistence.sql.persistence import GenericSQLPersistence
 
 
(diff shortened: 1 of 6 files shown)


app/domain/imports/models/models.py
```diff
@@ -93,15 +85,15 @@ class ImportResultStatus(StrEnum):
...
       field for more information.
-    - `cancelled`: Processing was cancelled by calling the API.
+    - `retrying`: Processing has failed, but is being retried.
     """
 
     CREATED = auto()
     STARTED = auto()
     COMPLETED = auto()
-    CANCELLED = auto()
     PARTIALLY_FAILED = auto()
```
Adam-Hammo, line 94: I'm thinking we might want to all-or-nothing success on a reference import, removing this status. It's a bit awkward allowing a record that is not fully conformant to our data structures. Thoughts?

app/domain/imports/models/sql.py
```diff
@@ -35,6 +38,59 @@
 from app.persistence.sql.persistence import GenericSQLPersistence
 
 
+class ImportResult(GenericSQLPersistence[DomainImportResult]):
```
Adam-Hammo, line 41: This has just moved north on the page to make the ImportBatch column_property easier to define

danielbreves (approved):

PR #291, opened 2025-09-18
Deduplicating imports

Adam-Hammo:
Feature-flagged integration of deduplication with the import process. Implements the left column on the Miro. 

This is mostly pathing, the main change to note is that robot automations move to post-deduplication, and accordingly the ES structure contains a couple of primitive deduplication details to percolate against (may be useful elsewhere too). Take this with a grain of salt, I'll kick up a discussion about our ES structure before this is merged.


app/domain/references/tasks.py
```diff
@@ -269,10 +269,35 @@ async def process_reference_duplicate_decision(
...
         )
 
+        if reference_duplicate_decision.active_decision:
+            requests = await detect_and_dispatch_robot_automations(
+                reference_service=reference_service,
+                reference_ids=[
+                    # Automate on the canonical reference if it exists,
+                    # otherwise the base reference (which is either a canonical
+                    # reference or needing more information to deduplicate).
+                    reference_duplicate_decision.canonical_reference_id
```
Adam-Hammo, line 290: NB will probably need to update this to use the deduplicated projection for percolation? Open to counter-thoughts here, but assume in general we'd like to avoid processing enhancements on duplicates?

app/domain/imports/tasks.py
```diff
@@ -147,7 +150,13 @@ async def import_reference(
             logger.info("No remaining retries for reference import, marking as failed.")
         return
 
-    if (
+    if duplicate_decision_id:
```
jack-o-the-green, line 153: Do we want to queue the deduplication task (or even create the duplicate decision in the first place) if the deduplication feature flag is turned off?

app/domain/references/models/es.py
```diff
@@ -265,6 +279,10 @@ def from_domain(cls, domain_obj: Reference) -> Self:
                 EnhancementDocument.from_domain(enhancement)
                 for enhancement in domain_obj.enhancements or []
             ],
+            at_rest=domain_obj.duplicate_decision is not None,
```
jack-o-the-green, line 282: Might be reading this wrong, but this looks like it would consider a reference `at_rest` if the duplication decision was still pending/nominated etc - is that right?

app/domain/references/services/deduplication_service.py
```diff
@@ -129,19 +129,22 @@ async def register_duplicate_decision_for_reference(
...
-                else DuplicateDetermination.UNSEARCHABLE
-                if not CandidateDuplicateSearchFieldsProjection.get_from_reference(
-                    reference
-                ).searchable
-                else DuplicateDetermination.PENDING
-            ),
+            duplicate_determination=_duplicate_determination,
             canonical_reference_id=canonical_reference_id,
+            # If not pending, the decision is terminal and hence active
+            active_decision=_duplicate_determination != DuplicateDetermination.PENDING,
```
jack-o-the-green, line 147: Is that because it can't be anything other than a terminal or pending here? i.e. it's not possible for it to be in `nominated` at this point? 

I'm wondering if it would be nicer to do a "in terminal states" check here?

app/domain/references/models/es.py
Adam-Hammo: Duplicate decision is only included if it's active, which only happens once it's reached a terminal state. Can this be made clearer? Should I rename the domain attribute to `active_duplicate_decision`?

app/domain/imports/tasks.py
Adam-Hammo: The feature flagging actually "occurs" earlier - it will always return None for `duplicate_decision_id` if the feature flag is off

app/domain/references/models/es.py
jack-o-the-green: `active_duplicate_decision` this would be great actually!

app/domain/references/services/deduplication_service.py
jack-o-the-green: Ahh gotcha, makes sense!

Issue #267, opened 2025-08-20
Canonicalization: allow percolation on reference + changeset


Adam-Hammo:
Currently robot automations percolate on a reference, if it's an import, or on an enhancement, if it's a robot fulfillment.

To automate canonicalization, we need to update the enhancement percolation flow. For instance, if a bibliographic enhancement is added to a reference, we need to know if the reference already has a bibliographic enhancement - if so, we should send it for canonicalization. On the other hand, if we then added an abstract, we don't want to notice the double-bibliography-enhancement and send it again, hence needing to know the changeset.

Thus canonicalization robots will need the robot fulfillment automation flow to provide both the updated reference and the added enhancement (a "changeset").

PR #295, opened 2025-09-23
Dedup/automation

Adam-Hammo:
The implication of deduplicated automations were pretty involved - to the point I ended up incidentally fulfilling #267.

TL;DR, more detail in the docs:
- percolation on imports is performed on the canonical reference after deduplication
- percolation on enhancements is performed on its own reference, even if it's a duplicate
  - This is a necessary sin - ofc open question as to if/how/should duplicates be enhanced, race condition exists in any case necessitating this
- automation queries now see both the entire reference and the changeset (#267). This both makes the percolating queries nicer and more powerful :)

A couple more unit tests to come on the new service methods.

### DIFF
--- a/app/domain/references/service.py
+++ b/app/domain/references/service.py
@@ -712,41 +712,10 @@ class ReferenceService(GenericService[ReferenceAntiCorruptionService]):
         See the note in docstring of detect_robot_automations().
         """
         enhancements = await self.sql_uow.enhancements.get_by_pks(enhancement_ids)
-        # Note: we do not bubble up to automate on the canonical reference here.
-        # Most likely, if we are fulfilling an enhancement here on a duplicate, some
-        # other process has changed the duplicate determination between request &
-        # fulfillment.
-        #
-        # Here are the broad justifying statements:
-        # [A]: Enhancements should be generated on canonical references where possible
-        #      as this provides the most context to the enhancing robot.
-        # [B]: Because this reference is a duplicate, we can be confident that
-        #      automations were triggered on the canonical reference when the duplicate
-        #      decision was made.
-        # Thus:
-        # - Because of A, we would rather not trigger automation on enhancements derived
-        #   purely from a duplicate reference.
-        # - Because of B, we can be confident that there is no missed automation pathway
-        #   by not bubbling up the automation trigger to the canonical reference.
-        #
-        # For a concrete-ish example (there are many forms this could take):
-        # - E is a domain inclusion example, requiring a DOI and an abstract
-        # - C is an existing reference with a good abstract but no DOI
-        # - D is a newly ingested reference with a DOI and a partial abstract
-        # - D is imported and incorrectly marked as canonical.
-        # - Automation is fired on D and sent to robot to add E (let's call this E(D))
-        # - A new duplicate decision is made marking D as duplicate of C
-        # - Automation is fired on C with D as the changeset (let's call this E(C,D))
-        #   - This is statement [B] above
-        # - Robot processes and returns E(D) on D
-        #   - We automate E(D) on D
-        #   - (Likely this is a no-op as most automations will filter for canonicals)
-        # - Robot processes and returns E(C,D) on C
-        #   - We automate E(C,D) on C (our preferred path per [A])
-        #
-        # N.B. this also retains the option to enhance duplicates independently if
-        # desired (relevant automations will need to not filter for
-        # duplicate_determination=canonical, this is just a niche bonus).
+        # Enhancements are always automated against the references they're imported on.
+        # In most cases this will be the canonical reference, triggered by automation.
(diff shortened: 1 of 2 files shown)


app/domain/references/repository.py
```diff
@@ -558,17 +559,14 @@ async def percolate(
...
             ]
-            reference_ids = {
-                match.id if isinstance(match, DomainReference) else match.reference_id
-                for match in matches
-            }
             robot_automation_percolation_results.append(
                 RobotAutomationPercolationResult(
-                    robot_id=result.robot_id, reference_ids=reference_ids
+                    robot_id=result.robot_id,
+                    reference_ids={match.id for match in matches},
```
jack-o-the-green: I'm wondering if this is nicer as 

`reference_ids={reference.id for reference in matches}`

app/domain/references/service.py
```diff
@@ -99,6 +101,109 @@ async def _get_reference(self, reference_id: UUID) -> Reference:
...
+    async def _get_deduplicated_reference(self, reference_id: UUID) -> Reference:
+        """
+        Get the deduplicated reference for a given reference.
+
+        :param reference_id: The ID of the reference to get the deduplicated view for.
+        :type reference_id: UUID
+        :return: The deduplicated reference.
+        :rtype: Reference
+        """
+        reference = await self.sql_uow.references.get_by_pk(
```
jack-o-the-green: Should this just call `_get_duplicated_references` with `[reference_id]`. I guess it's nicer to return not a list in this case,  but feels like there's some duplicate code in there.

app/domain/references/service.py
```diff
@@ -99,6 +101,109 @@ async def _get_reference(self, reference_id: UUID) -> Reference:
...
+            not reference.duplicate_decision
+            or not reference.duplicate_decision.canonical_reference_id
+        ):
+            msg = (
+                "Reference is not canonical but has no canonical reference id. "
+                "This should not happen."
+            )
+            raise RuntimeError(msg)
+
+        return await self._get_deduplicated_canonical_reference(
```
jack-o-the-green: Should this just be a call to return await self._get_deduplicatedl_reference()` with the canonical id? We already know it's a canonical reference at this point right? Or no because of the potential for chaining?

docs/procedures/robot-automation.rst
```diff
@@ -72,7 +79,35 @@ The automation criteria is implemented as an `Elasticsearch percolator query <ht
...
-Importantly, the percolator query matches on **changesets**. On a reference import, this is of course the entire reference, but on an enhancement import, it is the enhancement itself. The query may therefore need to handle both cases, as in the :ref:`example below <domain-inclusion-example>`. It is guaranteed that only one of reference or enhancement will be provided for each percolation document.
+There are two scenarios that can trigger percolation:
+
+- On deduplication, if the active decision has changed
+- On added enhancement
+
+Structure
+---------
+
+Each percolated document contains two fields: ``reference`` and ``changeset``. Both of these fields map to :class:`Reference <app.domain.references.models.models.Reference>` objects. ``reference`` is the complete reference, deduplicated, and ``changeset`` is the delta that was just applied. The repository is append-only, and so is the ``changeset`` - it only represents newly available information to the reference.
```
jack-o-the-green:
```suggestion
Each percolated document contains two fields: ``reference`` and ``changeset``. Both of these fields map to :class:`Reference <app.domain.references.models.models.Reference>` objects. ``reference`` is the complete reference, deduplicated, and ``changeset`` is the change that was just applied. The repository is append-only, and so is the ``changeset`` - it only represents newly available information to the reference.
```

jack-o-the-green: I might be over-hating on the term `delta` here

app/domain/references/service.py
Adam-Hammo: Yeah I kept it self-referential for chaining implications. I don't know if it's naive to say it might save 5 mins when we cross that bridge?

docs/procedures/robot-automation.rst
```diff
@@ -125,29 +160,23 @@ This percolator query matches only on new references that do not have an abstrac
 Request Domain Inclusion Annotation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-This percolator query matches on new references that have an abstract, or new enhancements that are abstracts. This is an example of how the orchestration starts to piece together - if the above automation is executed, and an abstract is created, this automation will then be triggered.
+This percolator query matches on new references that have received an abstract. This is an example of how the orchestration starts to piece together - if the above automation is executed, and an abstract is created, this automation will then be triggered.
```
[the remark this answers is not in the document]
Adam-Hammo: Yeah this is poorly written, was meant to imply both cases. Updated the first sentence to:
`This percolator query matches on references that have received an abstract. This might either be on import, or on addition of a new abstract enhancement.`

PR #302, opened 2025-09-26
Add locks to azure service bus operations

Adam-Hammo:
Got a different variation on the service bus error overnight, which finally led me to:

image

~Naturally, clicking that link just takes you to the python azure service bus docs which don't mention concurrency locking.~ Link is probably meant to point here: https://learn.microsoft.com/en-us/python/api/overview/azure/servicebus-readme?view=azure-python#thread-safety


mootpointer (approved):
Assuming that our hypothesis is that the issue is with the client's lack of thread-safety and simultaneous access being the issue, it's worth giving this a shot!

PR #306, opened 2025-09-29
 GET endpoint for /v1/robot-enhancement-batches/{batch_id}/

jack-o-the-green:
Implements #300 

Added a GET endpoint for `/v1/robot-enhancement-batches/{batch_id}/` to allow robots to refresh signed urls while working with a batch. 

Also fixed some test warnings in the import service tests.

### DIFF
--- a/libs/sdk/pyproject.toml
+++ b/libs/sdk/pyproject.toml
@@ -33,7 +33,7 @@ license = "Apache-2.0"
 name = "destiny_sdk"
 readme = "README.md"
 requires-python = "~=3.12"
-version = "0.4.0"
+version = "0.4.1"
 
 [tool.pytest.ini_options]
 addopts = ["--color=yes", "--import-mode=importlib", "--verbose"]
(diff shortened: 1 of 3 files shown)


libs/sdk/src/destiny_sdk/client.py
```diff
@@ -130,7 +130,7 @@ def poll_robot_enhancement_batch(
         :rtype: RobotEnhancementBatch | None
         """
         response = self.session.post(
-            "/robot-enhancement-batch/",
+            "/robot-enhancement-batches/",
```
danielbreves, line 133: Thanks! Could you bump the SDK version please?

danielbreves (approved):
Made one minor comment, looks great otherwise! 👍

libs/sdk/src/destiny_sdk/client.py
jack-o-the-green: Ooh good reminder! Thanks!

PR #311, opened 2025-10-02
Dedup/e2e

Adam-Hammo:


app/domain/references/models/sql.py
```diff
@@ -216,12 +216,19 @@ class ExternalIdentifier(GenericSQLPersistence[DomainExternalIdentifier]):
     )
 
     __table_args__ = (
```
Adam-Hammo, line 218: Highlighting this & the relevant migration as the only significant application change here. This was a required constraint change as we don't treat identifiers as unique anymore. The two new indexes should provide type->identifier and type->other_type->identifier lookups. 

Happy to cherry-pick into a more relevant PR if we think that's better.

libs/sdk/src/destiny_sdk/enhancements.py
```diff
@@ -360,7 +360,12 @@ class EnhancementFileInput(BaseModel):
     visibility: Visibility = Field(
         description="The level of visibility of the enhancement"
     )
-    enhancement_type: EnhancementType = Field(description="The type of enhancement.")
+    enhancement_type: EnhancementType | None = Field(
+        default=None,
+        deprecated=True,
+        description="The type of enhancement. "
+        "Deprecated in favour of content.enhancement_type.",
+    )
```
Adam-Hammo, line 368: This is the only enhancement model (apart from SQL) with a root-level enhancement_type attribute.

Dockerfile.e2e
```diff
@@ -1,18 +1,12 @@
...
-RUN --mount=type=cache,target=/root/.cache/uv \
-    --mount=type=bind,source=uv.lock,target=uv.lock \
-    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
-    uv sync --locked --no-install-project --dev --group e2e
-
-COPY tests/e2e /e2e
-
-# Place executables in the environment at the front of the path
-ENV PATH="/app/.venv/bin:$PATH"
+# Doesn't use mounts to allow python to build the image (docker-py doesn't support buildkit)
```
Adam-Hammo, line 11: I'll see shortly if we can make the same change to the root Dockerfile - if so, I'll delete this one :)

app/domain/references/repository.py
```diff
@@ -516,6 +516,19 @@ def __init__(self, client: AsyncElasticsearch) -> None:
             RobotAutomationPercolationDocument,
         )
 
+    async def add(self, obj: DomainRobotAutomation) -> DomainRobotAutomation:
+        """
+        Add a robot automation to the repository and index it for percolation.
+
+        Wraps the base add method to ensure the index is refreshed.
```
Adam-Hammo, line 523: Don't ask how long it took me to figure this out.

app/domain/references/models/sql.py
mootpointer: I think that in a separate PR might make sense – just to best tell the story of the change.

Adam-Hammo: Shall do downstream

Dockerfile.e2e
Adam-Hammo: We _can_, but it's not the recommended pattern: https://docs.astral.sh/uv/guides/integration/docker/#intermediate-layers. Will keep the Dockerfiles separate.

tests/e2e/factories.py
```diff
@@ -0,0 +1,246 @@
...
+                AbstractContentEnhancementFactory(),
+                AnnotationEnhancementFactory(),
+                LocationEnhancementFactory(),
+            ]
+        )
+    )
+    reference_id = factory.Faker("uuid4")
+
+
+class ReferenceFactory(factory.Factory):
```
jack-o-the-green, line 210: I'm wondering if we want to move these to the base level for `tests/` as I'm thinking there's a bit of unit test cleanup we could use these for - for example https://github.com/destiny-evidence/destiny-repository/blob/main/tests/unit/domain/references/services/test_deduplication_service.py#L22

Adam-Hammo: Will do!

PR #316, opened 2025-10-09
General version bump and UUIDv7 on incoming records

Adam-Hammo:
Implements #448 by:
- updating to Python 3.14 (and updating relevant packages)
  - NB while not the bulk of the changes this is where I expect any problems would arise
- making all new ID generation UUIDv7
- expanding ID interfaces to expect both UUID4 and UUID7

This will provide denser and sortable PKs across the board, providing better dev experience and more efficient indexes. 

**All users will need to bump their SDK versions** as older SDK versions validate strictly against UUID4s. The version released with this PR will allow UUID4 and UUID7.

On its own, this fulfills the UUID section of #416.

Following this at some point in the future will be https://github.com/destiny-evidence/destiny-repository/issues/449 to update our existing IDs - there's some exploratory work on that ticket that I did while the context was fresh, but it's not an immediate priority.

Also worth noting, I came across this concern of UUID7 for security: https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-01.html#name-security-considerations. My view is it's pedantic and not applicable to our project but good to be across.

### DIFF
--- a/libs/sdk/pyproject.toml
+++ b/libs/sdk/pyproject.toml
@@ -35,7 +35,7 @@ license = "Apache-2.0"
 name = "destiny_sdk"
 readme = "README.md"
 requires-python = ">=3.12, <4"
-version = "0.8.0"
+version = "0.9.0"
 
 [project.optional-dependencies]
 labs = []
--- a/uv.lock
+++ b/uv.lock
@@ -666,7 +666,7 @@ test = [
 
 [[package]]
 name = "destiny-sdk"
-version = "0.8.0"
+version = "0.9.0"
 source = { editable = "libs/sdk" }
 dependencies = [
     { name = "cachetools" },


tests/e2e/conftest.py
```diff
@@ -213,32 +213,47 @@ async def pg_lifecycle(pg_sessionmanager: AsyncDatabaseSessionManager):
...
+    with (
         # If elasticsearch is failing to start, check the container logs. Exit code 137
         # means out of memory. Annoyingly, this either means:
         # - Docker daemon doesn't have enough memory allocated (mem_limit is too high)
         # - Elasticsearch doesn't have enough memory allocated (mem_limit is too low)
         # Fun!
-        mem_limit="2g",
-    ).with_name(f"{container_prefix}-elasticsearch") as elasticsearch:
+        ElasticSearchContainer(
+            "elasticsearch:9.0.2",
```
mootpointer: Why have we moved backwards in version?

Adam-Hammo: This now exactly matches what we use in cloud (I was chasing so many compatibility errors so I standardised it)

Adam-Hammo:
Thanks @mootpointer, I've updated the PR:
- Define `destiny_sdk.UUID = UUID4 | UUID7` and use it accordingly
  - This being part of the SDK should hopefully make it clear that it's used for extra validation at the interfaces
  - Internal domain models will use stdlib `UUID`, with the anti-corruption layer handling that transition
- Standardise `uuid.UUID -> UUID` stdlib imports
- Revert DB upgrade (I'll leave the next.js one in this time if that's okay?)

libs/sdk/src/destiny_sdk/identifiers.py
```diff
@@ -1,10 +1,26 @@
...
 from enum import StrEnum, auto
 from typing import Annotated, Literal, Self
 
-from pydantic import UUID4, UUID7, BaseModel, Field, TypeAdapter, field_validator
+from pydantic import BaseModel, Field, PositiveInt, TypeAdapter, field_validator
+
+from .core import UUID
+
+# Case-insensitive patterns for DOI URL prefix stripping
+_DOI_URL_PREFIX_RE = re.compile(r"^(?:https?://)?(?:dx\.)?doi\.org/", re.IGNORECASE)
```
mootpointer: Did this come with the merge too?

PR #323, opened 2025-10-15
Avoid concurrent use of SQLAlchemy session

Adam-Hammo:
Current implementation attempts to use asyncio to perform concurrent requests with the same SQLAlchemy session - this is a no-no: https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#using-asyncsession-with-concurrent-tasks. Noticed when testing in staging, added test coverage.

This PR adds a bulk method and refactors the single method to wrap bulk method.


Adam-Hammo:
> assuming this is the only concurrent use?

Correct

PR #325, opened 2025-10-16
remap no-candidate path to unsearchable

Adam-Hammo:
Tiny PR to relax the presumption that our unevaluated search strategy can determine if a reference is canonical.


jack-o-the-green (approved):
Makes sense to me!

PR #326, opened 2025-10-17
Spring cleaning: robot polling & deduplication

Adam-Hammo:
Cleaning out some old code paths and columns in the wake of recent feature work (primarily polling, removing the old repo->robot flows).

There's some other code that's technically orphaned (eg the repo&SDK code for repo->robot hmac auth) that feels like it's potentially still worth keeping - open to opinions.


app/migrations/versions/6f6d08aab362_deprecate_collision_strategy_update_type.py
```diff
@@ -0,0 +1,30 @@
...
+Create Date: 2025-10-02 23:27:25.113341+00:00
+
+"""
+from collections.abc import Sequence
+from typing import Union
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = '6f6d08aab362'
```
jack-o-the-green: Given you're repaired the chain of revision identifiers, we're not worried about risks of alembic mismatching up with our tables if we make those manual adjustments you've detailed in the PR description? I'm cagey about the amount of mucking about we're doing here 😅

cli/register_robot.py
```diff
@@ -88,7 +88,6 @@ def argument_parser() -> argparse.ArgumentParser:
     try:
         robot_to_register = destiny_sdk.robots.RobotIn(
             name=args.name,
-            base_url=args.base_url,
```
jack-o-the-green: in the `argument_parser()` function there's another reference to base_url we should remove? Or at least mark as not required.

jack-o-the-green (commented):
Love a cleanup PR. 

I'm for removing the orphaned code called out in the description `(eg the repo&SDK code for repo->robot hmac auth)`. If it's useful in the future it's not the end of the world to pull it back out of the repo history. Perhaps we could remove it in it's own separate branch but I think it's more confusing to keep it.

app/migrations/versions/6f6d08aab362_deprecate_collision_strategy_update_type.py
Adam-Hammo: I too am cagey, fortunately we're on this before it got to production though. 
If we're getting this in sooner rather than later, then my preference would be to downgrade dev & staging by 1 version, then deploy the PR and let alembic re-upgrade with the split versions. How does that sound?
In my view, manual becomes the better option if other database changes are deployed first - I wasn't expecting to have this approved before I even logged back on! 🚤

Adam-Hammo:
> Love a cleanup PR.
> 
> I'm for removing the orphaned code called out in the description `(eg the repo&SDK code for repo->robot hmac auth)`. If it's useful in the future it's not the end of the world to pull it back out of the repo history. Perhaps we could remove it in it's own separate branch but I think it's more confusing to keep it.

I went to do this in https://github.com/destiny-evidence/destiny-repository/pull/331 but realised that it's not actually fully orphaned - the repo still uses the SDK implementation for its server-side auth.

PR #327, opened 2025-10-20
Investigate sporadic GH action errors

Adam-Hammo:
Investigating the below that have started occurring >50% of the time since last week.

image

I don't exactly know the cause but while investigating found that `setup-python` is redundant with `setup-uv` (as long as we invoke uv, which we do). So, head in sand and removing instead.


jack-o-the-green (approved):
LGTM

Issue #320, opened 2025-10-10
Discard exact enhancements


Adam-Hammo:
Like with imports, discard incoming enhancements if they are already exactly existing on the reference (everything short of IDs and DB timestamps must match). 

This makes replaying tasks idempotent.

Issue #318, opened 2025-10-10
Break up validation & ingestion


Adam-Hammo:
Similar to imports, distribute the validation and ingesting of enhancements to multiple smaller tasks, one per enhancement.

This requires a bit of higher-level thought as it will decouple the result file from the ingesting process.

Adam-Hammo:
This can also include using the pending_enhancement table better to store results etc - instead of relying on the blob files.

PR #333, opened 2025-10-22
Discard exact duplicate enhancements

Adam-Hammo:
Implements #320, discarding exact duplicate enhancements if they already exist on the reference.

I've applied this to only the relevant reference, not its deduplicated projection (i.e., if a duplicate contains the enhancement it will still be added to the canonical). Open to opinions here - my thinking is it's just better to not be too aggressive.

Also added a minor bugfix and documentation update.


app/domain/references/models/models.py
```diff
@@ -476,12 +476,10 @@ class ReferenceDuplicateDeterminationResult(BaseModel):
...
 
     @model_validator(mode="after")
-    def check_canonical_reference_id_populated_iff_canonical(self) -> Self:
+    def check_canonical_reference_id_populated_iff_duplicate(self) -> Self:
         """Assert that canonical must exist if and only if decision is duplicate."""
-        if (
-            self.canonical_reference_id
-            is not None
-            == (self.duplicate_determination == DuplicateDetermination.DUPLICATE)
+        if (self.canonical_reference_id is not None) != (
```
Adam-Hammo: Oops - something we'd certainly pick up when we try to plug the dedup algo in

jack-o-the-green (commented):
Do we want a way of tracking how many duplicate enhancements we're getting? Right now it feels like duplicates will get hidden unless we're going through the validation results files ourselves. Might be overkill at this stage. 

I agree with only discarding duplicate enhancements when they clash on one reference. So a canonical and a duplicate can both have "identical" enhancements.

Adam-Hammo:
> Do we want a way of tracking how many duplicate enhancements we're getting?

My original thinking here was that it might make sense to include that if/when we break up loading enhancement result files into per-line tasks like we've done with imports, as this'll include some amount of restructuring (see #318 ). However, it feels probable to me that in any case adding `exact_duplicate` to `PendingEnhancementStatus` would be the solution in both contexts, and is easily implementable here. What do you think?

PR #341, opened 2025-10-23
Try using a new receiver on auto lock renewer

Adam-Hammo:
Speculative PR to try and mitigate the [Azure lock renewal failures](https://covidence.slack.com/archives/C099F6PR8R2/p1761182194599399).

We've seen elsewhere (#302) that concurrent receiver usage can be problematic. We can't lock the receiver in the auto lock renewer as it loops in a separate process. My hope is using a different receiver across renewing and sending will help (note that a new receiver per-message is too unperformant).

Separately, we're using the renewer on many tasks that don't actually need it - only the "big" tasks (loading enhancements, distributing imports) take a noticeable amount of time, everything else is quite atomic and a non-renewed lease of 1 minute (az default) would probably suffice. While not the root cause of the problem, it's probably enough to make it disappear?

Some other next steps we can consider:
- pull out the `AutoLockRenewer` from the service bus SDK and tweak it to add locks
- wrap stuff in retries - note we might need to reinstantiate the whole broker after a failure as the [state seems unrecoverable](https://covidence.slack.com/archives/C099F6PR8R2/p1761183269161169?thread_ts=1761182415.375289&cid=C099F6PR8R2) for the particular task if it gets picked up by the same task worker

Worth noting again that there's no downstream impact - the task itself gets completed in the first instance, and then gets picked up again and no-opped as the DB state shows it is complete. In that corner case where we get an inconsistent state, the task eventually gets dead-lettered.


app/core/azure_service_bus_broker.py
```diff
@@ -164,6 +171,9 @@ async def kick(self, message: BrokerMessage) -> None:
                 "message_id": message.task_id,
                 "correlation_id": message.task_id,
             },
+            annotations={
+                "renew_lock": message.labels.get("renew_lock", False),
```
jack-o-the-green: Just confirming - this means we (by default) don't renew locks? and then we renew locks on specific tasks (the ones that take longer)

Adam-Hammo: Correct - as now most of our tasks are very atomic (eg importing a single line from a file) and won't need to hold the lock for long

app/core/azure_service_bus_broker.py
```diff
@@ -201,7 +211,21 @@ async def listen(self) -> AsyncGenerator[AckableMessage, None]:
 
                 # Process each message
                 for sb_message in batch_messages:
-                    self.auto_lock_renewer.register(self.receiver, sb_message)
+                    if isinstance(sb_message, ServiceBusReceivedMessage):
```
jack-o-the-green: `nitpick` perhaps move this to its own function for exacting the annotations. Then it can return `{}` and you'd be able to make the next line `annotations.get("renew_lock", False)` which makes the whole listen function easier to read (for me at least)

jack-o-the-green: Could even abstract that whole piece into to allow `if should_renew_lock`

tests/core/test_azure_service_bus_broker.py
```diff
@@ -226,3 +226,39 @@ async def mock_send_messages(message: AmqpAnnotatedMessage) -> None:
     assert isinstance(sent_message, AmqpAnnotatedMessage)
     assert sent_message.header is not None
     assert sent_message.header.priority == 5
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("renew_lock", [True, False, None])
```
jack-o-the-green: This test is adding deprecation warnings for pytest, should fix before merging. 

```
PytestDeprecationWarning: asyncio test 'test_only_renew_lock_when_specified[True]' requested async @pytest.fixture 'broker' in strict mode. You might want to use @pytest_asyncio.fixture or switch to auto mode. This will become an error in future versions of flake8-asyncio.
```

app/core/telemetry/taskiq.py
```diff
@@ -31,6 +31,7 @@
 async def queue_task_with_trace(
     task: AsyncTaskiqDecoratedTask | tuple[str, str],
     *args: object,
+    renew_lock: bool = False,
```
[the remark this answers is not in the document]
Adam-Hammo: I'm a bit torn on this - I increasingly feel that long-running tasks are a bit of an anti-pattern in most cases, so I wonder if that extra step to need to turn on lock renewal (and acknowledging the default by doing so) is a helpful cognitive check. What do you think?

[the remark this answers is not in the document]
jack-o-the-green: Ahh okay - that makes sense to me. I was going for a cognitive check of "should this or should this not have a lock renewal" wheras if we want to move away from long running tasks completely then the implication of the current implementation is better. 

Additionally, we could replace `renew_lock` with `long_running` which might make more sense at the task level and also imply more badness. Not picky on this though.

PR #342, opened 2025-10-24
Add earlier idempotency check

Adam-Hammo:
Task is already technically idempotent through the `exact_duplicate` check, this circumvents that hassle also.

### DIFF
--- a/app/domain/imports/tasks.py
+++ b/app/domain/imports/tasks.py
@@ -100,9 +100,6 @@ async def import_reference(
         import_result = await import_service.get_import_result_with_batch(
             import_result_id
         )
-        if not import_result.import_batch:
-            msg = "Import result is missing its import batch. This should not happen."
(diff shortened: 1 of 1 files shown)


app/domain/imports/tasks.py
```diff
@@ -105,6 +105,18 @@ async def import_reference(
             raise RuntimeError(msg)
         trace_attribute(Attributes.IMPORT_BATCH_ID, str(import_result.import_batch_id))
 
+        if import_result.status in (
```
jack-o-the-green: Wondering if this should go above `if not import_result.import_batch:` or we'd get double errors if the previous task raised a runtime error?

Adam-Hammo: Fair call, will do!

jack-o-the-green (approved):
LGTM

Issue #338, opened 2025-10-22
Updates to robot API and handling


jack-o-the-green:
- Update the RobotEnhancementBatch request flow to allow robots to provide an optional batch lease time, with a sensible default for backwards compatibility.
- Use this lease time to populate `expires_at` on PendingEnhancements
- Add an endpoint that allows robots to renew a lease on a RobotEnhancementBatch 
- I think it makes sense to do this too, as part of the work https://github.com/destiny-evidence/destiny-repository/issues/294

PR #345, opened 2025-10-27
Implemented leasing and renewing api

Adam-Hammo:
Implements #338 adding leasing to the API layer

### DIFF
--- a/app/api/responses.py
+++ b/app/api/responses.py
@@ -1,0 +1,25 @@
+"""Standard API response models."""
+
+from fastapi.encoders import jsonable_encoder
+from fastapi.responses import JSONResponse
+from pydantic import BaseModel, Field
+
+
+class StringResponseContent(BaseModel):
+    """Model for simple string API responses."""
+
+    detail: str
+
+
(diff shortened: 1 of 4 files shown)


app/domain/references/service.py
```diff
@@ -922,6 +936,28 @@ async def create_robot_enhancement_batch(
...
+        await self.sql_uow.robot_enhancement_batches.verify_pk_existence(
+            [robot_enhancement_batch_id]
+        )
+        expiry = apply_positive_timedelta(lease_duration)
+        await self.sql_uow.pending_enhancements.bulk_update_by_filter(
+            filter_conditions={
+                "robot_enhancement_batch_id": robot_enhancement_batch_id,
+                # If a robot lets a pending enhancement expire, it must use a
+                # new robot enhancement batch to re-process it.
+                "status": PendingEnhancementStatus.PROCESSING,
```
danielbreves: This will fail silently if no pending enhancements are renewed right? I wonder if we should throw an error in that case.

danielbreves (approved):

PR #347, opened 2025-10-28
Discovery/simple UI

Adam-Hammo:
Repository "Operational" UI v0.0.0

Structured as a next.js subdirectory - easily separable into a different repository in the future, but I felt was worth keeping with the API for now as it keeps the infra together and highlights the coupling to the repository.

Users in the JT_AD tenant must be added to the `ui_users` group for access, which gives them the `reference.reader` role.

This is currently deployed to staging - I'll leave it there for the time being in case anybody wants to see how this shapes up. https://destiny-repository-ui-stag-app.salmonpebble-3bc0d53b.swedencentral.azurecontainerapps.io/references , try DOI `10.1136/gutjnl-2022-bsg.330`.

TODO:

- [x] Better local dev setup
- [ ] Other improvements will come in further PRs unless there's anything showstopping.

Obligatory disclaimer: a good chunk of the JS was vibed in the name of speed. I do pledge to straighten out any weirdness and ensure we've got an extensible UI once we've delivered some simple features and gotten the hang of it!

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,6 +14,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 permissions:
   id-token: write


infra/app/main.tf
```diff
@@ -167,8 +180,12 @@ module "container_app" {
     client_id    = azurerm_user_assigned_identity.container_apps_identity.client_id
   }
 
-  env_vars = local.env_vars
-  secrets  = local.secrets
+  env_vars = concat(local.env_vars, [{
+    name  = "CORS_ALLOW_ORIGINS",
+    value = jsonencode(["*"])
```
Adam-Hammo: We can limit this to the UI host once we got some DNS. Until then there's a circular dependency where the API needs the UI URL for CORS and the UI needs the API URL for calling.

ui/entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
Adam-Hammo: See also `runtimeConfig.ts`

infra/app/main.tf
```diff
@@ -262,6 +279,57 @@ module "container_app_tasks" {
   ]
 }
 
+module "container_app_ui" {
```
mootpointer: This is fine for now, but we should consider deploying this as an [Azure Static Web app](https://learn.microsoft.com/en-us/azure/static-web-apps/deploy-nextjs-static-export?tabs=github-actions) (which could also manage CORS stuff for us as well).

mootpointer (approved):
Looks fine for an initial cut. I haven't gone too deep into the particulars, since I'm treating this as somewhat of a prototype/MVP.

As we move forward we should have a discussion about stack (next.js vs vanilla JS vs whatever)

infra/app/main.tf
Adam-Hammo: Yeah I'd like to explore that - went with container app as a known quantity but the SWAs do look handy.

Issue #350, opened 2025-10-30
Bulk lookup of light-weight records by IDs


Adam-Hammo:
Scope: simple quick-win API + UI + export for lookup of multiple identifiers. Capped at sensible values.

**API Proposal**
- `GET /references/?identifier=identifier:type(?:other_type)&identifier=...`
  - Impose a sensible limit (100?)
  - Used for synchronous GUI display
  - Can potentially collapse the existing `GET /references/?identifier=...&identifier_type=...` into this. Both return the same response so I think it makes sense - we'd need a query param for `?expand` though perhaps to allow the single-reference lookup to still return an entire reference.
- **DESCOPED FROM THIS TICKET** `POST /references/export/` (hoping `export` provides a clear delineation with `search`) 
  - ```
    {
        "identifiers":[
            {
                  "identifier": "str|int",
                  "identifier_type": "str",
                  "other_identifier_name": "str|None",
            },
            ...
        ]
        "file_type":"csv|jsonl",
    }
    ```
  - Returns a signed URL to the file.  
  - For now, impose a limit (10,000?) and process synchronously. I don't expect this to be terribly slow, it's similar to the robot enhancement batch workflow.
  - In the future, convert to async task and expose a `GET /references/export/?request_id=` to allow for any number of identifiers.
    - I'm wary of implementing this first-up, there's a few considerations we can get bogged in (eg persistence structure for async requests, separate task queueing so it doesn't take 30 minutes during an import, etc)
- Both endpoints return slimmed projections of each reference (see below).
  - @mootpointer is this a hard requirement? I'm a little wary of implementing endpoints that don't return everything at this stage, because it may well become a problem for somebody who needs for eg abstracts. Particularly for exports I wonder if we should just return all, maybe only the GET->UI flow needs the projection.
  - Though I know it is far easier to add to a response model than to subtract from it down the road

**UI Proposal**
- TBD. Probably some form of collapsible table. Will get extraction cooking first

### Space I used for notes

**Sending a large payload**
- GET with body
  - Not standard
  - Elasticsearch does it though 🤔 
  - Some gateways drop the body. Ask me how I know
- POST with body
  - Not strictly standard for getting data as there's no server side-effect but pretty widely used
- Async POST with body & get with request ID
  - This might be easier than it sounds but opens some cans of worms
- GET with massive query string
  - This feels tenuous. We can probably get up to 8,000 chars in but even then that won't fill 1,000 IDs.
[...]

PR #351, opened 2025-10-31
Bulk Lookup & Display

Adam-Hammo:
Bulk lookup and display of lightweight references per #350. 

Limit somewhat arbitrarily set to 100 identifiers (configurable after deployment). It's not clear until we deploy where the limit (if any) on URL length itself will be imposed - presumably something in Azure's networking will cap it at 2,000 or 8,000 chars. In any case, we shouldn't bring this out too big unless we also add a second limit to the UI requesting - it's too early for our UI to be slow!

Leaning heavily on serverside validation to give helpful messages to the frontend.

Some screenshots:
image
image
image
image
image


app/domain/references/repository.py
```diff
@@ -329,90 +343,6 @@ def __init__(self, session: AsyncSession) -> None:
             SQLExternalIdentifier,
         )
 
-    @trace_repository_method(tracer)
-    async def get_by_type_and_identifier(
```
Adam-Hammo: Orphaned since dedup

app/domain/references/routes.py
```diff
@@ -218,26 +221,48 @@ async def get_reference(
     return anti_corruption_service.reference_to_sdk(reference)
 
 
+def parse_identifiers(
```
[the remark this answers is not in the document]
mootpointer: This is probably fine for now – there may be an option to use Pydantic type unions to improve this a bit further, but I feel fine about here.

app/domain/references/routes.py
```diff
@@ -218,26 +221,48 @@ async def get_reference(
...
-    external_identifier = ExternalIdentifierSearch(
-        identifier=identifier,
-        identifier_type=identifier_type,
-        other_identifier_name=other_identifier_name,
-    )
-    reference = await reference_service.get_reference_from_identifier(
-        external_identifier
-    )
-    return anti_corruption_service.reference_to_sdk(reference)
+    identifiers: Annotated[list[IdentifierLookup], Depends(parse_identifiers)],
```
mootpointer: I wonder if this is where we might do some limiting to prevent potential DoSes.

```suggestion
    identifiers: Annotated[list[IdentifierLookup], Len(max_length=settings.max_identifier_lookup_count), Depends(parse_identifiers)],
```

Adam-Hammo: Actioned this and also a hard cap in the validation (to validate against comma-separated identifiers).

mootpointer:
What does it look like when I supply say 10 valid IDs, but 2 of them are not found?

Adam-Hammo:
> What does it look like when I supply say 10 valid IDs, but 2 of them are not found?

Somewhat indeterminate - on a destiny ID it will fail, on an external ID it won't return one. It's probably easiest now to fail it serverside with a useful message, and defer any "results + query information" return structure for once we've got querying in place.

Adam-Hammo:
Now like this (1 exists, 3 do not). What do you think?
image

mootpointer (approved):
All good, barring a nitpick on a docstring.

PR #355, opened 2025-11-05
Reindexing increase polling interval

jack-o-the-green:


infra/app/main.tf
```diff
@@ -756,6 +756,11 @@ resource "azurerm_container_app_job" "es_index_migrator" {
         name  = "OTEL_ENABLED"
         value = var.telemetry_enabled
       }
+
+      env {
+        name  = "REINDEX_STATUS_POLLING_INTERVAL"
+        value = var.es_migrator_reindex_polling_interval
+      }
```
Adam-Hammo: For my own knowledge, not to block - can we override these on a single run?

PR #356, opened 2025-11-06
Added created_at field into enhancement data models and elasticsearch documents

jack-o-the-green:
This will allow us to prioritize enhancements for projection based on when they were added to the reference.

### DIFF
--- a/app/domain/references/models/es.py
+++ b/app/domain/references/models/es.py
@@ -113,9 +113,12 @@ class EnhancementDocument(InnerDoc):
     visibility: Visibility = mapped_field(Keyword(required=True))
     source: str = mapped_field(Keyword(required=True))
     robot_version: str | None = mapped_field(Keyword())
-    created_at: datetime.datetime = mapped_field(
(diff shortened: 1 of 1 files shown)


app/domain/base.py
```diff
@@ -61,13 +59,15 @@ class SQLTimestampMixin(SQLAttributeMixin):
     Timestamps should not be directly set by the application.
     """
 
-    created_at: datetime.datetime = Field(
-        default_factory=utc_now,
+    created_at: datetime.datetime | None = Field(
+        default=None,
```
jack-o-the-green: I've updated the default value here because that way the domain model never automatically sets a timestamp, which feels like an extra layer of security. Shout out if this feels like the wrong approach!

Adam-Hammo: This feels like a good idea - thematically it makes a bit of sense to me too, because when we create the Reference model we still haven't "created" it at the database, so None is a valid value. That might be a stretch though 😛

jack-o-the-green: That follows my thought train - if it's not persisted, in my mind it isn't created.

PR #357, opened 2025-11-06
Add convenience properties to reference

TimRepke:
In working with Max, we thought this might be a useful addition to the Reference model.

### DIFF
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -13,6 +13,7 @@ from pydantic import (
     UUID4,
     BaseModel,
     Field,
+    PositiveInt,
     TypeAdapter,
     model_validator,
 )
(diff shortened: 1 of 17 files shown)


libs/sdk/src/destiny_sdk/references.py
```diff
@@ -49,6 +57,114 @@ def from_es(cls, es_reference: dict) -> Self:
...
+            if enhancement.content.enhancement_type == EnhancementType.BIBLIOGRAPHIC:
+                yield enhancement.content
+
+    def annotations(
+        self,
+        source: str | None = None,
+        annotation_type: AnnotationType | None = None,
+        scheme: str | None = None,
+        label: str | None = None,
+    ) -> Generator[Annotation, None, None]:
```
Adam-Hammo: I'd maybe suggest these interfaces don't need to be generators - we've already got the data in memory, and a list might be a simpler interface for users (and would be consistent with the docstring).

libs/sdk/src/destiny_sdk/references.py
```diff
@@ -49,6 +57,114 @@ def from_es(cls, es_reference: dict) -> Self:
...
+                return meta.title
+        return None
+
+    def bibliographics(self) -> Generator[BibliographicMetadataEnhancement, None, None]:
+        """List bibliographic metadata enhancements."""
+        for enhancement in self.enhancements or []:
+            if enhancement.content.enhancement_type == EnhancementType.BIBLIOGRAPHIC:
+                yield enhancement.content
+
+    def annotations(
```
Adam-Hammo:
```suggestion
    def get_annotations(
```

libs/sdk/src/destiny_sdk/references.py
```diff
@@ -49,6 +57,114 @@ def from_es(cls, es_reference: dict) -> Self:
             ],
         )
 
+    def _get_id(self, kind: ExternalIdentifierType) -> str | None:
```
Adam-Hammo: PM_ID would actually return an int here which we should cast to a string before returning, or reflect in the signature. I'd also maybe suggest `_get_identifier()` and `identifier_type` over `kind` in the signature for a more consistent interface.

Adam-Hammo (commented):
Hey folks, thanks for your suggestions! I've added a few comments for you, the main missing piece here is a bit of testing in `libs/sdk/tests/unit/`.

More generally, I'm keen to discuss whether the SDK is the best place for this interface at the moment. We might need to be somewhat mindful with SDK changes as we will either need to continually support or deliberately deprecate each property/method as the repository & references evolve. 

This particular problem is something we aim to solve with projections that inherently handle these ideas. That's not to say we can't include these methods now though if it would be useful! 

Let me know your thoughts :)

Adam-Hammo:
Thanks @TimRepke, we've discussed as well and agree that a labs module would be a good fit, as a package extra. I'll come back this week with the setup for that so we can get this merged in.

Adam-Hammo (approved):
Looks good to me!

PR #358, opened 2025-11-07
Optimise find by identifiers query

Adam-Hammo:
Staging testing exposed that the `find_with_identifiers` query had performance issues, primarily from the `any()` operator. This takes a subquery approach to optimise the query.


Adam-Hammo:
> Do you think we should be running performance tests at some point in our CI? I am thinking they'd eat a lot of github actions minutes 😞

The vast majority of the time is actually taken in seeding the database, which we can certainly do in a faster manner than what I did here, so it might be achievable in a relatively performant way if we invested a little time into the test setup. It may be more of a concern on the Elasticsearch side than the Postgresql side as well, I don't know how much more complex we'd get at the database in the near future.

Issue #352, opened 2025-11-03
Allow searching for records using a query-string


mootpointer:
Create an endpoint which allows passing a query string which allows full text search of references.

It should allow the searching of title, abstract and author fields. By default the search query should search both title and abstract fields. Individual fields can be added by using the `field:query` syntax. 

The syntax should also allow for the combination of boolean queries, fuzzy matches, and phrase slop/nearby queries.

Responses should be a summary of the record, with a link to the full record retrievable from the `/references` endpoint.

This will require two significant parts:

1. An API endpoint at `/references/search` to handle the requests, which parses the request and properly formats the response
2. Projection of the data into ElasticSearch in a way which will support querying.

jack-o-the-green:
For a first pass, we will implement the following projection of references into elasticsearch 

One title, abstract and authors at the base level of the ReferenceDocument.

Prioritize fields based on the following enhancement selection criteria
* Firstly, prioritize enhancements on the canonical reference
* Secondly, prioritize most recent enhancements (we'll need to add these to the model)

Concretely
* If there's an abstract on the canonical, use that
* If there's two abstracts on the canonical, use the most recent
* If there's no abstracts on the canonical, use the most recent abstract from all duplicates.

Issue #349, opened 2025-10-30
Create a search endpoint on the repository


mootpointer:
Create an endpoint which allows users to retrieve records from a full-text search of the repository.

#### Query params

| param| description| example|
| - | - | -|
|`q`| Full text search query which gets passed through to ElasticSearch| `abstract:climate AND NOT title:reply` |
|`annotation`| A filter to limit records to only include those with a given annotation. Multiple filters can be supplied which will be combined using a boolean AND | `<scheme>:<label>@[<>]score`
|`page`| Page number indexed from 1. 20 records per page | `3` |

---

This work will likely be delivered in stages:

#### Stage 1: Query string all the things
Provide a search endpoint which permits a query string. This will involve projecting the data currently in a highly nested format currently into something flatter to allow this. It would also require some engagement to get the structure of the returned document right (though initially it may be the nested set of documents).

#### Stage 2: Pagination
Layer upon this query api the ability to navigate through pages of results up to a limit of 10,000 records.

#### Stage 3: Filter annotations outside the query string
We see a lot of use cases within the current demonstrator where we filter records based on annotations, and this might be separate to the query string submitted by a user. Enable this through an `annotation` parameter.

#### Stage 4: Filter by published year
Again we see filters by published year, and we see value in having a specific filter for this

#### Stage 5: Sorting
We're not quite sure on what the likely requirements are going to be for sorting beyond relevance, so we've left this for later. It would likely be competing with other streams of work and API endpoints (export/download, cross tab APIs, etc.)

PR #366, opened 2025-11-10
352 query string search api

Adam-Hammo:
Implements API side of #352.

Following this will be PR(s) to add pagination, explicit filters and sorting - see #349.

Documentation looks like:
image
image

### DIFF
--- a/tests/e2e/imports/test_imports.py
+++ b/tests/e2e/imports/test_imports.py
@@ -57,7 +57,12 @@ async def test_happy_import(  # noqa: PLR0913
         )
 
     assert sum(summary["results"].values()) == n_refs
-    assert summary["results"]["completed"] == n_refs
+    try:
+        assert summary["results"]["completed"] == n_refs
+    except AssertionError:
+        # Temporary handler to diagnose flakiness in CI
+        print("Import summary:", summary, flush=True)  # noqa: T201
+        raise
     assert not summary["failure_details"]
 
     # Check the references are in the database


app/domain/base.py
```diff
@@ -32,12 +32,13 @@ def check_serializability(self) -> None:
 )
 
 
-class SQLAttributeMixin(BaseModel):
+class RepositoryIdMixin(BaseModel):
```
Adam-Hammo: Sorry, I didn't mean this to be so pervasive - happy to revert it out. It just struck me that we were using `SQLAttributeMixin` to just imply everything with an ID, even in Elasticsearch contexts, and I whacked the mole.

jack-o-the-green: I like the rename, but perhaps we should just do it in its own PR?

app/domain/references/routes.py
```diff
@@ -205,6 +212,43 @@ async def enhancement_request_hybrid_auth(
...
+        status.HTTP_400_BAD_REQUEST: {
+            "description": "Bad Query String",
+            "model": APIException,
+        }
+    },
+    description="Search for references using a query string in "
+    "[Lucene syntax](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-query-string-query#query-string-syntax)"
+    ". If the query string does not specify search fields, the search will query over "
+    f"[{', '.join(SearchService.default_search_fields)}]. The query string can only "
+    "search over fields on the root level of the Reference document.",
```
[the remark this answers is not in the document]
jack-o-the-green: You can pull the field names out of a pydantic model with `list(ReferenceSearchFields.model_fields)` (the ReferenceSearchFields in introduced in the PR I just raised). So we could use that to generate doc for the available root level search fields.

tests/integration/test_es_references.py
```diff
@@ -647,3 +659,195 @@ async def test_canonical_candidate_search(
...
+    should_match: bool,
+):
+    """Test various Lucene query syntax scenarios."""
+    bibliographic_enhancement: BibliographicMetadataEnhancement = (
+        BibliographicMetadataEnhancementFactory.build(
+            title="test paper and all that",
+            publication_year=2023,
+        )
+    )
+    abstract_enhancement: AbstractContentEnhancement = (
```
jack-o-the-green: Are any of the parameterised tests actually searching the abstract/because the abstract is randomly generated we're not expecting it to generate any hits right?

jack-o-the-green (approved):
LGTM - one query about tests 

Also second musing on whether we want to add some `abstract` search cases now that we can include that in elasticsearch? Happy for that to be in a follow up PR but don't want to miss it!

tests/integration/test_es_references.py
Adam-Hammo: Ah yep that was hanging code from when I forgot we didn't have the projection - now that we do I'll use it in a parameterised test or two

PR #368, opened 2025-11-11
Reference Labs

Adam-Hammo:
Hi @TimRepke, just a couple of ideas we've got on the labs concept, primarily aiming for a clean separation of concerns and a little bit of future-proofing. Let me know what you think :)

### DIFF
--- a/docs/sdk/labs.rst
+++ b/docs/sdk/labs.rst
@@ -20,6 +20,6 @@ There may be additional dependencies required to use some features of this modul
 
 References
 ----------
-.. autoclass:: libs.sdk.src.destiny_sdk.labs.references.ReferenceLabs
+.. autoclass:: libs.sdk.src.destiny_sdk.labs.references.LabsReference
     :members:
     :inherited-members: BaseModel, str, Reference
--- a/libs/sdk/tests/unit/labs/test_references.py
+++ b/libs/sdk/tests/unit/labs/test_references.py
@@ -3,7 +3,7 @@
 import json
 from pathlib import Path
 
-from destiny_sdk.labs.references import ReferenceLabs
+from destiny_sdk.labs.references import LabsReference
 from destiny_sdk.references import Reference
 
 
@@ -11,7 +11,7 @@ def _read_references():
     test_data_path = Path(__file__).parent.parent / "test_data/destiny_references.jsonl"
     with test_data_path.open() as test_references_file:
         for line in test_references_file:
-            yield ReferenceLabs(reference=Reference.from_es(json.loads(line)))
+            yield LabsReference(reference=Reference.from_es(json.loads(line)))
 
 
 def test_annotations():


TimRepke:
Thanks for adding stubs for the documentation and package dependencies for PR #357 

On the changes to the extended `Reference` class: I'm not sure what not using inheritance does here. To me, this just makes it less convenient to use and adds overheads during instantiation (first the normal reference then the labs-reference). Inconvenience is additional imports and deeper nesting when one wants to access the base reference.

libs/sdk/src/destiny_sdk/labs/references.py
```diff
@@ -8,22 +8,29 @@
...
     AnnotationType,
     BibliographicMetadataEnhancement,
     EnhancementType,
 )
 from destiny_sdk.identifiers import ExternalIdentifierType
-from destiny_sdk.references import Reference as BaseReference
+from destiny_sdk.references import Reference
+
 
+class ReferenceLabs(BaseModel):
```
TimRepke: Do we already have a naming style? For me, `LabsReference` is more intuitive—if it needs renaming at all. I guess people can always do `from destiny_sdk.labs.references import LabsReference as Reference` for a drop-in replacement.

```suggestion
class LabsReference(BaseModel):
```

TimRepke (changes requested):
Thanks for adding stubs for the documentation and package dependencies for PR #357

On the changes to the extended Reference class: I'm not sure what not using inheritance does here. To me, this just makes it less convenient to use and adds overheads during instantiation (first the normal reference then the labs-reference). Inconvenience is additional imports and deeper nesting when one wants to access the base reference.

Adam-Hammo:
Thanks @TimRepke, my main motivation for nesting the models was primarily to maintain separation between the stable and experimental concepts. If the stable `Reference` gets a `.title` property graduated, then anybody using `LabsReference` can't access that stable property as it's overwritten by the experimental version. This would allow access to the stable interface with `LabsReference.reference.title`. This could eventually also lead to a violation of the Liskov Substitution Principle.

The premise is that I think a `LabsReference` should not be able to overwrite `Reference` methods - eg it's a `has-a` not a `is-a` relationship. Let me know if you disagree.

libs/sdk/src/destiny_sdk/labs/references.py
Adam-Hammo: No, I'm happy with `LabsReference`

libs/sdk/tests/unit/labs/test_references.py
```diff
@@ -3,14 +3,15 @@
...
+from destiny_sdk.labs.references import ReferenceLabs
+from destiny_sdk.references import Reference
 
 
 def _read_references():
     test_data_path = Path(__file__).parent.parent / "test_data/destiny_references.jsonl"
     with test_data_path.open() as test_references_file:
         for line in test_references_file:
-            yield Reference.from_es(json.loads(line))
+            yield ReferenceLabs(reference=Reference.from_es(json.loads(line)))
```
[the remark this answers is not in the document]
TimRepke: Sure, that would make things more convenient.

TimRepke:
> If the stable `Reference` gets a `.title` property graduated, then anybody using `LabsReference` can't access that stable property as it's overwritten by the experimental version. 

Ok, I see that point. In that case, people could go with `labs_reference.super().title`, but that's absolutely cursed.
Other than that, I'm not sure how relevant this is in our case: In my experience when processing loads of data, every overhead adds up. Let's just see if this ever becomes an issue. If so, we can just change it—it's labs ;)

docs/sdk/labs.rst
```diff
@@ -0,0 +1,25 @@
...
+    uv add destiny-sdk --extra labs
+
+
+.. contents::
+    :depth: 2
+    :local:
+
+References
+----------
+.. autoclass:: libs.sdk.src.destiny_sdk.labs.references.ReferenceLabs
```
TimRepke:
```suggestion
.. autoclass:: libs.sdk.src.destiny_sdk.labs.references.LabsReferece
```

libs/sdk/tests/unit/labs/test_references.py
TimRepke:
```suggestion
            yield LabsReference.from_es(json.loads(line))
```

TimRepke (approved):

PR #370, opened 2025-11-12
Enhancement sorting by prioritisation for projection

jack-o-the-green:
Implemented the enhancement prioritsation for projection described here https://github.com/destiny-evidence/destiny-repository/issues/352#issuecomment-3494108333. 

Also broke down existing tests into smaller, more debuggable tests as I ran into some issues troubleshooting sort when interacting with the big ones.

### DIFF
--- a/app/domain/references/models/projections.py
+++ b/app/domain/references/models/projections.py
@@ -113,7 +113,7 @@ class ReferenceSearchFieldsProjection(GenericProjection[ReferenceSearchFields]):
         cls, canonical_id: uuid.UUID, enhancements: list[Enhancement] | None
     ) -> list[Enhancement]:
         """
-        Order a references enhancements by priority for projecting in reverse order.
+        Order a references enhancements by priority for projecting in increasing order.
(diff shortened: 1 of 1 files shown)


app/domain/references/models/projections.py
```diff
@@ -94,21 +109,41 @@ def get_from_reference(
         )
 
     @classmethod
-    def _priority_sorted_enhancements(
+    def __priority_sorted_enhancements(
         cls, canonical_id: uuid.UUID, enhancements: list[Enhancement] | None
     ) -> list[Enhancement]:
         """
-        Order a references enhancements by priority for projecting.
+        Order a references enhancements by priority for projecting in reverse order.
```
Adam-Hammo:
```suggestion
        Order a references enhancements by priority for projecting in increasing order.
```
I feel this makes it a little more clear?

jack-o-the-green: It most certainly does, thanks for suggesting!

PR #378, opened 2025-11-13
Extend lock duration and make it configurable

Adam-Hammo:
Now that only a handful of tasks are using the locks, this is a bit safer to push out to a very comfortable duration. Was at 3 hours, now 12.


jack-o-the-green (approved):
LGTM

Issue #376, opened 2025-11-12
Stage 3: Annotation Projection


Adam-Hammo:
## Background
Our annotations are currently stored in a deeply nested structure. Nested structures cannot be queried with Elasticsearch query strings. They also require deep contextual knowledge to query otherwise.

Hence, we need to make our annotations queryable by projecting them to some flat structure on the document in Elasticsearch.

## Proposal
**Structure**
Three new fields at the root of the Elasticsearch document:
```
annotations: Array[Keyword] # A list of True annotations in the `scheme/label` format (same as API interface)
evaluated_schemes: Array[Keyword] # A list of schemes that have been evaluated by robots on this reference
inclusion:destiny: Number # The score on the inclusion:destiny annotation
```

With this, we can query:
- positive annotations on the `annotations` field
- negative annotations on `evaluated_schemes` and `NOT annotations`
- unevaluated annotations on `NOT evaluated schemes`
- score domain inclusion annotation on range query

**Examples**
Document:
```
annotations=[taxonomy:exposure/CO2, taxonomy:exposure/Heat]
evaluated_schemes=[taxonomy:exposure, inclusion:destiny]
inclusion:destiny=0.2
```
_(Note that `inclusion:destiny` would be included in `annotations` if it had a score higher than the threshold)_

Elasticsearch Queries:
| Query  |  Hit?  |
|---|---|
| `annotations:taxonomy:exposure/CO2`  | Yes  |
| `annotations:taxonomy:exposure/Pathogens`  | No  |
| `NOT annotations:taxonomy:exposure/Pathogens`  | Yes |
| `annotations:inclusion:destiny` | No |
| `inclusion:destiny:>0.1` | Yes |
| `inclusion:destiny:>0.5` | No |
| `_exists_:evaluated_schemes:taxonomy:outcomes` | No |

API Annotation Filters:
| Parameter | Hit? |
|---|---|
| `taxonomy:exposure/CO2`  | Yes  |
| `taxonomy:exposure/Pathogens`  | No  |
| `inclusion:destiny` | No |
| `inclusion:destiny@0.1` | Yes |
| `inclusion:destiny@0.5` | No |
| `taxonomy:outcomes/Worse` | No |

## Assumptions
- Labels are unique (we know they’re not but we know we’ll fix it)
  - If labels aren't unique and have the same value, there's no actual impact to this assumption
- Score annotations are low-cardinality (generally domain inclusion), not in complex taxonomies
- <1000 domains...

[...]

PR #381, opened 2025-11-14
Annotation projection for search

Adam-Hammo:
Implements #376

### DIFF
--- a/app/domain/references/models/projections.py
+++ b/app/domain/references/models/projections.py
@@ -95,7 +95,9 @@ class ReferenceSearchFieldsProjection(GenericProjection[ReferenceSearchFields]):
                         ]:
                             if key in cls._singly_projected_annotations:
                                 singly_projected_annotations[key] = annotation
+
                         _annotations_by_scheme[annotation.scheme].append(annotation)
+
                     annotations_by_scheme |= _annotations_by_scheme
 
             annotations = cls.__positive_boolean_annotations(annotations_by_scheme)


app/domain/references/models/projections.py
```diff
@@ -63,24 +70,55 @@ def get_from_reference(
...
-    def __normalised_reference_search_fields(
+    def __normalised_reference_search_fields(  # noqa: PLR0913
         cls,
         abstract: str | None,
         authorship: list[destiny_sdk.enhancements.Authorship],
         publication_year: int | None,
         title: str | None,
+        annotations: set[str],
+        evaluated_schemes: set[str],
+        destiny_inclusion_score: float | None,
```
jack-o-the-green: Thinking about these parameter lists getting kinda long - do we want to more away from the projection being made up of entirely class methods and maybe have it have a reference search fields attribute which we normalise when required? Then this method could just be called on `self`

I don't know if this is actually nicer, as we probably want to store the normalised reference search fields anyway, which would involve updating the class attribute in place and that feels a bit messy.

app/domain/references/services/search_service.py
```diff
@@ -58,6 +49,20 @@ def _build_publication_year_query_string_filter(
             f"TO {publication_year_range.end or '*'}]"
         )
 
+    def _build_annotation_query_string_filter(
+        self,
+        annotation: AnnotationFilter,
+    ) -> str:
+        """Build an annotation filter for Elasticsearch query string."""
```
jack-o-the-green: Worth adding a syntax example in the documentation?

app/domain/references/models/projections.py
Adam-Hammo: How do you feel about something like this? https://github.com/destiny-evidence/destiny-repository/pull/381/commits/6e689398e090acf97cb38efc41b5481c98411b0a

jack-o-the-green: Oh very nice!

app/domain/references/models/projections.py
```diff
@@ -63,24 +70,55 @@ def get_from_reference(
...
+                    # performed, as an annotation that was once present that is missing
+                    # in a later enhancement should be treated as removed.
+                    #
+                    # NB this makes the assumption that annotation schemes will always
+                    # be wholly represented in a single enhancement.
+                    _annotations_by_scheme: dict[
+                        str, list[destiny_sdk.enhancements.Annotation]
+                    ] = defaultdict(list)
+                    for annotation in enhancement.content.annotations or []:
+                        if annotation.scheme == "inclusion:destiny":
```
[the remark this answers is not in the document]
Adam-Hammo: Yeah kind of a separation of "hardcoding" from "normal" ES fields. Probably in a later PR? But I'd be okay to implement here if we want

jack-o-the-green (approved):
LGTM

PR #382, opened 2025-11-14
Improved search test hierarchy

Adam-Hammo:
I wasn't particularly happy with the hierarchy I'd implemented for the search testing, so rejigged it here. Summary of the changes:

- Generic `repository` unit tests, not tied to the DESTINY model. These verify our understanding of query strings & associated parameters with Elasticsearch.
- `search_service` tests moved to unit folder. These verify our constructing of the query strings from internal models.
  - In a perfect world these might mock Elasticsearch but really I think it's easier to verify the results of the query than to do complex string comparisons.
- `router` tests thinned out to only those that mock the service calls. These verify our parsing of query parameters.
- `search` integration tests added. These are relatively comprehensive and verify the full flow.

I'll be the first to admit my testing setup is sometimes lacking, so please don't hold back if you have feedback!

### DIFF
--- a/tests/es_utils.py
+++ b/tests/es_utils.py
@@ -72,10 +72,8 @@ async def create_test_indices(client: AsyncElasticsearch):
 async def delete_test_indices(client: AsyncElasticsearch):
     """Delete all indices after tests."""
     for index_alias in index_managers:
-        index_manager = index_managers[index_alias](client)
-        current_index_name = await index_manager.get_current_index_name()
-        if current_index_name:
-            await client.indices.delete(index=current_index_name)
+        for index in await client.indices.get(index=f"{index_alias}*"):
+            await client.indices.delete(index=index)
 
 
 async def clean_test_indices(client: AsyncElasticsearch):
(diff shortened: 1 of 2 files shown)


tests/unit/persistence/es/test_repository.py
```diff
@@ -0,0 +1,367 @@
...
+from elasticsearch.dsl.field import Integer, Text
+from elasticsearch.helpers import async_bulk
+
+from app.core.exceptions import ESQueryError
+from app.domain.base import SQLAttributeMixin
+from app.persistence.es.persistence import GenericESPersistence
+from app.persistence.es.repository import GenericAsyncESRepository
+
+
+class DomainSimpleDoc(SQLAttributeMixin):
```
jack-o-the-green: I implemented something similar here for index manager tests https://github.com/destiny-evidence/destiny-repository/blob/65963c13b55d946b0ba7dfbf88c381373df10532/tests/integration/test_es_index_manager.py#L20

Be worth extracting these out to a conf test and using them in both places, so we have a single dummy doc implementation. I like the increased fields of your docs better than the ones I did, so I think we should keep these ones.

Adam-Hammo: After great pain I've finally made it work 😅 Good shout, and found another weird case while doing so

Issue #339, opened 2025-10-22
Periodic cron job


jack-o-the-green:
- Decide a retry limit 
- Create periodic cron task that searches for PendingEnhancements with their lease up and moves them to expired state
- The cron task should add new PendingEnhancements to the table to replace the expired ones, if the retry limit has not been reached 
- The cron task should populate the `retry_of` field. 

It likely make sense to use a periodic container app job to run the cron

Adam-Hammo:
Noting we might want to consider observability here - both in honeycomb, and maybe some level of alerting if it fails a few times in a row. (We might be able to set up a honeycomb trigger to alert if trace levels drop below a threshold... ugly as that is)
Not a critical path though, this will fail safely in that we'll just have "downtime" for expired enhancements that won't be re-queued.

PR #383, opened 2025-11-17
#339: Add cron job to expire and replace stale pending enhancements

danielbreves:
Part of https://github.com/destiny-evidence/destiny-repository/issues/339

Adds cron job to expire and replace stale pending enhancements. At this stage this only works in development when running the taskiq scheduler (via `scheduler` docker compose profile). I'm planning to set up Azure Container App Jobs for production as a separate task.

### DIFF
--- a/app/domain/references/repository.py
+++ b/app/domain/references/repository.py
@@ -588,6 +588,7 @@ class PendingEnhancementSQLRepository(
             SQLPendingEnhancement,
         )
 
+    @trace_repository_method(tracer)
     async def count_retry_depth(self, pending_enhancement_id: UUID) -> int:
         """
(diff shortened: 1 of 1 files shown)


app/domain/references/repository.py
```diff
@@ -577,6 +588,81 @@ def __init__(self, session: AsyncSession) -> None:
             SQLPendingEnhancement,
         )
 
+    async def count_retry_depth(self, pending_enhancement_id: UUID) -> int:
```
Adam-Hammo: Nit: I think we could trace these methods. I believe there's a `trace_repository_method` decorator or similar

Adam-Hammo (approved):
This looks great!

PR #387, opened 2025-11-17
Add logging to azure service bus listening

Adam-Hammo:
Temporarily increasing observability to whack moles


jack-o-the-green (approved):
LGTM yay observability

Issue #389, opened 2025-11-17
Add raw enhancement type and filter from elasticsearch and robot batch results


jack-o-the-green:
## Background
We should allow arbitrary enhancements on imports, and they should be delivered to robots. We should not allow robots to create arbitrary enhancements. 

Additionally, arbitrary enhancements should not be stored in elasticsearch as any nesting could rapidly expand our indexed fields. 

## Technical scope 
* Add a `raw` enhancement type to our known enhancement content types
* Filter enhancements of this type when projecting to elasticsearch
* Prevent creating enhancements of this type by robots. 

## Drawbacks
These enhancements won't be returned when searching with elasticsearch but as we want to move away from elasticsearch storing entire references anyway, this will be fixed in future #377

PR #391, opened 2025-11-18
Add raw enhancement type 

jack-o-the-green:
Addresses #389 

Dependent on https://github.com/destiny-evidence/destiny-repository/pull/397 being merged first.

This PR 
* Adds the raw enhancement type (currently just allows any fields and marks them all as extra) 
* Prevents robots from creating raw enhancements 
* Prevents raw enhancements from being indexed in elasticsearch

### DIFF
--- a/libs/sdk/src/destiny_sdk/enhancements.py
+++ b/libs/sdk/src/destiny_sdk/enhancements.py
@@ -2,9 +2,9 @@
 
 import datetime
 from enum import StrEnum, auto
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, Literal, Self
 
-from pydantic import UUID4, BaseModel, Field, HttpUrl
+from pydantic import UUID4, BaseModel, Field, HttpUrl, model_validator
 
 from destiny_sdk.core import _JsonlFileInputMixIn
 from destiny_sdk.visibility import Visibility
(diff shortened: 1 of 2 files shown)


app/domain/references/models/es.py
```diff
@@ -98,6 +98,19 @@ class EnhancementContentDocument(GenericNestedDocument):
         Nested(AnnotationDocument, required=False)
     )
 
+    def clean(self) -> None:
+        """
+        Automatically called when saving a document.
```
Adam-Hammo: Is this called by ES? I did a quick google but couldn't find anything.

tests/factories.py
Adam-Hammo: We'll need to add `RawEnhancementFactory()` here also

libs/sdk/src/destiny_sdk/enhancements.py
```diff
@@ -299,12 +301,39 @@ class LocationEnhancement(BaseModel):
...
+    An enhancement for storing raw/arbitrary/unstructured data.
+
+    Data in these enhancements is intended for future conversion into structured form.
+
+    This enhancement accepts any fields that are passed to it. These enhancements cannot
+    be created by robots.
+    """
+
+    enhancement_type: Literal[EnhancementType.RAW] = EnhancementType.RAW
+    source_export_date: datetime.datetime = Field(
```
Adam-Hammo: Should this be optional?

jack-o-the-green: I feel like it shouldn't be - the data enhancements are big and they come out of what appear to be living, changing reviews currently. I think that data export time is an important piece of information for these types of enhancements. 

Happy to be disagreed with though!

PR #393, opened 2025-11-18
Add api auth role assignment for ui user group

Adam-Hammo:
Gives any UI users the same access on the API.

Group may need to be renamed on the future, but it might be best to let it take shape first.

### DIFF
--- a/infra/app/auth.tf
+++ b/infra/app/auth.tf
@@ -207,6 +207,13 @@ resource "azuread_app_role_assignment" "developer_to_auth" {
   resource_object_id  = azuread_service_principal.destiny_repository_auth.object_id
 }
 
+
+resource "azuread_app_role_assignment" "ui_users_to_auth" {
+  app_role_id         = "00000000-0000-0000-0000-000000000000"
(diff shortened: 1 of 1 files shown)


infra/app/auth.tf
```diff
@@ -207,6 +207,13 @@ resource "azuread_app_role_assignment" "developer_to_auth" {
   resource_object_id  = azuread_service_principal.destiny_repository_auth.object_id
 }
 
+
+resource "azuread_app_role_assignment" "ui_users_to_auth" {
+  app_role_id         = "00000000-0000-0000-0000-000000000000"
+  principal_object_id = var.ui_users_group_id
+  resource_object_id  = azuread_service_principal.destiny_repository_auth.object_id
```
danielbreves: It could be worth considering whether we may want a different api access for UI users, since destiny_repository_auth has quite [broad permissions](https://github.com/destiny-evidence/destiny-repository/pull/393/files#diff-d43436ee691123b4a1be741d202aded4e2c23e8863cae12a26e19b115ff0ffb0R172-R184) at the moment.

Issue #400, opened 2025-11-24
Add ERIC id option to external identifiers


jack-o-the-green:
Allow References to be identified by their ERIC ids 

https://eric.ed.gov/?id=ED581143

jack-o-the-green:
**Note for us** 
ERIC ID's are called "ERIC Numbers" by ERIC, which has a separate concept of identifier as "standardized proper nouns and a controlled vocabulary for ERIC. ERIC identifiers, which were improved in 2016, can be used as limiters to increase the precision of searches for research associated with laws, assessments, and locations."

For example, the record https://eric.ed.gov/?id=ED581143 has `Identifiers - Location: United Kingdom (England)`

I think it still makes sense to call our external identifier and identifier, as that's within our domain. But felt this was useful context.

PR #408, opened 2025-11-25
Adding ERIC Number external identifiers

jack-o-the-green:
Adding ERIC identifiers

Eric number vs eric identifier context here https://github.com/destiny-evidence/destiny-repository/issues/400#issuecomment-3573542729

https://github.com/destiny-evidence/destiny-repository/issues/400

### DIFF
--- a/ui/components/forms/IdentifierTypeSelect.tsx
+++ b/ui/components/forms/IdentifierTypeSelect.tsx
@@ -2,6 +2,7 @@
 
 const options = [
   { value: "doi", label: "DOI" },
+  { value: "eric", label: "ERIC Number" },
   { value: "open_alex", label: "OpenAlex" },
   { value: "pm_id", label: "Pubmed ID" },
   { value: "other", label: "Other" },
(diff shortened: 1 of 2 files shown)


libs/sdk/src/destiny_sdk/identifiers.py
```diff
@@ -47,6 +51,32 @@ def remove_doi_url(cls, value: str) -> str:
...
+class ERICIdentifier(BaseModel):
+    """
+    An external identifier representing an ERIC Number.
+
+    An ERIC Number is defined as a unqiue identifiying number preceeded by
+    EJ (for a journal article) or ED (for a non-journal document).
+    """
+
+    identifier: str = Field(
+        description="The ERIC Number of the refrence.", pattern=r"E[D|J][0-9]+$"
```
Adam-Hammo:
```suggestion
        description="The ERIC Number of the reference.", pattern=r"E[D|J][0-9]+$"
```

Adam-Hammo (approved):
LGTM, one typo nit.

There's one other place I think we list identifiers, on the UI dropdown - don't have a strong feeling either way to whether we should update it. I think you'd still be able to search it in the batch textbox with `eric:ED123456`

PR #414, opened 2025-11-27
Fix rabbitmq port

Adam-Hammo:
Rework expiry task to be called directly on the `worker` container.

`fix 1` commit shows an alternate approach with a working broker, but I'm a little cagey having moving parts in the e2e tests that aren't containerised like the rest.

Could possibly also be solved by creating a docker network fixture for e2e tests and using that on each container - this would have to be exposed to the local machine though for e2e tests to call the API.

Error message before fix:
```
docker.errors.APIError: 500 Server Error for http+docker://localhost/v1.48/containers/3fe9fddc6d4c341e8812e571737f02e553d04130693aed1716c1628a86fc8673/start: Internal Server Error ("failed to set up container networking: driver failed programming external connectivity on endpoint e2e-rabbitmq (a7d839e5f93c755d9b81e2e3561cf0a31641757e2219b487c313143d3d69c446): Bind for 0.0.0.0:5672 failed: port is already allocated")
```


danielbreves:
This is much nicer. Thanks!

PR #426, opened 2025-12-07
402/improve eppi parsing

jack-o-the-green:
Updates to the EPPI parsing out the back of receiving more EEF data to play with. 

* Updates to prefix stripping on identifiers
* Parser now creates a seperate jsonl file containing the references it failed to parse.

### DIFF
--- a/libs/eppi_importer/import_from_eppi.py
+++ b/libs/eppi_importer/import_from_eppi.py
@@ -126,10 +126,10 @@ def main() -> None:
     with Path(args.output).open("w") as f:
         f.writelines(ref.to_jsonl() + "\n" for ref in references)
 
-    failed_refs_path = args.output.removesuffix(".jsonl") + "-failures.jsonl"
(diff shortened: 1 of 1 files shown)


libs/eppi_importer/import_from_eppi.py
```diff
@@ -117,6 +125,11 @@ def main() -> None:
     with Path(args.output).open("w") as f:
         f.writelines(ref.to_jsonl() + "\n" for ref in references)
 
+    failed_refs_path = args.output.removesuffix(".jsonl") + "-failures.jsonl"
+
+    with Path(failed_refs_path).open("w") as f:
+        f.writelines(json.dumps(ref, ensure_ascii=False) + "\n" for ref in failed_refs)
```
jack-o-the-green: I'm toying with whether it's worth dumping it out straight back into eppi format - i.e. as json like 

```
{ "References": [
    { "Codes": [], "Outcomes": [] ... }
    { "Codes": [], "Outcomes": [] ... }
]}
```
So we could biff it back through this parser in future. Thoughts?

PR #439, opened 2025-12-16
Search UI

Adam-Hammo:
Simple search UI for exploring the repository

image
image
image

### DIFF
--- a/tests/unit/persistence/es/test_repository.py
+++ b/tests/unit/persistence/es/test_repository.py
@@ -200,9 +200,17 @@ async def test_query_string_search_with_fields(
     assert len(results.hits) == 1
     assert str(results.hits[0].id) == doc_id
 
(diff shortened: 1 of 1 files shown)


app/persistence/es/repository.py
```diff
@@ -223,8 +223,7 @@ async def search_with_query_string(
         """
         trace_attribute(Attributes.DB_QUERY, query)
         search = (
-            AsyncSearch(using=self._client)
-            .doc_type(self._persistence_cls)
```
Adam-Hammo: Turns out `doc_type` is a [long deprecated interface](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html), not that you'd know that from the Elasticsearch python mappings. This was leading to weird duplicates when searching. Using index directly resolved these.

PR #442, opened 2025-12-17
Add telemetry to e2e tests

Adam-Hammo:
This will run automatically in github actions, and can be enabled locally. Should help us diagnose flakey errors and other issues.

https://ui.honeycomb.io/destiny-evidence/environments/github-actions/datasets/e2e-runner-test/result/qhLpYZFXdGi/trace/2q9UjmXSDYP?fields[]=s_name&fields[]=s_serviceName&span=35f1454d8a0530cd

### DIFF
--- a/app/domain/references/repository.py
+++ b/app/domain/references/repository.py
@@ -274,8 +274,7 @@ class ReferenceESRepository(
         :rtype: list[ESScoreResult]
         """
         search = (
-            AsyncSearch(using=self._client)
-            .doc_type(self._persistence_cls)
+            AsyncSearch(using=self._client, index=self._persistence_cls.Index.name)
             .query(
                 Q(
                     "bool",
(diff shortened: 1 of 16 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -38,4 +38,8 @@ jobs:
         run: uv sync --locked --group test
 
       - name: Run test with pytest via uv
-        run: DOCKER_HOSTNAME=172.17.0.1 TESTS_USE_RABBITMQ=true uv run pytest tests/e2e --build
+        env:
```
jack-o-the-green: I might be missing something here, but I can't see where the github actions environment is configured for the E2E tests.

Adam-Hammo: Good point, let me try in dev as well as staging

PR #447, opened 2026-01-08
Add configurable timeouts to oauth client

Adam-Hammo:
@L-ENA noted that some searches would time out when using the SDK. This PR adds both a client-level and request-level timeout to the SDK client to allow longer-running searches.


danielbreves (approved):

L-ENA:
Thank you for implementing this!

PR #452, opened 2026-01-12
Reference duplication tree depth is now a constant

jack-o-the-green:
Removed the settings access from the sql layer of the application, as this was still importing the full settings object when running alembic migrations. 

VSCode env variable caching got me the first time around.


Adam-Hammo (approved):
Nice!

PR #453, opened 2026-01-13
Remove localhost redirect uri from auth_v2

jack-o-the-green:
We don't want this redirect uri. Context here https://covidence.slack.com/archives/C0810ARPCRG/p1768255530034719?thread_ts=1767683955.758679&cid=C0810ARPCRG.


jack-o-the-green:
I've gone ahead and removed all of the public client redirects from the external directory ui client, as these were only for testing and are also causing state applies with each plan/apply.

Issue #416, opened 2025-11-27
Discovery: dependencies for ingesting all of OpenAlex


Adam-Hammo:
# Background
We're looking to ingest "all" of OpenAlex (up to 450mio records). This is a ~100x increase in data and requires some planning to get ahead of.

✅: Done
🚧: In progress
❗: Needs starting
❌: Won't do now
❓: Needs decision/clarification
Anything without an emoji needs categorising

# Requirements
## Pre-requisite
**Items that must be implemented before we begin mega-ingestion, or that we have otherwise decided should be**
### 🚧 Infrastructure scale
- ❓What scale do we want to maintain in development and staging?
- [Budgeting sheet here](https://docs.google.com/spreadsheets/d/1OZ45QuX2Raun5mKeKAoDv5W-oS_6fNYU9qYY2DZMNIM/edit?usp=sharing)
- Postgres should be relatively straightforward
  - Template PR: #457 
- Elasticsearch
  - Sharding in migrations and templated scale here: #456
  - ❓ #377 opens the possibility to smaller document sizes. I scoped at https://github.com/destiny-evidence/demonstrator-ui/pull/42 what it might look like for the UI to handle this, and it seems tenable.
    - This might be too much for pre-ingestion but could allow us to downsize in the future and justify the initial cost. We should get a >50% reduction in individual document size.
  - ❓ Should we consider serverless? Pricing is opaque, might be cheaper, might be much more expensive, but should be easier to manage. 
- Worker replicas should probably be increased while ingesting (current max is 10)

### ❓ Importer
- ❓This might be trivial, but do we have a concrete plan for who/what will do the importing? Is it possible for the incremental importer to handle this by running on historical dates?

### ❗ Trace sampling
- See #464

### ✅ Binit containers
- See #181 

### 🚧 Pending Database Changes
- Changes we need to make before scale makes them harder
- Elasticsearch
  - 🚧 Index analyzers for deduplication searching: #455
  - 🚧 Index analyzers for general searching (prefix analyzers?): #ticket-me
- Postgres
  - 🚧 UUID4 -> UUID7 #317  
    - NB only #448 is required before ingestion.
  - 🚧 v16->v18 #454 
  - 🚧 Enums (dependent on #181)
    - #422 

### ✅ New Enhancement Types
- Adding new enhancements to the repository to allow the bulk import to populate them
- ✅ Repository-side
  - ✅ Added:
    - #432
      - NB only #436 required before ingestion
[...]

PR #454, opened 2026-01-13
Upgrade to postgres v18

### LINKED ISSUE #517: ⛔ Upgrade Postgres to v18

Adam-Hammo:
Taking advantage of diligence around other upgrades at #316 to upgrade our postgres ahead of #416.

"Should" take <15 minutes of downtime to upgrade. PIT rollbacks will work to downgrade if required. Requires ~20% free storage, which we have.

https://learn.microsoft.com/en-us/azure/postgresql/configure-maintain/concepts-major-version-upgrade

Image


jack-o-the-green (approved):
LGTM - though this will need updating as the database terraform was recently split into its own file

Issue #163, opened 2025-06-23
Improve indexing performance


Adam-Hammo:
Elasticsearch indexing currently occurs at ~15,000 records per minute. This is a hindrance to us being able to iterate over index structures as it takes a couple of hours to rebuild an index completely.

Probably not the highest priority, but would be pretty simple to split out the indexing into multiple task workers. The parent task would recreate the index, and then provide ordered `reference_id` offsets to _n_ tasks that then call `reference_service.index_references` on their chunk.

Adam-Hammo:
We have sharding per #456 so parallelisation will be very helpful here.

PR #456, opened 2026-01-14
Specify Elasticsearch shards in a migration

Adam-Hammo:
Adding sharding to support #416.

**Ballpark**
At present in production we have 8 million documents taking up 46.35GB across two shards (a single primary shard and a single replica shard). An upper bound for the number of documents we expect is 500 million, so we can expect around 2.9TB (let's call it 3).

**Shards**
This PR adds the ability to specify the number of shards when performing index migrations.

Elasticsearch recommends [10-50GB per shard](https://www.elastic.co/docs/deploy-manage/production-guidance/optimize-performance/size-shards). Striking down the middle that's 30GB per shard, at 1.5TB per index, we should initially set this to 50 shards at a minimum (open to thoughts on if we should go higher, eg 100 shards to 15GB per).

**Next**
Reindexing needs to be parallelised to take advantage of the sharding, though this isn't a strict blocker for ingestion, just for our first reindex post-ingestion. #163


mootpointer:
The other option would be to consider [Elastic Cloud Serverless](https://www.elastic.co/cloud/serverless) which might end up costing less in total (though costs would obviously spike while we're ingesting large amounts)

PR #459, opened 2026-01-15
Migrate from init containers to container app job for database migrations: Part I

jack-o-the-green:
Implements the majority of #181 

### Background
Replaces the init containers with a database migration container app job that is updated automatically to the same image as the app when deploying. Migrations now run on a single replica and do not retry in the event of failure. 

We are now able to migrate (hah) away from using the environment tags on our docker images, and work purely with the commit sha. 

### When migration happens now 
Previously db migrations used to be applied by the init containers _before_ the new replica started up, while the old replica was running. This pipeline deploys the new replica and _then_ runs the migrations. I made this decision because we can make our code backwards compatible with old versions of the database, which we will want to do for the majority of migrations. 

Happy run shown [here](https://github.com/destiny-evidence/destiny-repository/actions/runs/21016071769) as well as [honeycomb logs](https://ui.honeycomb.io/destiny-evidence/environments/development/datasets/db-migrator-development/result/7bEPhnJB631)

Migrations are also capped at 8 hours, there may be specific cases where we need to update this. 

### Downgrading migrations
This can now be done by manually triggering the container app job with a downgrade command, huzzah! 

### Bugfix included
The promote.yml workflow was not updating the scheduled container app jobs images as the staging/dev deployment was. This previously worked because the `production` image tag used by the scheduled jobs was updated to the latest deployed image for the init container. Since we're not doing that anymore, I've added the scheduled jobs image updates to the promote workflow. 

There's a reasonable amount of duplication between our promote and deploy flows now, I will raise a PR after this one to address this and avoid future similar mistakes. This will also allow us to test the github action outside of using it to promote to production. 

### Not covered in this PR 
You'll notice we're still using the credentialed url to connect to the database on the container app job. Our alembic stuff needs updating to work with managed identities (e.g. token generation and then db url construction). We typically hit some delays/snags in this area, so I've split this work into two as this first portion allows us to move forward with enum migrations on large tables without the manual finangling previously required to do single replica deployments.

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -93,7 +93,7 @@ jobs:
             az containerapp job update \
               --name "db-migrator-${{ vars.ENVIRONMENT_NAME }}" \
               --resource-group ${{vars.RESOURCE_GROUP}} \
-              --image ${{vars.REGISTRY_NAME}}.azurecr.io/${{vars.APP_NAME}}:${{ needs.deploy-production.outputs.image_tag }}
+              --image ${{vars.REGISTRY_NAME}}.azurecr.io/${{vars.APP_NAME}}:${{ needs.deploy-production.outputs.image_tag }} \
               --command alembic upgrade head
       - name: Trigger database migrations
         uses: Azure/cli@v2.0.0
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -195,7 +195,7 @@ jobs:
             az containerapp job update \
               --name "db-migrator-${{ vars.ENVIRONMENT_NAME }}" \
               --resource-group ${{vars.RESOURCE_GROUP}} \
-              --image ${{vars.REGISTRY_NAME}}.azurecr.io/${{vars.APP_NAME}}:${{ needs.build.outputs.image_tag }}
+              --image ${{vars.REGISTRY_NAME}}.azurecr.io/${{vars.APP_NAME}}:${{ needs.build.outputs.image_tag }} \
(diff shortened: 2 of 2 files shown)


jack-o-the-green:
You can ignore the Terraform failure - I accidentally had this configured to run against my infra-sandbox environment. When it reruns it'll skip that one.

.github/workflows/promote.yml
```diff
@@ -97,3 +74,31 @@ jobs:
...
+          tenant-id: ${{ vars.AZURE_TENANT_ID }}
+          subscription-id: ${{ vars.AZURE_SUBSCRIPTION_ID }}
+      - name: Deploy built image to database migrator job
+        uses: Azure/cli@v2.0.0
+        with:
+          inlineScript: |
+            az containerapp job update \
+              --name "db-migrator-${{ vars.ENVIRONMENT_NAME }}" \
+              --resource-group ${{vars.RESOURCE_GROUP}} \
+              --image ${{vars.REGISTRY_NAME}}.azurecr.io/${{vars.APP_NAME}}:${{ needs.deploy-production.outputs.image_tag }} \
```
Adam-Hammo: This is really nice

Adam-Hammo (approved):
Looks good! Really keen for this

infra/app/main.tf
```diff
@@ -227,24 +210,6 @@ module "container_app" {
     }
   }
 
-  init_container = {
```
Adam-Hammo: Don't let the door hit you on the way out

danielbreves (approved):

PR #460, opened 2026-01-15
Specify trusted identifier when testing exact deduplication

Adam-Hammo:
Fixes a flake in our unit tests where if all generated identifiers are of type Other, exact deduplication won't run (as it needs at least one non-other identifier).

### DIFF
--- a/tests/unit/domain/references/services/test_deduplication_service.py
+++ b/tests/unit/domain/references/services/test_deduplication_service.py
@@ -35,6 +35,18 @@ def reference() -> Reference:
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


jack-o-the-green:
I think the same flake is possible in `test_find_exact_duplicate_happy_path` because it uses this reference fixture https://github.com/destiny-evidence/destiny-repository/blob/62b7f2bd63671bb195f0ad853c49eaeadb1f1519/tests/unit/domain/references/services/test_deduplication_service.py#L33

jack-o-the-green (approved):
LGTM - thanks!

PR #462, opened 2026-01-16
Add terraform precommit hooks

jack-o-the-green:
Added linting, and formatting precommit hooks for terraform. This sets us up well to be able to do an upgrade on all provider versions that we're using, as well as an upgrade of the terraform version itself (runs fine on v1.14.3 so we want to pin to that as minimum)

Eventually we'd like to expand from the `recommended` to `all` rulsets for tflint, however we have resources that break naming conventions, which requires manipulation of the state via `moved {}` blocks to fix, which is a more potential-to-break activity and can follow on in it's own isolated PR. 

This turned up some variables we're not using, so that's great!

Note: I tried adding terraform validate but there's some issues with this hook their workarounds did not solve github.com/antonbabenko/pre-commit-terraform?tab=readme-ov-file#terraform_validate. Doesn't feel like a priority to work through right now, so just took it out again.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,4 +57,3 @@ repos:
     hooks:
       - id: terraform_tflint
       - id: terraform_fmt
-      - id: terraform_validate


infra/app/terraform.tf
```diff
@@ -45,6 +45,11 @@ terraform {
       source  = "honeycombio/honeycombio"
       version = "0.37.0"
     }
+
+    random = {
+      source  = "hashicorp/random"
+      version = "3.7.1"
+    }
```
Adam-Hammo: What's this needed for?

Adam-Hammo: Yeah fair enough

PR #463, opened 2026-01-16
560: Keycloak Auth

### LINKED ISSUE #560: Add support for Keycloak auth to DESTINY repo

danielbreves:
Introduces support for Keycloak as an alternative auth provider to Azure AD. 

* Added a new `KeycloakJwtAuth` class that is used instead of `AzureJwtAuth` when `auth_provider` is set to `keycloak`.
* Added keycloak as an optional container to docker-compose so we can test it locally
* Added a new `ROBOT_WRITER` role to the `AuthRole` enum for consistency.
* Updated CLI authentication in `cli/auth.py` to support both Azure and Keycloak providers, using the `auth_provider` setting.
* Added a KeycloakOAuthMiddleware to the SDK as an alternative to the Azure OAuthMiddleware.

Related to https://github.com/destiny-evidence/destiny-repository/issues/217
Fixes #560

### DIFF
--- a/keycloak/local-destiny-realm.json
+++ b/keycloak/local-destiny-realm.json
@@ -14,14 +14,6 @@
   "ssoSessionIdleTimeout": 1800,
   "ssoSessionMaxLifespan": 36000,
   "offlineSessionIdleTimeout": 2592000,
-  "defaultOptionalClientScopes": [
-    "administrator.all",
-    "import.writer.all",
-    "reference.reader.all",
-    "reference.deduplicator.all",
-    "enhancement_request.writer.all",
-    "robot.writer.all"
-  ],
   "roles": {
     "realm": [
       {
@@ -143,14 +135,6 @@
         "https://oauth.pstmn.io/v1/callback"
       ],
       "webOrigins": ["http://localhost:3000", "https://oauth.pstmn.io"],
-      "defaultClientScopes": [
-        "openid",
-        "profile",
-        "email",
-        "roles",
-        "web-origins",
-        "reference.reader.all"
-      ],
       "optionalClientScopes": [
         "administrator.all",
         "import.writer.all",
--- a/tests/e2e/auth/keycloak-import/test-destiny-realm.json
+++ b/tests/e2e/auth/keycloak-import/test-destiny-realm.json
@@ -14,14 +14,6 @@
   "ssoSessionIdleTimeout": 1800,
   "ssoSessionMaxLifespan": 36000,
   "offlineSessionIdleTimeout": 2592000,
-  "defaultOptionalClientScopes": [
-    "administrator.all",
-    "import.writer.all",
-    "reference.reader.all",
-    "reference.deduplicator.all",
-    "enhancement_request.writer.all",
-    "robot.writer.all"
-  ],
   "roles": {
     "realm": [
       {


app/api/auth.py
```diff
@@ -18,6 +18,8 @@
 from uuid import UUID
 
 import destiny_sdk
+from authlib.jose import JsonWebKey, JsonWebToken
```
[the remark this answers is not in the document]
mootpointer: Also worth noting that `authlib` is transitively depended on  through `destiny-sdk`, but not explicitly listed as a dependency for `destiny-repository`

app/api/auth.py
```diff
@@ -396,6 +399,230 @@ async def __call__(
...
+
+        # Set telemetry attributes
+        span = trace.get_current_span()
+        span.set_attribute(Attributes.USER_AUTH_METHOD, "keycloak-jwt")
+
+        if sub := verified_claims.get("sub"):
+            span.set_attribute(Attributes.USER_ID, sub)
+        if name := verified_claims.get("name"):
+            span.set_attribute(Attributes.USER_FULL_NAME, name)
+        if email := verified_claims.get("email"):
```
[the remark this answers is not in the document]
jack-o-the-green: Ideally we'd use an id or something

app/core/config.py
```diff
@@ -365,8 +365,39 @@ class Settings(BaseSettings):
...
+    keycloak_issuer_url: str | None = Field(
+        default=None,
+        description=(
+            "The issuer URL for token validation. Defaults to keycloak_url if not set. "
+            "Useful when the token issuer differs from the internal Keycloak URL "
+            "(e.g., in Docker where tokens are issued with localhost but JWKS is "
+            "fetched via internal network)."
+        ),
+    )
+    keycloak_realm: str = Field(
```
jack-o-the-green: This is probably a wider question of what should be configurable. Do we want a situation where the keycloak realm can be called something other than `destiny`, can we think of a scenario where that would change?

keycloak/destiny-realm.json
```diff
@@ -0,0 +1,236 @@
...
+  },
+  "groups": [
+    {
+      "name": "consumers",
+      "realmRoles": ["reference.reader"]
+    },
+    {
+      "name": "developers",
+      "realmRoles": [
+        "administrator",
```
jack-o-the-green: add `robot.writer`?

keycloak/destiny-realm.json
```diff
@@ -0,0 +1,236 @@
...
+      "publicClient": false,
+      "bearerOnly": true,
+      "standardFlowEnabled": false,
+      "implicitFlowEnabled": false,
+      "directAccessGrantsEnabled": false,
+      "serviceAccountsEnabled": false,
+      "protocol": "openid-connect"
+    }
+  ],
+  "users": [
```
jack-o-the-green: Might be misunderstanding the realm here - but does this get loaded in to every environment? If so, we shouldn't have the test user here. I'm assuming not but wanted to double check!

tests/e2e/auth/keycloak-import/destiny-realm.json
```diff
@@ -0,0 +1,224 @@
...
+    "import.writer.all",
+    "reference.reader.all",
+    "reference.deduplicator.all",
+    "enhancement_request.writer.all",
+    "robot.writer.all"
+  ],
+  "roles": {
+    "realm": [
+      {
+        "name": "administrator",
```
jack-o-the-green: Do we want the robot roles in here as well? I think I might be missing how the below test for creating a robot is passing without configuration of the robot bits in the realm. Can you run me through what I'm missing?

keycloak/destiny-realm.json
jack-o-the-green: I think I've answered my own question looking at the realm PR on destiny-shared-infra :D

keycloak/destiny-realm.json
```diff
@@ -0,0 +1,236 @@
...
+  "resetPasswordAllowed": true,
+  "editUsernameAllowed": false,
+  "bruteForceProtected": true,
+  "accessTokenLifespan": 300,
+  "accessTokenLifespanForImplicitFlow": 900,
+  "ssoSessionIdleTimeout": 1800,
+  "ssoSessionMaxLifespan": 36000,
+  "offlineSessionIdleTimeout": 2592000,
+  "defaultOptionalClientScopes": [
+    "administrator.all",
```
mootpointer: I don't know that we need to create client scopes for every role, or if we can have one scope that has role mappings – I'm still learning how to KeyCloak properly.

mootpointer (commented):
I've been crunching on how our current roles and scopes work, and Claude got grumpy at me when I asked about how to think about them so... Not sure whether we want to deal with this separately or from the get-go as we move to Keycloak: https://claude.ai/share/85677599-a31c-4860-99b1-f8217c581bd0

libs/sdk/src/destiny_sdk/client.py
```diff
@@ -448,6 +449,117 @@ def auth_flow(
                 yield request
 
 
+class KeycloakOAuthMiddleware(httpx.Auth):
```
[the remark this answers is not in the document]
danielbreves: Would you mind if we defer it until we've had a chance to test this ourselves in production? Since we don't have SSO setup it's probably not quite ready for external consumers.

tests/e2e/auth/keycloak-import/test-destiny-realm.json
```diff
@@ -0,0 +1,224 @@
...
+    "import.writer.all",
+    "reference.reader.all",
+    "reference.deduplicator.all",
+    "enhancement_request.writer.all",
+    "robot.writer.all"
+  ],
+  "roles": {
+    "realm": [
+      {
+        "name": "administrator",
```
danielbreves: Great question, I was under the impression I had included the scope in the developers group, but it seems I didn't! And the reason it passes seems to be because [Keycloak allows any user to request a scope if it doesn't have any role scope mappings](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/services/util/DefaultClientSessionContext.java#L281-L284) (thanks Claude for finding this)!

keycloak/local-destiny-realm.json
```diff
@@ -0,0 +1,236 @@
...
+  "resetPasswordAllowed": true,
+  "editUsernameAllowed": false,
+  "bruteForceProtected": true,
+  "accessTokenLifespan": 300,
+  "accessTokenLifespanForImplicitFlow": 900,
+  "ssoSessionIdleTimeout": 1800,
+  "ssoSessionMaxLifespan": 36000,
+  "offlineSessionIdleTimeout": 2592000,
+  "defaultOptionalClientScopes": [
+    "administrator.all",
```
danielbreves: I've created a separate issue for this here: https://github.com/destiny-evidence/destiny-shared-infra/issues/20

tests/e2e/auth/keycloak-import/test-destiny-realm.json
jack-o-the-green: Huh that's interesting and good to know! I would have thought the opposite would be more correct.

tests/e2e/auth/conftest.py
```diff
@@ -0,0 +1,238 @@
...
+    Args:
+        keycloak_url: Base URL for Keycloak (e.g., http://localhost:8080)
+        scopes: Space-separated list of scopes to request
+
+    Returns:
+        The access token string
+
+    """
+    token_url = f"{keycloak_url}/realms/destiny/protocol/openid-connect/token"
+    max_retries = 30
```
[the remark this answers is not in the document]
danielbreves: Good catch, I don't need this all, it was left over from debugging.

Issue #422, opened 2025-12-04
Migrate away from postgres enums


jack-o-the-green:
## Background
Due to issues with adding new values to postgres enum types we've decided to move away from them https://github.com/destiny-evidence/destiny-repository/pull/343#discussion_r2458931314. Some of that work was already completed in #397. 

We want to clean up the remaining enums and replace then with varchar fields so that we're doing things in a consistent manner. Each one will need to be applied as a seperate migration in it's own PR/ 

## Remaining enums (sizes at time of ticket creation, check before migrating!)
These are the remaining enums (or 5 but visibility is used in two places!).
* Import record status, table less than 1GB no trouble
* Visibility on reference, table less than 1GB no trouble (it's like 870MB)
* Enhancement request status, table less than 1GB no trouble
* Duplicate determination on reference duplicate decision, table less than 1GB no trouble
* Visibility as an enum on enhancements. Prod enhancements table is 20.3GB (inc indices). We're using 43% of the storage capacity of prod currently, expect doing a migration on this to raise us to min 75% storage capacity but still appears doable at this db size

## Already migrated 
 ✅  Import result status on import result, table less than 1GB no trouble

PR #476, opened 2026-01-19
Increase the database size to allow enum removal migrations on the enhancement table

jack-o-the-green:
Required as part of #422 to remove the visibility enum due to size of the `enhancement` table. Required in staging as well as production because staging currently has the largest database. 

Storage tier increased to the minimum supported storage tier for the 128GB database size (https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/postgresql_flexible_server#storage_tier-defaults-based-on-storage_mb)

### DIFF
--- a/infra/app/database.tf
+++ b/infra/app/database.tf
@@ -1,13 +1,13 @@
 locals {
   database_migrator_name = "db-migrator-${var.environment}"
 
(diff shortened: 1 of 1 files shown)


Adam-Hammo (approved):
LGTM

PR #493, opened 2026-01-20
Fix e2e test flake

Adam-Hammo:
Fixes an e2e test flake seen here: https://ui.honeycomb.io/destiny-evidence/environments/github-actions/datasets/e2e-runner-test/result/jRzgX6TYjUn/trace/6kZai4czkP4?fields[]=s_name&fields[]=s_serviceName&span=4604367cd084af08

The API-triggered deduplication task was being queued inside the SQL UOW, so there was a rare race condition where the task can be received and attempted before the actual object it operates on is committed.


jack-o-the-green (approved):
LGTM

PR #497, opened 2026-01-20
Refactor eppi parser to handle single references.

### LINKED ISSUE #458: refactor eppi parser to have non-private methods for parsing the whole file (as it currently exists) as well as just a single reference (we need this functionality in `deet`)
https://github.com/destiny-evidence/destiny-repository/blob/984de1d7d41478e7c9fa45bd96d6a97e248731ca/libs/sdk/src/destiny_sdk/parsers/eppi_parser.py#L213

would basically re-jig the loop to loop over one reference and run a `parse_one_reference` method or something like that. right now we are adding functionality that uses all the private methods defined in this parser in `deet`.

i am happy to implement this and send a PR if this is OK.

nikloynes:
first off, apologies for the long branch name. i went straight from the issue and didn't think. 

this PR adds a method (`parse_reference`) for parsing a single EPPI reference, as opposed to the whole parsed dict resulting from eppi.json with n references. this is a requirement for `deet`, where we want to attach citations in destiny format to a given reference from an eppi json. 

also added a new method (`parse_full_eppi`) which replaces the existing `parse_data` method. `parse_data` is retained, wrapping around `parse_full_eppi`, but could be deprecated. this will require changing the tests to reference the new method rather than the old one. 

Closes #458

### DIFF
--- a/libs/sdk/src/destiny_sdk/enhancements.py
+++ b/libs/sdk/src/destiny_sdk/enhancements.py
@@ -5,7 +5,7 @@ import json
 from enum import StrEnum, auto
 from typing import Annotated, Any, Literal, Self
 
-from pydantic import UUID4, BaseModel, Field, HttpUrl, model_validator
+from pydantic import UUID4, BaseModel, ConfigDict, Field, HttpUrl, model_validator
 
 from destiny_sdk.core import _JsonlFileInputMixIn
 from destiny_sdk.identifiers import Identifier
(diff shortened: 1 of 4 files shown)


nikloynes:
@Adam-Hammo set back to draft to fix tests failing in CI

Adam-Hammo:
@nikloynes is this still needed?

nikloynes:
@Adam-Hammo apologies, i completely forgot to fix this after it failed an errand integration test. let me give it another go before going on leave on tues, and i'll ping you for review and you can 👍 or 👎.

nikloynes:
@jack-o-the-green also missed your comment, apologies. this is to validate and populate incoming eppi data in `deet` to destiny citation format.

PR #518, opened 2026-01-28
504: Add test for cross field match search query

### LINKED ISSUE #504: Validate the behaviour of cross matching queries for title & abstract searching
We want to be able to use a **query string** to search across both title and abstract fields. We need to confirm that the behaviors matches the cross_fields behavior in **match queries**

> The cross_field type tries to solve these problems at query time by taking a term-centric approach. It first analyzes the query string into individual terms, then looks for each term in any of the fields, as though they were one big field. 

> for instance, when querying the first_name and last_name fields for Will Smith, the best match is likely to have Will in one field and Smith in the other.

**Match doc**
https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-multi-match-query#type-cross-fields

**Query string doc** Note the blended boolean queries.  
https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-query-string-query#query-string-min-should-match-cross

**Technical Scope**
Just write some tests that demonstrate the behaviour, and if it behaves as desired, raise a ticket for supporting cross_fields behavior in our search api, or if it's already possible document it and use it for the default abstract + title searching. 

eg. If I have a doc like this
```
{
   "first_name": "George",
   "last_name": "Harrison"
}
```

I should be able to match it with the query string `george AND Harrison` with `cross_fields=true`.

danielbreves:
Validates the behaviour of cross matching queries for title & abstract searching.

When running the query_string query against multiple fields ES uses [best_fields](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-query-string-query#query-string-multi-field-parms) as the default type. 

When we give [fields](https://github.com/destiny-evidence/destiny-repository/blob/28d656f5d74587cd30400c4109618d8312d7ea0a/app/domain/references/services/search_service.py#L28C5-L31) to ES it [expands each query term to an OR clause](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-query-string-query#query-string-multi-field) like this:

`(title:george OR abstract:george) AND (title:harrison OR abstract:harrison)`

We document search in https://destiny-evidence.github.io/destiny-repository/procedures/search.html#query-string-required:

```
# Get references with both "climate change" and "health" anywhere in the title or abstract:
?q=climate change AND health
```

Is this sufficient documentation?

Fixes #504

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -13,7 +13,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           sparse-checkout: .
       - name: Set up Docker Buildx
@@ -26,7 +26,7 @@ jobs:
   build-ui:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           sparse-checkout: ui/
       - id: filter-ui
--- a/.github/workflows/deploy-to-container-app-jobs.yml
+++ b/.github/workflows/deploy-to-container-app-jobs.yml
@@ -1,0 +1,78 @@
+name: Deploy to Container App Jobs
+
+concurrency:
+  # Allow only one active workflow run per environment
+  group: ${{ github.workflow }}-${{ inputs.environment }}
+  cancel-in-progress: true
+
+on:
+  workflow_call:
+    inputs:
+      environment:
+        type: string
+        description: environment to deploy to
+        required: true
+      image-tag:
+        type: string
+        description: image tag to deploy
+        required: true
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  deploy-container-apps-jobs:
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.environment }}
+    steps:
+      - name: "Az CLI login"
+        uses: azure/login@v2.1.0
+        with:
+          client-id: ${{ vars.AZURE_CLIENT_ID }}
+          tenant-id: ${{ vars.AZURE_TENANT_ID }}
+          subscription-id: ${{ vars.AZURE_SUBSCRIPTION_ID }}
+
+      - name: Update scheduled container app jobs images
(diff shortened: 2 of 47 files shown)


tests/integration/test_search.py
```diff
@@ -324,3 +324,258 @@ async def test_query_string_error_handling(
...
+    assert response.status_code == status.HTTP_200_OK
+    data = response.json()
+
+    assert data["total"]["count"] == 1
+    title = data["references"][0]["enhancements"][0]["content"]["title"]
+    assert "George Harrison" in title
+
+
+@pytest.fixture
+async def cross_field_idf_references(es_client: AsyncElasticsearch):
```
danielbreves: This test overfits to demonstrate the difference between `best_fields` and `cross_fields`, but I'm not entirely convinced `cross_fields` is the best option for this use case.

danielbreves:
Worth noting this warning from ES docs:

>Warning
> The cross_fields type blends field statistics in a complex way that can be hard to interpret. The score combination can even be incorrect, in particular when some documents contain some of the search fields, but not all of them. You should consider the [combined_fields](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-combined-fields-query) query as an alternative, which is also term-centric but combines field statistics in a more robust way.

https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-multi-match-query#type-cross-fields

Adam-Hammo:
Wow, `combined_fields` even uses a title+abstract example!

mootpointer:
So confirming the observed behaviour:

 - `best_fields` will only produce one result
 - `cross_fields` will produce more results, but in this case with lower precision?
 
 I think this validates what we need to know about the search options – we can talk to folks about enabling cross-field searching (perhaps as a per-query option)

danielbreves:
@mootpointer `best_fields` returns 2 matches, ranking the irrelevant one (with all terms in abstract) higher than the relevant one with terms spread across fields. `cross_fields` ranks the relevant one higher as it looks at both fields as one, but it also matches all control references (10 references including change and health in the abstract).

jaybea (approved):
LGTM, let's just Let It Be.

PR #519, opened 2026-01-28
Rebuilding an index maintains number of shards

jack-o-the-green:
Previously rebuilding an index would always reset the shard count to one


Adam-Hammo (approved):
LGTM

PR #520, opened 2026-01-29
Persist index settings over rebuilds and migrations and enable slowlogs

### LINKED ISSUE #502: Configure slowlogs on our elasticsearch clusters
https://www.elastic.co/docs/reference/elasticsearch/index-settings/slow-log

We want log emission and some guidance on visualization/searching of the logs to help us identify slow running queries

jack-o-the-green:
Fixes #502 

I've implemented this in a way that combines incoming settings changesets with the existing settings on an index for both migrations and rebuilds (though rebuilds apply an empty changeset and just maintain existing index settings). 

As we expand out the capabilities of the migration, it felt like it was going to become easy to miss maintaining settings if we did them on a per-setting value basis (e.g. rebuilds missing index shards #519), so I've gone for a more abstracted way of doing things. This should allow us to expand the post body of the system/index/{index_alias}/repair endpoint to allow specification of index setting changesets for a rebuild, or let us add additional settings via the es migrator commandline tool without changing the index_manager itself

I've added slowlog thresholds but I I _haven't_ implemented them on a per-index basis, just set default thresholds for all indices. We can come back and change that later if we need to, but felt a bit overkill for the time being. They can be configured through terraform which will add them to the container app job

### DIFF
--- a/app/persistence/es/index_manager.py
+++ b/app/persistence/es/index_manager.py
@@ -99,24 +99,6 @@ class IndexManager:
         except elasticsearch.NotFoundError:
             return None
 
-    async def get_current_number_of_shards(self) -> int | None:
-        """
-        Get the number of shards for the current index.
-
-        Returns:
-            Number of shards or None if no current index
-
-        """
-        current_index_name = await self.get_current_index_name()
-        if current_index_name is None:
-            return None
-
-        index_settings = await self.client.indices.get_settings(
-            index=current_index_name
-        )
-        settings = index_settings[current_index_name]["settings"]["index"]
-        return int(settings["number_of_shards"])
-
     @tracer.start_as_current_span("Rebuild index")
     async def rebuild_index(self) -> None:
         """
--- a/tests/integration/test_es_index_manager.py
+++ b/tests/integration/test_es_index_manager.py
@@ -1,6 +1,7 @@
 """Tests for the elasticsearch index manager."""
 
 from collections.abc import AsyncGenerator
+from typing import Any
 
 import pytest
 
@@ -67,7 +68,8 @@ async def test_initialise_es_index_happy_path(index_manager: IndexManager):
     assert current_version == 1
 
     # Assert that the index has the correct number of shards
-    assert await index_manager.get_current_number_of_shards() == 2
+    index_settings = await get_current_index_settings(index_manager)
+    assert int(index_settings["number_of_shards"]) == 2
 
 
 async def test_initialise_es_index_is_idempotent(index_manager: IndexManager):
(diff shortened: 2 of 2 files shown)


jack-o-the-green:
Tested this out on development 

reference_v4
```
{
  "reference_v4": {
    "settings": {
      "index": {
        "routing": {
              ....
          }
        },
        "number_of_shards": "3",
        "provided_name": "reference_v4",
        "creation_date": "1769483474583",
        "number_of_replicas": "1",
        "uuid": "Rd_c5B8yRtebE0xEWHB7nw",
        "version": {
          "created": "9009000"
        }
      }
    }
  }
}
```

migration command 

```
python -m app.utils.es.es_migration --migrate --alias reference
```

reference_v5 settings 
```
{
  "reference_v5": {
    "settings": {
      "index": {
        "routing": {
            ...
        },
        "search": {
          "slowlog": {
            "threshold": {
              "query": {
                "warn": "1000ms",
                "trace": "125ms",
                "debug": "250ms",
                "info": "500ms"
              }
            }
          }
        },
        "number_of_shards": "3",
        "provided_name": "reference_v5",
        "creation_date": "1769727687074",
        "number_of_replicas": "1",
        "uuid": "Q40J5M2lQemGqEiHwJv2CA",
        "version": {
          "created": "9009000"
        }
      }
    }
  }
}
```

You can see the number of shards is persisted and the slowlog config enabled for search queries

app/persistence/es/index_manager.py
```diff
@@ -468,3 +479,54 @@ async def rollback(
...
+        Args:
+            current_settings: Current index settings
+            settings_changeset: Changeset to apply to current settings
+
+        Returns:
+            Index settings with changes applied.
+
+        """
+        current_settings = await self._get_reusable_index_settings(index_name)
+        return always_merger.merge(current_settings, settings_changeset)
```
Adam-Hammo: What would we do here if we wanted to remove a setting? I'm not sure we can click-ops it as it requires a new index to be created to apply the settings, I think. Would setting it to `None` suffice?

I think there's also strange behaviour for lists. For eg, (picking a random list type from the index settings spec) if we ever specified `settings["index"]["sort.fields"] = ["title"]`, and we wanted to change that to `["publication_year"]`, I think the `always_merger` would produce `["title", "publication_year"]`.

app/persistence/es/index_manager.py
```diff
@@ -468,3 +479,54 @@ async def rollback(
 
         logger.info("Rolled back from %s to %s", current_index, target_index)
         return target_index
+
+    async def _get_settings_with_changeset(
```
Adam-Hammo: Very nitpicky - could we maybe call this `_get_updated_settings`? This name implies to me it would return a changeset alongside settings

app/persistence/es/index_manager.py
jack-o-the-green: > I think there's also strange behaviour for lists. For eg, (picking a random list type from the index settings spec) if we ever specified settings["index"]["sort.fields"] = ["title"], and we wanted to change that to ["publication_year"], I think the always_merger would produce ["title", "publication_year"].

Nice catch on this - I've removed the deepmerge dependency and just gone back to my initial implementation of the settings merge which handles this case. Added some extra tests for it.

Adam-Hammo (approved):
LGTM!

PR #524, opened 2026-01-29
Distribute elasticsearch repair

Adam-Hammo:
Fulfillment/mitigation (we'll take a vote on which) of parallelisation section in #163.

This PR breaks up index repairs into many small tasks. This improves performance by adding parallelism and resilience by breaking up the units of work.

On index repair request:
- Queues `Repair Index` task
- The `Repair Index` task splits up the `reference` table into ordered partitions by `id` using [ntile](https://www.postgresql.org/docs/current/functions-window.html)
- Each partition is sent off as a separate `Repair Index Chunk` task with a `min_id` and `max_id`
  - (These are 400B each, so even millions of these are okay at Azure Service Bus)
- Each `Repair Index Chunk` task then indexes every `Reference` with `id` in the given range

I've run this in development - took [8 minutes for ~800k references](https://ui.honeycomb.io/destiny-evidence/environments/development/datasets/destiny-repository-task-development/result/g71ik4zg1F?tab=overview). This is a 3x improvement over our last repair in development (though these aren't exactly like-for-like comparisons).

Our Elasticsearch instance didn't like it very much - which is probably good news, because we should be able to squeeze a lot more juice out by increasing compute.
image

Sorry about the messy commit history - temporarily pulled #520 in so I didn't break the dev index. Diff is clean.

### Operational notes

With default settings (`chunk_size=1000`, `max_async_tasks=10`), each worker requires
**~1.1 GB of available RAM**.

This should be considered when scaling worker count or increasing concurrency.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,12 +22,12 @@ from app.persistence.sql.session import (
     db_manager,
 )
 from tests.db_utils import (
-    SimpleSQLModel,
     alembic_config_from_url,
     clean_tables,
     tmp_database,
 )
 from tests.es_utils import create_test_indices, delete_test_indices
+from tests.persistence_models import SimpleSQLModel
 
 settings = get_settings()
 MIGRATION_TASK: asyncio.Task | None = None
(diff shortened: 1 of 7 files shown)


jack-o-the-green (approved):
LGTM - thanks for doing those updates to the test utils!

PR #529, opened 2026-02-02
513: Make persistence config environment-specific

### LINKED ISSUE #513: Make persistence sizes environment-specific
We want to be able to configure instance sizes and other parameters per-environment, as our storage and usage across development, staging and production will all be different.

At present we have `is_production/is_development` local vars. These are not applied universally across the configuration. This approach will also inevitably lead to nested ternaries - are there other approaches we should consider?

**Acceptance Criteria:**
Our environments have specific and reasonable sizes/instance types for persistence.

danielbreves:
We want to be able to configure instance sizes and other parameters per-environment, as our storage and usage across development, staging and production will all be different.

This PR should enable different config per environment, but it should cause no config changes at this stage.

Fixes #513

### DIFF
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -381,6 +381,17 @@ class Settings(BaseSettings):
         description=("Override the default Elasticsearch indexing chunk size."),
     )
 
+    es_reference_repair_chunk_size: int = Field(
+        default=1000,
+        description=(
+            "Number of reference records to process in a single distributed task "
+            "when repairing or rebuilding the reference index in Elasticsearch. "
+            "Be wary that if increased too far, then the "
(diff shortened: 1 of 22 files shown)


infra/app/locals.tf
```diff
@@ -7,8 +7,41 @@ locals {
     "reference*",
     "robot-automation-percolation*"
   ]
-  is_production  = var.environment == "production"
-  is_development = var.environment != "production" && var.environment != "staging"
+
+  environment_configs = {
```
jack-o-the-green: I'm thinking we should have it default to the dev config in the event that there isn't a matching environment name - I've used an `infra-sandbox` environment before to test big changes while dev was busy and I believe this would prevent that environment from deploying.

PR #531, opened 2026-02-02
Flatten ES mapping for reference index

### LINKED ISSUE #516: Remove indexing of nested data
Once existing consumers don't depend on it (#515) we can look to remove all nested documents (identifiers and enhancements) from `ReferenceDocument`. This currently is captured by `ReferenceDomainMixin`.

A staged approach may be suitable here - perhaps starting with abstracts as a low-hanging but very juicy fruit.

We also need to consider whether we should project identifiers for search (currently, we use identifiers only as a lookup, which we can do at the database level). Gut feel is we can probably leave them for now.

Once cutover, the domain model "tied" to a `ReferenceDocument` is now a `ReferenceSearchFields`, not a `Reference`.

**Percolation gotcha**
Brief refresher on percolation:
- We store queries on the `robot_automation_percolation` index
- We get a reference from the Postgres database, in full rich nested format
- We "search" this reference against the `robot_automation_percolation` index. (we also provide a changeset, in an identical format)
- The `robot_automation_percolation` index returns any queries that are satisfied by the reference & changeset

The catch here is that `robot_automation_percolation` must know what structure to expect the `Reference` to be in. It validates both queries and search data against this mapping.

At the moment, I think it's fair to assume that we might need more than just our basic Elasticsearch projections to perform these automation decisions. Ergo, we should continue to include some information in the `robot_automation_percolation` mapping about the nested format (currently called `ReferenceDomainMixin`). We don't need to think too hard about how the actual `ReferenceDomainMixin` looks at this point - it should be very easy to update when required now that it no longer stores data.

TL;DR - we should keep `ReferenceDomainMixIn` in some capacity. It's no longer important to the `reference` index, but is to `robot_automation_percolation`. I'd support any naming/structural changes that might make this delineation clearer. 

**Acceptance criteria**
No nested data being indexed in Elasticsearch.

jaybea:
## Summary

Replace nested identifiers/enhancements in the `reference` ES index with a lean structure:

- Identifiers stored as ES Flattened type (`{"doi": "...", "pmid": "..."}`)
- Search fields (title, abstract, authors) remain at root level
- `to_domain()` returns minimal Reference (id + visibility only)
- Percolation index retains nested structure via `ReferenceDomainMixin`

Includes guard test to ensure new `ExternalIdentifierType` values get explicitly mapped.

## Notes

- Should merge after #530 (two-stage search) which removes consumer reliance on `to_domain()` returning full data
- This is a mapping-breaking change - will require reindex from PostgreSQL on deploy

Closes #516

### DIFF
--- a/tests/integration/test_search.py
+++ b/tests/integration/test_search.py
@@ -6,6 +6,7 @@ import pytest
 from elasticsearch import AsyncElasticsearch
 from fastapi import FastAPI, status
 from httpx import ASGITransport, AsyncClient
+from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.api.exception_handlers import (
     es_exception_handler,
@@ -13,7 +14,10 @@ from app.api.exception_handlers import (
 )
 from app.core.exceptions import ESQueryError, ParseError
 from app.domain.references import routes as references
-from app.domain.references.repository import ReferenceESRepository
+from app.domain.references.repository import (
+    ReferenceESRepository,
+    ReferenceSQLRepository,
+)
 from tests.factories import (
     AbstractContentEnhancementFactory,
     AnnotationEnhancementFactory,
@@ -24,6 +28,15 @@ from tests.factories import (
 )
 
 pytestmark = pytest.mark.usefixtures("session")
+
+
+def get_bibliographic_content(ref: dict) -> dict:
+    """Find bibliographic enhancement content from a reference response."""
+    for enh in ref["enhancements"]:
+        if enh["content"]["enhancement_type"] == "bibliographic":
+            return enh["content"]
+    msg = "No bibliographic enhancement found"
+    raise ValueError(msg)
 
 
 @pytest.fixture
@@ -53,7 +66,7 @@ async def client(
 
 
 @pytest.fixture
-async def search_references(es_client: AsyncElasticsearch):
+async def search_references(es_client: AsyncElasticsearch, session: AsyncSession):
     """Create and index test references with various metadata and annotations."""
     references = [
         ReferenceFactory.build(
@@ -183,8 +196,11 @@ async def search_references(es_client: AsyncElasticsearch):
     )
 
     es_repository = ReferenceESRepository(es_client)
+    sql_repository = ReferenceSQLRepository(session)
     for reference in references:
         await es_repository.add(reference)
+        await sql_repository.merge(reference)
+    await session.commit()
     await es_client.indices.refresh(index="reference")
     return references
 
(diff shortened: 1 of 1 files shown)


app/domain/references/models/es.py
```diff
@@ -340,30 +342,66 @@ def from_domain(cls, reference: Reference) -> Self:
...
-        """Create a domain model from this persistence model."""
-        return ReferenceDomainMixin.to_domain(self)
+        """
+        Create a minimal domain model from this persistence model.
+
+        Since ES is now search-only, full hydration should be done from PostgreSQL.
+        This returns a minimal Reference with only the ID and visibility.
+        """
+        # meta.id can be a UUID object (when created directly) or string (from ES)
+        ref_id = self.meta.id if isinstance(self.meta.id, UUID) else UUID(self.meta.id)
```
Adam-Hammo: I think you shouldn't need to do this - passing in `str` or `UUID` to the pydantic `Reference` object should then cast it to `UUID` under the hood.

app/domain/references/models/projections.py
```diff
@@ -13,11 +14,55 @@
...
+    ExternalIdentifierType.PM_ID: "pmid",
+    ExternalIdentifierType.PRO_QUEST: "proquest",
+    ExternalIdentifierType.ERIC: "eric",
+    # OTHER uses other_identifier_name as key
+}
+
+
+def flatten_identifiers(
+    identifiers: list[LinkedExternalIdentifier] | None,
+) -> dict[str, str] | None:
```
Adam-Hammo: We might need a more sophisticated approach to this. Reference&identifier_type is not a unique key, some references may have more than one valid identifier of the same type For example, perhaps we find two references with two different PM IDs are duplicates. 

We'd either want to project _all_ IDs of the canonical and the duplicate out to Elasticsearch (suspect this might get tricky to project in a flattened format?) or add some level of priority handling like we do for enhancements.

app/domain/references/models/es.py
```diff
@@ -340,30 +342,66 @@ def from_domain(cls, reference: Reference) -> Self:
...
     class Index:
         """Index metadata for the persistence model."""
 
         name = "reference"
 
+    visibility: Visibility = mapped_field(Keyword(required=True))
+    duplicate_determination: DuplicateDetermination | None = mapped_field(
+        Keyword(required=False),
+    )
+    identifiers: dict[str, str] | None = mapped_field(Flattened(required=False))
```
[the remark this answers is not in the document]
jaybea: I agree! I initially was concerned about deduplication candidate queries, but identifiers don't add relevance.

app/domain/references/models/es.py
jaybea: Good catch, Pydantic v2 does this for us.

app/domain/references/models/projections.py
jaybea: Complexity goes away! with removing identifiers from ES.

PR #533, opened 2026-02-03
406: Set up alerting on database metrics

### LINKED ISSUE #406: Set up alerting on database metrics
**Background**
We want to get database alerts based on our remaining storage capacity, CPU, and memory usage so that we can take action to upgrade our database if needed and/or know if something has put unexpected load on it. 

**Technical scope**
* Investigate what Azure has to offer wrt to flexibility on alerting and implement something that feels like a good start. This might be 
  * Alert if we a rapidly increasing the size of the database 
  * Alert if we cross a certain threshold of percentage utilization. 

Will be somewhat dependent on what is available via azure. 
May be dependent on #405

danielbreves:
We want to get database alerts based on our remaining storage capacity, CPU, and memory usage so that we can take action to upgrade our database if needed and/or know if something has put unexpected load on it.

Fixes #406


infra/app/alerts.tf
```diff
@@ -0,0 +1,383 @@
...
+    operator         = "GreaterThan"
+    threshold        = local.env.db_memory_critical_pct
+  }
+
+  action {
+    action_group_id = azurerm_monitor_action_group.alerts[0].id
+  }
+}
+
+resource "azurerm_monitor_metric_alert" "app_restart_count" {
```
danielbreves: These will replace the ones I created via clickops.

PR #535, opened 2026-02-03
Handle race condition in deduplication task gracefully

### LINKED ISSUE #534: Handle race condition in deduplication task gracefully
## Problem

When concurrent workers process related references (e.g., two references sharing the same OpenAlex ID), they may both attempt to update the same `ReferenceDuplicateDecision`, causing `SQLIntegrityError`.

**Impact:** Flaky e2e tests in CI due to non-deterministic task ordering.

**Current behavior:**
- Task fails with `SQLIntegrityError`
- TaskIQ retries the task
- On retry, the idempotency check catches it (decision is no longer PENDING)
- Self-healing but noisy (error logs, retry overhead, flaky e2e tests)

**Stack trace (from CI):**
```
SQLIntegrityError: Unable to perform operation on ReferenceDuplicateDecision.
  File "app/persistence/sql/repository.py", line 440, in merge
```

## Root Cause

Two workers pick up decisions for references that resolve to the same canonical. Worker A commits first; Worker B's merge fails with an integrity constraint violation.

## Solution

Catch `SQLIntegrityError`, rollback the invalidated session, re-fetch the decision, and skip if already processed:

```python
except SQLIntegrityError as e:
    if e.lookup_model != "ReferenceDuplicateDecision":
        raise  # Not the expected collision
    await sql_uow.rollback()  # Clear invalid session state
    updated_decision = await reference_service.get_reference_duplicate_decision(id)
    if updated_decision.duplicate_determination != DuplicateDetermination.PENDING:
        logger.info("Decision was processed by another worker, skipping.", ...)
        return
    raise  # Still pending = different integrity issue
```

**Key details:**
- `lookup_model` check prevents swallowing unrelated integrity errors
- Explicit `rollback()` required. After `IntegrityError`, SQLAlchemy session is in an invalid state and won't accept queries until rolled back
- Logs `collision` for observability without adding log volume

## Acceptance Criteria

- [ ] Race condition handled gracefully (no error logs for expected case)
- [ ] E2E tests pass consistently
- [ ] No regression in deduplication correctness

jaybea:
## Summary
- Catch `SQLIntegrityError` in deduplication task and skip if another worker already processed the decision
- Add explicit `rollback()` to clear invalid session state before re-fetching
- Check `lookup_model` to avoid swallowing unrelated integrity errors
- Log `collision` for observability

Fixes #534

## Test plan
- [ ] `uv run pre-commit run --all-files` passes
- [ ] `uv run pytest tests/unit/domain/references/ -k dedup` passes
- [ ] E2E tests pass consistently

### DIFF
--- a/tests/unit/domain/references/test_tasks.py
+++ b/tests/unit/domain/references/test_tasks.py
@@ -5,10 +5,13 @@ from unittest.mock import AsyncMock
 
 import pytest
 
+from app.core.exceptions import SQLIntegrityError
 from app.domain.references.models.models import (
+    DuplicateDetermination,
     EnhancementRequest,
     PendingEnhancementStatus,
     Reference,
+    ReferenceDuplicateDecision,
     ReferenceWithChangeset,
     RobotAutomationPercolationResult,
     RobotEnhancementBatch,
@@ -19,6 +22,7 @@ from app.domain.references.services.anti_corruption_service import (
 )
 from app.domain.references.services.enhancement_service import ProcessedResults
 from app.domain.references.tasks import (
+    process_reference_duplicate_decision,
     validate_and_import_robot_enhancement_batch_result,
 )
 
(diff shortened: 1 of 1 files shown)


Adam-Hammo (approved):
Looks good, nice one!

PR #536, opened 2026-02-03
Chunk up distribution of import batch

Adam-Hammo:
This aims to reduce the duration of the `Distribute Import Batch` task by chunking together network calls. Currently the task runs for quite a while in order to queue the hundreds of thousands of tasks. [Example here](https://ui.honeycomb.io/destiny-evidence/environments/staging/datasets/destiny-repository-task-staging/result/k15mAs8XCnD)

While this won't noticeably improve the high-level performance of the import process, it will improve our resilience by having shorter-running tasks.

### DIFF
--- a/app/domain/imports/service.py
+++ b/app/domain/imports/service.py
@@ -12,7 +12,7 @@ from sqlalchemy.exc import DBAPIError
 
 from app.core.config import get_settings
 from app.core.exceptions import SQLIntegrityError
-from app.core.telemetry.attributes import Attributes, trace_attribute
+from app.core.telemetry.attributes import Attributes
 from app.core.telemetry.logger import get_logger
 from app.core.telemetry.otel import get_current_trace_link, new_linked_trace
 from app.core.telemetry.taskiq import queue_task_with_trace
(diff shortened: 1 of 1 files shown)


jaybea (changes requested):
Good optimisation, bulk insert and parallel queuing should help a lot.

Two small questions / suggestions:

1) It seems like he “Queue import reference task” spans would show ~0 ms because they’re really just measuring coroutine creation, not actual queueing or execution.

) If `asyncio.gather` throws (e.g. network error), we could end up "orphaning" up to 1000 `ImportResult`s.

Adam-Hammo:
These are good points - I might have been rushed in my approach.

The aim with the tracing was to separate the "queue task" log events from the "Distribute import batch chunk" trace (as this would cause that trace to have lots of events). But, since the span closes before the tasks are gathered as you note, Refinery won't necessarily have that log event available to help make a decision on whether to sample the "Queue import reference task" trace.

If it throws a network error, `asyncio.gather` will reraise and the task will fail out - regardless of if there are 1 or 1,000 ImportResults. It'll get picked back up, from the start of the file, and so we will hopefully eventually get the import done. The final state will look funky at this point, with some ImportResults in `created`, a total count > than the number of lines, but the file itself is fully ingested - not a new problem, just a QOL thing we'll address at some point. 
This is part of the reason we try to have unit-sized tasks, to minimise these side-effects and let the service bus provide resiliency.

I wonder if a better approach might be to limit (perhaps just with a recommendation) the size of import_batch files. We have a nice `record->batch` hierarchy of imports that is kind of under-utilised.

PR #537, opened 2026-02-04
adding ReferenceFileInput compatibility; adding first_author property

nikloynes:
`LabsReference` offers nice solutions for workflows in `deet`, specifically retrieving core attributes from a citation. this model doesn't support `ReferenceFileInput` which is the type we have to use given our dependency on `destiny_sdk.parsers.eppi_parser.py`. further, it didn't have a property for `first_author`. 

Summary:
- added `ReferenceFileInput` as potential type for `reference` field
- added a `@property` for `first_author`, returning string
- sdk tests are passing.

### DIFF
--- a/libs/sdk/pyproject.toml
+++ b/libs/sdk/pyproject.toml
@@ -35,7 +35,7 @@ license = "Apache-2.0"
 name = "destiny_sdk"
 readme = "README.md"
 requires-python = ">=3.12, <4"
-version = "0.9.0"
+version = "0.9.1"
 
 [project.optional-dependencies]
 labs = []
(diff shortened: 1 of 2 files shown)


libs/sdk/src/destiny_sdk/labs/references.py
```diff
@@ -80,6 +81,17 @@ def title(self) -> str | None:
                 return meta.title
         return None
 
+    @property
+    def first_author(self) -> str | None:
+        """The first author of the reference as str."""
+        for meta in self.it_bibliographics():
+            if meta.authorship is not None and len(meta.authorship) > 0:
+                for author in meta.authorship:
+                    if author.position == AuthorPosition.FIRST:
```
Adam-Hammo: @nikloynes we have position on some authorships so I've added this in, just in case the list's order is incorrect. Let me know if this doesn't suit your usecase.

PR #538, opened 2026-02-05
Handle empty search results in two-stage search

jaybea:
## Summary

Fixes 500 error when search returns no results.

The two-stage search calls `get_deduplicated_references` with an empty list when 
ES returns no hits, which triggers a `ValueError` from the validation that requires 
exactly one of `reference_ids` or `references` to be provided (`bool([]) == bool(None)`).

## Changes

- Skip `get_deduplicated_references` call when `search_result.hits` is empty
- Add regression test for empty search results

## Test plan

- [x] `test_empty_search_results` verifies 200 response with empty results
- [x] All existing search tests pass

### DIFF
--- a/app/domain/references/routes.py
+++ b/app/domain/references/routes.py
@@ -343,8 +343,12 @@ async def search_references(
         publication_year_range=publication_year_range,
         sort=sort,
     )
-    references = await reference_service.get_deduplicated_references(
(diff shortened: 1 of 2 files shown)


Adam-Hammo (approved):
Nice catch!

Issue #478, opened 2026-01-19
Estimate cost


Adam-Hammo:
https://docs.google.com/spreadsheets/d/1OZ45QuX2Raun5mKeKAoDv5W-oS_6fNYU9qYY2DZMNIM/edit?usp=sharing

PR #548, opened 2026-02-10
Scale production postgres database

### LINKED ISSUE #480: Scale Postgres instance
We need to increase our Postgres database size to accommodate the new data, and likely need to upgrade our instance for performance.

We also need to make the size configurable per-environment.

See #478 for an estimate on production sizing.

**Acceptance criteria**
All database instances upgraded to their prescribed sizes.

Adam-Hammo:
Closes #480.

4TB derived here: #478. We're expected to grow to just over 2TB, and 4TB is the next tier.

TF diff:
image

### DIFF
--- a/infra/app/locals.tf
+++ b/infra/app/locals.tf
@@ -56,7 +56,7 @@ locals {
       db_storage_tier = "P50"
       db_backup_days  = 35
       db_ha_enabled   = true
-      db_sku_name     = "GP_Standard_D8ds_v5"
+      db_sku_name     = "GP_Standard_D8ds_v5" # More memory -> better chance of indexes being cached in memory, improving performance
 
       alerts_enabled          = true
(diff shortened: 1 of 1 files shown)


infra/app/locals.tf
```diff
@@ -50,10 +52,11 @@ locals {
     }
 
     production = {
-      db_storage_mb   = 131072 # 128GB
-      db_storage_tier = "P10"
+      db_storage_mb   = 4193280 # 4TB with host caching
+      db_storage_tier = "P50"
       db_backup_days  = 35
       db_ha_enabled   = true
+      db_sku_name     = "GP_Standard_D8ds_v5"
```
jack-o-the-green: Is there value in detailing what the `sku`s mean or linking to something that does?

PR #551, opened 2026-02-12
Add peer connection closed resilience to import batch distribution

Adam-Hammo:
Sometimes when streaming an import batch file we get a `PeerConnectionClosed` error from Azure: https://ui.honeycomb.io/destiny-evidence/environments/staging/datasets/destiny-repository-task-staging/result/s1CyBUYt6Za/trace/D8CdfaDqFMz?fields[]=s_name&fields[]=s_serviceName&span=be0a570271237c4c

Our current handling of task failures is to explicitly retry, and otherwise fail out with alerting to make a human decision (as we're still learning the system boundaries and best course of recovery). Here I think we can safely retry.

This tracks how far we've gone through the file to allow us to quickly resume distribution.


Adam-Hammo:
Great suggestions, thank you - have applied

PR #552, opened 2026-02-12
Use a single container app environment for our container apps

### LINKED ISSUE #359: Consolidate container apps and container app jobs into a single container app environment per repository deployment
**Background** 
As our container module app creates a container app environment https://github.com/futureevidence/terraform-azure-container-app/blob/main/main.tf#L10, we're currently creating a seperate container app environment for the destiny-repository, the tasks app, and the UI. We don't want to do this as they should all be in the same container apps environment. 

We are still a long way off hitting the container app environment quota (18/50) when this ticket was raised, so there is not immediate urgency to complete this.

**Technical Scope**
* Remove container app environment creation from the container app module, or make it optional on request. 
    * Will need to allow passing in the container app env id to the container app module.
* Publish a new version of the module. 
* Update everything to a) create it's own container app environment and b) use the new module version
    * For destiny-repository, this is the UI, tasks app, application, and any container app jobs that are currently using the container app environment created for the application. 

**Note**
There's likely to be some annoying terraform state management as part of this - we may end up needing to recreate all the container app etc. This should be a safe action, as everything is already managed through terraform, but worth the call out.

jack-o-the-green:
Fixes #359 

Latest release of the container app module allows specification of a container app environment to use when creating the container app.

You can see it successfully rolled out to dev here https://app.terraform.io/app/destiny-evidence/workspaces/destiny-repository-development/runs/run-QVTXZvVuHGoZUf7t. The main app, the ui, and the tasks are all running in the same container app environment along with the container app jobs. 

#### Rollout
This rollout causes some downtime as the re-creation of the container apps requires us to deploy the correctly tagged image out once they're up and running. We'll want to make sure we do this at a time of relatively low traffic. 

This will also change the FQDN of the UI container app, meaning we will need to share these updated urls with the rest of the destiny folks. 

I'm thinking we'll do staging tomorrow, let it marinate over the weekend and then deploy to production early Monday morning.

### DIFF
--- a/infra/app/main.tf
+++ b/infra/app/main.tf
@@ -177,7 +177,7 @@ data "azurerm_container_app" "ui" {
 
 module "container_app" {
   source                          = "app.terraform.io/destiny-evidence/container-app/azure"
-  version                         = "1.9.0-beta"
+  version                         = "1.9.0"
   app_name                        = var.app_name
   cpu                             = var.container_app_cpu
   environment                     = var.environment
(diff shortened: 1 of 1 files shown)


Adam-Hammo (approved):
Code & rollout plan LGTM

PR #557, opened 2026-02-14
Escape double quotes in annotation filter query strings

jaybea:
## Summary

- Annotation labels interpolated into Lucene `query_string` quoted terms were not escaped, allowing `"` in the label to break out of the quoted context and inject arbitrary Lucene syntax
- Adds `escape_lucene_quoted_term()` to `app/utils/regex.py` (escapes `\` and `"` only)
- Applies escaping in the `annotations:"scheme/label"` code path in `SearchService._build_annotation_query_string_filter()`

## Scope

This fixes the quoted-term injection vector only. The `q` parameter intentionally exposes full Lucene query syntax to authenticated users, so the practical risk was low (same user, same data, same index). The unquoted wildcard path (`annotations:scheme*`) is not addressed here — it only escapes `:`, but since `q` already grants equivalent access, the residual risk is nil.

## Test plan

- [x] Unit test: `test_annotation_filter_escapes_quotes_in_label` — verifies `"` in label is escaped to `\"`
- [x] `pytest tests/unit` — all passing
- [x] `pre-commit run --all-files` — clean

### DIFF
--- a/tests/unit/domain/references/services/test_query_string_filters.py
+++ b/tests/unit/domain/references/services/test_query_string_filters.py
@@ -17,7 +17,7 @@ def test_annotation_filter_escapes_quotes_in_label():
     #   — the " terminates the quoted term early, leaving `quote"` as
     #     bare Lucene syntax that can alter query logic.
     service = _StubSearchService()
-    result = service._build_annotation_query_string_filter(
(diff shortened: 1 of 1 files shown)


jaybea:
No specific scenario, just did a vulnerability assessment, and this one showed up. It looked reasonable to do and seemed a nice correctness/parsing improvement.

PR #558, opened 2026-02-15
Add domain allowlist for import batch storage URLs

jaybea:
## Summary

- Validate `storage_url` at batch enqueue time against a configurable domain
  suffix allowlist (`settings.allowed_import_domains`, default:
  `["blob.core.windows.net"]`)
- Disable redirect-following in the batch distributor's HTTP client and
  explicitly reject non-2xx responses, preventing open-redirect bypasses
- Validation skipped in local dev (`running_locally=True`); empty allowlist
  disables the check entirely

## Rework note

The initial commits on this branch used DNS-based IP validation (resolve
hostname, reject non-global IPs, TTL cache, threading locks). Reviewer
feedback found this too complex for the threat model — Azure Container Apps
already limits the attack surface. The final commit replaces all of that with
a simple domain suffix allowlist.

## Changes

| File | What |
|------|------|
| `app/core/config.py` | Add `allowed_import_domains` setting |
| `app/utils/validate_url.py` | Domain suffix matching with dot boundary, case-insensitive |
| `app/domain/imports/routes.py` | Call validator in `enqueue_batch`, return 422 on failure |
| `app/domain/imports/service.py` | `follow_redirects=False`, `is_success` check (unchanged from initial PR) |
| `tests/unit/test_url_validation.py` | 3 tests (8 cases) covering domain matching, empty allowlist, local bypass |
| `tests/routers/test_imports.py` | Updated route-level mock for sync validator |
| `tests/unit/domain/imports/test_service.py` | Comment cleanup only |

## Test plan

- [x] Unit tests: 213 passed, 1 skipped
- [x] Pre-commit: 16/16 passed
- [x] Manual: set `ENV=staging`, confirmed 422 on disallowed domains, pass on
  `*.blob.core.windows.net`


jaybea:
This is the mechanism used in the 2019 Capital One breach. An attacker exploited an SSRF vulnerability in a WAF/proxy to hit the EC2 IMDS at 169.254.169.254, retrieved IAM role temporary credentials, and used those to exfiltrate ~100M customer records from S3.

In our case, `distribute_import_batch` runs asynchronously in a worker, so errors aren't returned in the HTTP response, so they don't route through the API. So the issues are greatly mitigated. As far as I can tell they seem to be internal service discovery via timing and unintended interactions with internal endpoints.

I looked at white-label and it would be easy today with just one importer, but requires maintenance whenever storage providers change. The DNS validation is a zero-maintenance generic defense. It blocks all non-globally-routable IPs regardless of provider. But I'd be happy to rework this to white-label + `follow_redirects=False`.

I looked at the robots, and thought those URLs aren't user-provided, they're signed blob storage URLs generated by our system via get_signed_url. No SSRF vector there I could find.

mootpointer:
I don't mind just having an allow-list of storage domains. I think resolving them is probably a bit more than I want to do at runtime. 

Looking at the Azure Container Apps deployment we have, SSRF is quite difficult, since there are less metadata services or similar accessible to the host. I think SSRF in this context is fairly well mitigated (although a lightweight allow-list would probably take it the rest of the way)

app/core/config.py
```diff
@@ -492,6 +492,15 @@ class Settings(BaseSettings):
         "This applies to both opentelemetry and standard logging.",
     )
 
+    allowed_import_domains: list[str] = Field(
```
Adam-Hammo: I frequently forget to do this - can we add the terraform variable and plumb it into the container app's env? That way to update the allowlist we only need to run a terraform apply, not a full code change.

tests/routers/test_imports.py
```diff
@@ -185,6 +185,29 @@ async def test_create_batch_for_import(
...
+    monkeypatch.setattr(
+        "app.domain.imports.routes.validate_storage_url",
+        MagicMock(side_effect=ValueError("storage_url hostname is not allowed.")),
+    )
+
+    batch_params = {"storage_url": "https://disallowed.example.com/data.jsonl"}
+    response = await client.post(
+        f"/v1/imports/records/{valid_import.id}/batches/", json=batch_params
+    )
+    assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
```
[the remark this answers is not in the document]
jaybea: Switched to 403. Makes more sense — the request is understood but the domain isn't permitted.

app/core/config.py
jaybea: Done.
I added `allowed_import_domains` variable in `variables.tf` (defaults to ["blob.core.windows.net"]) and plumbed it into the container app env via `jsonencode` in `main.tf`. Changeable with terraform  apply now.

Adam-Hammo (approved):
Looks good!

PR #559, opened 2026-02-15
Fix dedup shortcut side-effect race condition

jaybea:
## Problem

`test_deduplication_shortcut_from_scratch` fails intermittently in CI (observed on [main Feb 6](https://github.com/destiny-evidence/destiny-repository/actions/runs/21736093780), [main Feb 10](https://github.com/destiny-evidence/destiny-repository/actions/runs/21850970823), and [PR #557 Feb 14](https://github.com/destiny-evidence/destiny-repository/actions/runs/22015207870)).

When multiple references sharing trusted identifiers are imported concurrently, two workers both try to create side-effect `ReferenceDuplicateDecision(active_decision=True)` records for the same candidate, violating `uix_reference_one_active_decision_constraint`. The losing worker's transaction rolls back, leaving no active decision for that reference — causing `poll_duplicate_process` to time out.

## Root cause

1. Two references imported in the same batch get concurrent dedup tasks (worker runs 2 processes)
2. Both reach `shortcut_deduplication_using_identifiers`, both find the other as undeduplicated
3. Both try to create `active_decision=True` for the other simultaneously
4. One succeeds, one hits `SQLIntegrityError` on `uix_reference_one_active_decision_constraint`
5. Crashed task's transaction rolls back — no active decision for that reference
6. `poll_duplicate_process` times out → `TestPollingExhaustedError`

## Reproduction

Before applying the fix, we reproduced the race condition reliably by adding `await asyncio.sleep(5)` before the side-effect loop in `deduplication_service.py` (line 528). This widened the race window enough for both concurrent tasks to always collide. The test failed with the identical `TestPollingExhaustedError` seen in CI.

## Fix

Two layers of defense:

1. **Defensive guard check** (`deduplication_service.py`): Before creating side-effect decisions, bulk-fetch existing active decisions for all undeduplicated candidates using a single scalar SELECT (bypasses ORM identity map to see committed state from other transactions). Skip candidates that already have a non-UNSEARCHABLE active decision. UNSEARCHABLE candidates are explicitly not skipped — they are treated as undeduplicated per existing semantics.

2. **Task-level retry** (`tasks.py`): Add `remaining_retries` parameter (default 3). On `SQLIntegrityError`, verify it's specifically the active-decision unique constraint race (`lookup_model == "ReferenceDuplicateDecision"` AND `"(reference_id, active_decision)"` in the collision detail) before retrying. Other integrity errors (NOT NULL, FK, different unique constraints) re-raise immediately. On match: rollback, re-fetch the decision, and either skip (if already processed) or re-queue with decremented retries.

Bulk repository method `get_active_decision_determinations` fetches all active decisions in one query rather than one per candidate.

## Verification

After applying the fix, we verified with `await asyncio.sleep(1)` in the same location — the race window was widened, the collision occurred, the retry handled it, and the test passed. We then ran the full e2e deduplication suite 5 times clean with no failures.

## Acceptance

- [x] `uv run pytest tests/unit/ -v` — 207 passed
- [x] `uv run pytest tests/e2e/imports/test_deduplication.py -v` — 8 passed, 5 consecutive runs, zero flakes
- [x] Race reproduced on clean codebase (sleep(5) → deterministic failure)
- [x] Race survived with fix applied (sleep(1) → collision handled, test passed)
- [x] `uv run pre-commit run --all-files` — all passed

### DIFF
--- a/app/domain/imports/service.py
+++ b/app/domain/imports/service.py
@@ -3,6 +3,7 @@
 from uuid import UUID
 
 import httpx
+import tenacity
 from asyncpg.exceptions import DeadlockDetectedError  # type: ignore[import-untyped]
 from opentelemetry import trace
 from opentelemetry.instrumentation.httpx import HTTPXClientInstrumentor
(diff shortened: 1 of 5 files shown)


app/domain/references/tasks.py
```diff
@@ -260,6 +260,7 @@ async def repair_robot_automation_percolation_index() -> None:
 @broker.task
 async def process_reference_duplicate_decision(
     reference_duplicate_decision_id: UUID,
+    remaining_retries: int = 3,
```
jack-o-the-green: Do we need this number of retries for a race condition? Would not a single retry suffice?

jack-o-the-green: Nice, thanks Jaybe!

PR #567, opened 2026-02-26
563: Add Keycloak client credentials flow to SDK

### LINKED ISSUE #563: SDK: Add Keycloak client credentials flow
Parent: destiny-evidence/destiny-repository#223

Add client credentials (M2M) support to the SDK's Keycloak authentication, so services like the demonstrator-ui and openalex-incremental-updater can authenticate without user interaction.

### `KeycloakClientCredentialsFlow` in `keycloak_auth.py`

Add a new class alongside `KeycloakAuthCodeFlow`. Uses `authlib`'s `OAuth2Client` (already a dependency) with `token_endpoint_auth_method="client_secret_post"` and `grant_type="client_credentials"`. Returns a `TokenResponse` (reuse existing dataclass; `refresh_token` will be `None`).

### Extend `KeycloakOAuthMiddleware` in `client.py`

Add optional `keycloak_client_secret` parameter. When provided, instantiate `KeycloakClientCredentialsFlow` instead of `KeycloakAuthCodeFlow`. Add `_get_token_from_client_credentials` method following the same pattern as Azure's `_get_token_from_confidential_client`. No refresh token logic needed — client credentials tokens are re-acquired on expiry.

### Unit tests

Add `test_keycloak_client_credentials_auth_flow` in `libs/sdk/tests/unit/test_client.py` following the existing mock pattern (see `test_confidential_client_auth_flow`).

danielbreves:
## Summary
- Add `KeycloakClientCredentialsFlow` class in `keycloak_auth.py` for M2M authentication using authlib's `OAuth2Client` with `client_secret_post`
- Extend `KeycloakOAuthMiddleware` in `client.py` with optional `client_secret` parameter — when provided, uses client credentials flow instead of auth code flow
- Add unit tests for client credentials auth flow and token re-acquisition on 401 expiry

Closes #563

### DIFF
--- a/libs/sdk/src/destiny_sdk/client.py
+++ b/libs/sdk/src/destiny_sdk/client.py
@@ -11,7 +11,7 @@ from msal import (
     PublicClientApplication,
     UserAssignedManagedIdentity,
 )
-from pydantic import HttpUrl, TypeAdapter
+from pydantic import HttpUrl, SecretStr, TypeAdapter
 
 from destiny_sdk.auth import create_signature
 from destiny_sdk.core import UUID, sdk_version
@@ -260,7 +260,7 @@ class OAuthMiddleware(httpx.Auth):
         azure_client_id: str,
         azure_application_id: str,
         azure_login_url: HttpUrl | str | None = None,
-        azure_client_secret: str | None = None,
+        azure_client_secret: SecretStr | None = None,
         *,
         use_managed_identity: bool = False,
     ) -> None:
@@ -274,7 +274,7 @@ class OAuthMiddleware(httpx.Auth):
         :param azure_login_url: The Azure login URL.
         :type azure_login_url: str
         :param azure_client_secret: The Azure client secret.
-        :type azure_client_secret: str | None
+        :type azure_client_secret: SecretStr | None
         :param use_managed_identity: Whether to use managed identity for authentication
         :type use_managed_identity: bool
         """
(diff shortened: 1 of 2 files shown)


libs/sdk/src/destiny_sdk/client.py
```diff
@@ -469,13 +473,28 @@ class KeycloakOAuthMiddleware(httpx.Auth):
...
+
     """
 
-    def __init__(
+    def __init__(  # noqa: PLR0913
         self,
         keycloak_url: str,
         realm: str,
         client_id: str = "destiny-auth-client",
+        client_secret: str | None = None,
```
jack-o-the-green: Unsure of actual value here but pydantic provides a `SecretStr` [type](https://docs.pydantic.dev/2.1/usage/types/secrets/) that helps hide values from logging and tracebacks. Could use this for the client secret until we actually use the value.

danielbreves: Nice!

libs/sdk/src/destiny_sdk/client.py
```diff
@@ -521,13 +557,34 @@ def _get_token(self, *, force_refresh: bool = False) -> str:
...
+
+        Client credentials tokens have no refresh token, so on expiry
+        we simply re-acquire a new token.
+
+        :param force_refresh: Whether to force re-acquisition of the token.
+        :type force_refresh: bool
+        :return: The OAuth2 access token.
+        :rtype: str
+        """
+        if self._token and not force_refresh:
```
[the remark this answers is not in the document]
jack-o-the-green: Ahh fantastic, thanks for clarifying!

PR #581, opened 2026-03-02
Add ability to force destructive duplicate determinations

Adam-Hammo:
Deduplication has a couple of destructive state transitions:
- `duplicate`->`canonical`
- `duplicate`->`duplicate` (changing canonical reference)

(Basically, this is the "we made a false positive decision" flow).

These are destructive because they effectively remove data from the "old" canonical reference. There _can_ be quite broad implications for this, most notably any enhancements that were derived from a false duplicate's enhancements. These aren't addressed in this PR.

Presently, if we hit one of these cases we mark the decision as `decoupled`, preserving the reference's existing decision and deferring the decision for manual review. 

This PR allows duplicate decisions from the manual endpoint to bypass this check, meaning it can avoid creating decoupled decisions and also provide the mechanism to resolve automatically-detected decouplings.


app/domain/references/service.py
```diff
@@ -1281,26 +1282,42 @@ async def make_duplicate_decisions(
...
         Make duplicate decisions for a list of reference duplicate decisions.
 
-        If both canonical and duplicate determinations are present, canonical decisions
-        will be processed first to allow subsequent duplicate decisions to point to them
+        Decisions are processed in the order provided. If a duplicate decision
+        references a canonical that is declared in the same request, the canonical
+        must appear first.
         """
         results: list[ReferenceDuplicateDecision] = []
-        # Process canonical decisions first so they can then be duplicated
```
Adam-Hammo: This would have made it impossible to merge duplicate trees with a single call to the endpoint. I've made it an API user concern and made the documentation clear that decisions are processed in order.

Adam-Hammo:
Reraising for review - there's a couple of false positives in the EEF input data that this can help us resolve.

app/domain/references/service.py
```diff
@@ -1288,26 +1289,42 @@ async def make_duplicate_decisions(
...
+                # different canonical, or to be canonical itself. Re-apply any
+                # side effects for the old canonical reference, to allow it to
+                # retrigger and automations and recalculate enhancements without
+                # the old duplicate's data.
+                old_canonical = await self.sql_uow.references.get_by_pk(
+                    old_decision.canonical_reference_id,
+                    preload=["duplicate_decision"],
+                )
+                if old_canonical.duplicate_decision:
+                    await self.apply_reference_duplicate_decision_side_effects(
```
danielbreves: Worth adding a test for this case?

app/domain/references/services/deduplication_service.py
```diff
@@ -297,14 +306,15 @@ async def map_duplicate_decision(
             msg = "Only terminal duplicate determinations can be mapped."
             raise DeduplicationValueError(msg)
 
+        canonical_ref = None
```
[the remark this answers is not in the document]
Adam-Hammo: Oops

danielbreves (approved):

PR #592, opened 2026-03-18
Linked enhancement validation

### LINKED ISSUE #572: LinkedDataEnhancement validation and verification
### Background

See https://github.com/destiny-evidence/destiny-repository/issues/553 for problem definition of what is driving this work.

Phase 2 of #571. LinkedDataEnhancements require the most validation of all our enhancement types. This issue is for implementing that validation. 

### Technical Scope

- Integrate `pyld` for JSON-LD processing
- Integrate `pyshacl` for shape validation
- Define initial SHACL shapes for ESEA Investigation/Finding/Outcome
- Build a validation function that can be called at ingest time
- Decide: is SHACL validation blocking (reject invalid data) or advisory (accept with warnings)?

### Outstanding Questions 

Should we validate concept URIs against the actual SKOS vocabulary at ingest? This catches typos and invalid codes but requires loading the full vocabulary into memory. For a vocabulary of a few hundred concepts this is trivial, but it adds a dependency.

### Resources

- SHACL specification: https://www.w3.org/TR/shacl/
- pyld (Python JSON-LD processor): https://github.com/digitalbazaar/pyld
- pyshacl (Python SHACL validator): https://github.com/RDFLib/pySHACL

### Level 1: Structural validation (Pydantic)

The SDK validates that the enhancement has the required metadata fields (`context_uri`, `vocabulary_uri`) and that `data` is a valid JSON object. This is the same level of validation every other enhancement type gets.

### Level 2: JSON-LD validity

The `data` field must be processable as JSON-LD. This means:

- It contains a `@context` (either inline or as a URI reference)
- The `@context` URI matches the `context_uri` field
- The `@type` declarations use types defined in the vocabulary

This validation can be done with a JSON-LD processor library (e.g., `pyld` in Python). It doesn't require a triplestore — just parsing.

### Level 3: SHACL shape validation

SHACL (Shapes Constraint Language) is like JSON Schema but for RDF/linked data graphs. We define shapes that say things like:

- A Finding must have exactly one Context
- A Context must have at least one country code
- If an effect size is present, it must be a decimal number
- Education level values must come from the EducationLevel concept scheme

SHACL validation is the primary mechanism for catching data quality issues at ingest. It runs against the expanded JSON-LD (after context resolution) and produces specific, actionable error messages ("Finding _:b0 is missing required property esea:hasContext").

Python tooling: `pyshacl` is the standard library for this. It takes a data graph and a shapes graph and returns validation results.

**Important trade-off:** SHACL validation requires resolving the `@context` and expanding the JSON-LD into a graph, which means the vocabulary files need to be accessible. For testing and CI, these should be bundled locally rather than fetched from remote URIs. For production, caching the vocabulary files is essential — you don't want every ingest operation making HTTP requests to resolve contexts.

### Level 4: Vocabulary version compatibility (future)

When the vocabulary evolves (new concepts added, properties renamed), we'll need to verify that a LinkedDataEnhancement produced against vocabulary v1 is still valid under v2, or flag it for re-mapping. This is a future concern but the `vocabulary_uri` and `context_uri` fields are designed to support it.

Adam-Hammo:
Closes #572 

This is unit tested but not wired into actual flows as we can't yet resolve them. (#593)

Validates:
- `@context` exists and `==enhancement.context_uri`
- jsonld expansion
  - NB this would currently perform a failing http call in deployment as resolution isn't available yet
  - Testing uses an in-line `@context` instead to test
- No commas in URIs
  - See https://covidence.atlassian.net/wiki/spaces/Platforms/pages/1090256897/Proposal+Evidence+Synthesis+API+Surface+for+LinkedDataEnhancements+Issue+590
  - I might have jumped the gun on this one - we can have commas in URIs, just need to escape them in queries. May remove. @danielbreves thoughts?
- The content is valid according to the core ontology's shape file
  - For now this is a static and simple shape file in the repository

Does not yet validate:
- Whether URIs are present in the vocabulary - need to be able to retrieve the vocabulary

### DIFF
--- a/libs/sdk/src/destiny_sdk/enhancements.py
+++ b/libs/sdk/src/destiny_sdk/enhancements.py
@@ -478,14 +478,6 @@ class LinkedDataEnhancement(BaseModel):
 
     enhancement_type: Literal[EnhancementType.LINKED_DATA] = EnhancementType.LINKED_DATA
 
-    context_uri: HttpUrl = Field(
-        description=(
-            "The URI of the JSON-LD @context document used to interpret the data. "
-            "This pins the enhancement to a specific version of the vocabulary mapping."
-        ),
-        examples=["https://vocab.evrepo.org/context/v1.jsonld"],
-    )
-
     vocabulary_uri: HttpUrl = Field(
         description=(
             "The URI of the vocabulary against which this data was produced. "
@@ -496,14 +488,15 @@ class LinkedDataEnhancement(BaseModel):
 
     data: dict[str, JsonValue] = Field(
         description=(
-            "The JSON-LD content. When combined with the @context, this forms a "
-            "complete linked data graph."
+            "The JSON-LD content. Must contain an '@context' key with a valid "
+            "URI string. When combined with the @context, this forms a complete "
+            "linked data graph."
         ),
     )
 
     @model_validator(mode="after")
     def validate_context(self) -> Self:
-        """Validate that data contains a @context matching context_uri."""
+        """Validate that data contains a @context with a valid URI."""
         if "@context" not in self.data:
             msg = "data must contain a '@context' key."
             raise ValueError(msg)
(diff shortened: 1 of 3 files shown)


app/domain/references/services/linked_data_validation_service.py
```diff
@@ -0,0 +1,141 @@
...
+            return LinkedDataValidationResult(
+                conforms=False,
+                errors=[f"SHACL validation error: {exc}"],
+            )
+
+        if not conforms:
+            errors.append(f"SHACL validation failed:\n{results_text}")
+
+        return LinkedDataValidationResult(
+            conforms=len(errors) == 0,
```
danielbreves: Any reason not to return `conforms` here?

jack-o-the-green (commented):
Apologies for yet another review!

Issue #612, opened 2026-03-31
Reinvoke deduplication for incorrectly uncoupled references


jack-o-the-green:
## Background
We had a race condition in our task processing that lead to a small subset of reference duplicate decisions being incorrectly moved from `DUPLICATE` to `UNSEARCHABLE`. Now that #608 is deployed we expect this issue to be resolved, and what remains is to correct the duplicate decisions by reprocessing. 

## Technical Scope 
* Determine the `reference_ids` of all references that were moved from `DUPLICATE` to `UNSEARCHABLE` by the race condition (use the "find victims" query Jaybe put together here https://github.com/destiny-evidence/destiny-repository/pull/608#issuecomment-4149200789).
* Use the /invoke/ endpoint to trigger deduplication on those reference ids, this should correctly move them into DUPLICATE status. https://github.com/destiny-evidence/destiny-repository/blob/bd2016f67bb94b102bca3888815dd4b253c443e7/app/domain/references/routes.py#L766

Preference for use of the /invoke/ endpoint over direct database manipulation

jack-o-the-green:
We will use this to test the success of rolling out #610

PR #613, opened 2026-03-31
612 invoke deduplication cli script

jack-o-the-green:
In service of #612

Added a CLI script to invoke reference deduplication from a CSV file of reference IDs via the `/v1/references/duplicate-decisions/invoke/` endpoint. The reference_id csv file on #612 will be run with this as the following command 
```
uv run python -m cli.invoke_reference_deduplication --reference-id-file incorrectly_unsearchable_references.csv --env production
```

Also added a CLI script to check import batch status by querying the `/v1/imports/records/{id}/batches/{id}/summary/` endpoint and returning a validated `ImportBatchSummary`. I threw this together when investigating our batch distribution issue the other day (#599)

Have tested them both out locally

Worth noting that the `invoke` endpoint doesn't verify the existence of the reference_ids before adding pending reference decisions, so when I tried that out it created decisions that got stuck in pending. Happy to close this PR for the time being if we want to improve the endpoint.

### DIFF
--- a/cli/get_import_batch_summary.py
+++ b/cli/get_import_batch_summary.py
@@ -17,8 +17,8 @@ from .config import get_settings
 
 def check_import_batch_status(
     env: Environment, import_record_id: UUID, import_batch_id: UUID
-) -> destiny_sdk.robots.ProvisionedRobot:
-    """Register a robot to destiny repository."""
+) -> destiny_sdk.imports.ImportBatchSummary:
+    """Check the status of an import batch."""
     settings = get_settings(env)
 
     with httpx.Client() as client:


cli/get_import_batch_summary.py
```diff
@@ -0,0 +1,79 @@
...
+from app.core.config import Environment
+from cli.auth import CLIAuth
+
+from .config import get_settings
+
+
+def check_import_batch_status(
+    env: Environment, import_record_id: UUID, import_batch_id: UUID
+) -> destiny_sdk.robots.ProvisionedRobot:
+    """Register a robot to destiny repository."""
```
Adam-Hammo: Return type and docstring need updating

cli/invoke_reference_deduplication.py
```diff
@@ -0,0 +1,72 @@
...
+import httpx
+from fastapi import status
+
+from app.core.config import Environment
+from cli.auth import CLIAuth
+
+from .config import get_settings
+
+
+def invoke_reference_deduplication(env: Environment, reference_ids: list[str]) -> str:
```
Adam-Hammo:
```suggestion
def invoke_reference_deduplication(env: Environment, reference_ids: list[str]) -> int:
```

Adam-Hammo (commented):
Nice! Couple of small things

PR #627, opened 2026-04-08
Merge ontology into data graph for SHACL sh:class validation

jaybea:
## Summary

- Merge the vocabulary ontology into the data graph before SHACL validation so `sh:class` constraints can resolve concept types
- pyshacl's `sh:class` only checks `rdf:type` triples in the data graph, not `ont_graph`, even when `ont_graph` is explicitly passed
- Without this, every LDE referencing typed concepts (e.g. `EffectSizeMetricScheme/hedgesG`) fails validation because the type triple (`hedgesG rdf:type EffectSizeMetricConcept`) only exists in the vocabulary

## Context

Discovered during local end-to-end testing of the vocabulary-mapping-robot against the staging ESEA vocabulary. The robot output is valid JSON-LD — concept URIs resolve to typed nodes via the vocabulary — but SHACL validation rejected LDEs with effect estimates because it couldn't see the type assertions across graphs.

Confirmed experimentally: `inference="rdfs"` doesn't help either. The type triples must be present in the data graph for `sh:class` to find them.

## Changes

- `linked_data_validation_service.py`: one line — `data_graph += ontology` before `shacl_validate()`
- New test `test_ontology_concept_types_resolved_via_graph_merge`: validates that an LDE referencing `evrepo:hedgesG` via `effectSizeMetric` passes `sh:class evrepo:EffectSizeMetricConcept`. Verified the test fails without the fix.

## Validation performed

- `test_ontology_concept_types_resolved_via_graph_merge` passes with fix, fails without
- All existing SHACL validation tests pass (6/6)
- Local E2E against full EPPI dataset: 2,442 LDEs ingested from vocab robot. Remaining 4,131 failures are unrelated SHACL shape issues on the robot side (hasFinding minCount).

### DIFF
--- a/app/domain/references/services/linked_data_validation_service.py
+++ b/app/domain/references/services/linked_data_validation_service.py
@@ -93,6 +93,10 @@ class LinkedDataValidationService:
             )
 
         # Step 3: SHACL validation
+        # Merge ontology into data graph so sh:class constraints can resolve
+        # concept types. pyshacl's sh:class only checks rdf:type triples in the
+        # data graph, not ont_graph — even with ont_graph set.
+        data_graph += ontology
         try:
             conforms, _graph, results_text = shacl_validate(
                 data_graph,
--- a/tests/unit/domain/references/services/test_linked_data_validation_service.py
+++ b/tests/unit/domain/references/services/test_linked_data_validation_service.py
@@ -74,6 +74,46 @@ VALID_DATA = {
 }
 
 
+VALID_DATA_WITH_ONTOLOGY_CONCEPT = {
+    "@context": {"evrepo": EVREPO},
+    "@type": "evrepo:LinkedDataEnhancement",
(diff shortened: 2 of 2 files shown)


app/domain/references/services/linked_data_validation_service.py
```diff
@@ -93,6 +93,10 @@ async def validate(
             )
 
         # Step 3: SHACL validation
+        # Merge ontology into data graph so sh:class constraints can resolve
+        # concept types. pyshacl's sh:class only checks rdf:type triples in the
+        # data graph, not ont_graph — even with ont_graph set.
+        data_graph += ontology
```
Adam-Hammo: If we set `inference="rdfs"` instead of doing the manual merge, does this have the same behaviour?

Adam-Hammo (approved):

PR #632, opened 2026-04-09
Make created_at an optional field for context-specific use

Adam-Hammo:
Follow-up to #624 - the pydantic model was asserting that robots should provide `created_at`, which is untrue. It is set at the SQL layer of the repository.

### DIFF
--- a/tests/unit/domain/references/services/test_anti_corruption_service.py
+++ b/tests/unit/domain/references/services/test_anti_corruption_service.py
@@ -31,12 +31,11 @@ class TestEnhancementFromSdk:
                 process=AbstractProcessType.OTHER,
                 abstract="Test abstract content.",
             ),
+            created_at=datetime.datetime.now(datetime.UTC),
         )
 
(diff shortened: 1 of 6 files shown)


tests/unit/domain/references/services/test_anti_corruption_service.py
```diff
@@ -0,0 +1,42 @@
...
+            visibility=Visibility.PUBLIC,
+            content=destiny_sdk.enhancements.AbstractContentEnhancement(
+                process=AbstractProcessType.OTHER,
+                abstract="Test abstract content.",
+            ),
+        )
+
+    def test_created_at_is_stripped(self, service, sdk_enhancement):
+        """SDK-provided created_at should be ignored during translation."""
+        sdk_enhancement.created_at = datetime.datetime.now(datetime.UTC)
```
jack-o-the-green: This is now failing with the frozen field

jack-o-the-green (approved):
LGTM

PR #635, opened 2026-04-10
Concurrent token provider support

### LINKED ISSUE #630: Accept Keycloak JWT tokens alongside Azure AD
## Summary

Allow the destiny-repository API to accept JWT tokens from both Azure AD and Keycloak simultaneously, so that evidence-repo-ui can authenticate via Keycloak while all existing consumers continue using Azure AD.

Currently `auth_provider` is a global either/or setting — the API validates tokens from Azure *or* Keycloak, never both. We need a multi-issuer mode.

## Approach

The auth module already has both `AzureJwtAuth` and `KeycloakJwtAuth` implementing the same `AuthMethod` interface, and `HybridAuth` demonstrates dispatching between auth methods based on credential type.

A new multi-issuer JWT auth strategy could:
1. Peek at the token's `iss` claim (without full validation) to determine the issuer
2. Delegate to `AzureJwtAuth` or `KeycloakJwtAuth` accordingly
3. Be selected by `choose_auth_strategy()` when both Azure and Keycloak settings are configured

The Keycloak client would need the appropriate scopes/roles configured (at minimum `REFERENCE_READER` for evidence-repo-ui search and detail pages).

## Why

evidence-repo-ui needs auth for abstract visibility. Keycloak is the auth provider for the taxonomy-builder ecosystem. Rather than switching the entire API to Keycloak, we can accept tokens from both issuers — no disruption to existing Azure AD consumers.

## Related

- destiny-evidence/evidence-repo-ui#2 — Evidence Search Application epic

Adam-Hammo:
Closes #630 

Adds a `MultiIssuerJwtAuth` that routes authentication handling to the existing `AzureJwtAuth` or `KeycloakJwtAuth` based on the `iss` in the token.

Note this does not completely unblock the UI Auth, we still need to [create a public client](https://github.com/destiny-evidence/destiny-shared-infra/issues/32).

Terraform failure(s) are from the [sporadic AZ secrets API issues](https://covidence.slack.com/archives/C0ADTLS1675/p1776047453031099?thread_ts=1776047169.895449&cid=C0ADTLS1675), I promise.

Validated in dev with the below:

```python
# /// script
# requires-python = ">=3.14"
# dependencies = [
#     "destiny-sdk==0.11.4",
# ]
#
# [[tool.uv.index]]
# name = "testpypi"
# url = "https://test.pypi.org/simple/"
# explicit = true
#
# [tool.uv.sources]
# destiny-sdk = { index = "testpypi" }
# ///

from destiny_sdk.client import OAuthClient, KeycloakOAuthMiddleware, OAuthMiddleware

keycloak_res = OAuthClient(
    base_url="https://destiny-repository-deve-app.gentlecoast-c1c9497a.swedencentral.azurecontainerapps.io",
    auth=KeycloakOAuthMiddleware(
        keycloak_url="https://auth.evidence-repository.org",
        realm="destiny",
        client_id="destiny-auth-client-development",
    ),
).search("climate AND health")

azure_res = OAuthClient(
    base_url="https://destiny-repository-deve-app.gentlecoast-c1c9497a.swedencentral.azurecontainerapps.io",
    auth=OAuthMiddleware(
        azure_login_url="https://login.microsoftonline.com/f870e5ae-5521-4a94-b9ff-cdde7d36dd35",
        azure_client_id="12da13eb-0019-408b-b2e3-1898d2a07d5f",
        azure_application_id="0a4b8df7-5c97-42b2-be07-2bb25e06dbb2",
    ),
).search("climate AND health")

assert keycloak_res.total != 0
assert azure_res.total != 0
print(keycloak_res.total == azure_res.total)
```


libs/sdk/src/destiny_sdk/keycloak_auth.py
```diff
@@ -261,12 +267,11 @@ def log_message(self, format: str, *args: Any) -> None:  # noqa: A002, ANN401
             msg = f"Authentication failed: {received_response['error']}"
             raise RuntimeError(msg)
 
-        # Exchange authorization code for tokens using authlib
-        # authlib handles PKCE verification automatically
```
Adam-Hammo, line 265: https://docs.authlib.org/en/stable/client/oauth2.html#add-pkce-for-authorization-code

jaybea (approved):
Runtime code looks good. I ran the `auth_provider="both"` path localy against a Keycloak container and verified the full `MultiIssuerJwtAuth` matrix (real PKCE token with scope / without scope / no iss / fake iss / garbage / no auth — all six branches hit their correct path, messages match `app/api/auth.py:658-708` verbatim). The explicit `code_verifier=` in `keycloak_auth.py` looks to be required. Without it authlib's internal verifier doesn't survive the HTTPServer callbak loop and the exchange returns `invalid_grant`.

I wasn't able to exercise the Azure validator end-to-end (no live Azure token to hand), so the Azure side of the multi-issuer routing is verified only by the unit tests in `tests/core/test_multi_issuer_auth.py`.

These are nits, take action on them if you think appropriate:

I know we'd like to phase out Azure auth, but just to make sure we can still run against an Azure-only setup in the meantime, two Terraform things maybe worth a look before merge (both verified with a local `terraform plan` against an isolated copy of `variables.tf`):

1. `shared_keycloak_url` has no default, so Azure-only workspaces now fail plan. Adding `default = ""` fixes it — `main.tf:136` already guards the usage with `var.auth_provider != "azure" ? ... : ""`. Re-ran the plan with the default in place and confirmed it drops out of the required list.
2. `shared_container_registry_name` and `shared_resource_group_name` are newly required from the `destiny-evidence-shared` variable set, so each destiny-repository workspace (prod/staging/dev) needs it attached before merge. `infra/app/README.md:237-238` still lists the old variable names.

Runtime looks good.

Adam-Hammo:
Thanks @jaybea - re your terraform notes, I have already applied the `destiny-shared-infra` variable set to staging and production, so I'm not sure we need the variable-level default in that case. Let me know if I've misunderstood! I've updated the README.

Issue #640, opened 2026-04-23
Address candidate canonical nomination timeouts


Adam-Hammo:
Since the bulk OpenAlex ingestion, candidate canonical nomination times out at Elasticsearch. Potential contributing reasons:
- Increased size of Elasticsearch index leading to slower queries
- Increased horizontal scale of workers leading to more parallel queries

This causes all `process_reference_duplicate_decision` tasks to fail out. Furthermore, these tasks take a long time to do so:
- We have a 30s timeout at Elasticsearch
- We retry 5 times at Elasticsearch
- So, each task takes 3 minutes
- In turn, this is longer than our 1 minute default lock duration
- So, each task is ALSO requeued up to 10 times!

We need to decide an approach to handle this. It may be:
- Turning off candidate nomination for now (it's not even used downstream)
- Updating the query (#278)
- Something else

See also: https://covidence.slack.com/archives/C099F6PR8R2/p1776911734874049

Adam-Hammo:
Another thing to note: the requeuing of tasks due to losing the lock caused a race condition to fire:
- `reference_duplicate_decision` with id `x` is fired. We do not refresh locks on these as we expect these tasks to finish quickly try to minimise the surface on which we need that mechanism. 
  - NB we can control the default lock duration up to 5 minutes, this might be worth doing!
  - This whole thing comes about because we have had many issues with azure service bus's lock renewal function in async python, so this is at least a contributing factor. Moving away from ASB would mitigate this scenario.
- The task is picked up, and set to `pending`.
- After 1 minute, azure service bus requeues the task, assuming it is lost. This task is picked up - we now have two tasks actively processing the same duplicate decision.
  - [Example trace of one of these follow-up tasks](https://ui.honeycomb.io/destiny-evidence/environments/production/result/D26czy7V9vT/trace/k4Xer3a7bk3?fields[]=s_name&fields[]=s_serviceName&span=7a344b9be2969b5c) 
- After 2 minutes, the query on the first task finishes, and the decision is processed (`nominated`, then the terminal `unsearchable`, then the decision is set to active).
- Then, the query on the second task finishes, and the decision is also processed. It tries to set the state to `nominated`:
  - Repository call: `repository.update_by_pk(id, determination='nominated')`
  - Then repository returns `record.to_domain()`
  - But, in `to_domain()`, pydantic validates that active decisions may only have terminal states. Thus, it throws an error, as `nominated` is not terminal. 
[...]

Issue #278, opened 2025-08-26
Implement and evaluate candidate nominator


Adam-Hammo:
The prior sub-tickets will only include a proof-of-concept de-duplication candidate nominator. We should more carefully look at the implementation and evaluate it.

Adam-Hammo:
There's a bit of an implicit dependency on the incoming "deep deduplication" too - the performance of that will determine how wide of a net we want to cast.

PR #641, opened 2026-04-23
Add feature flag to toggle off candidate canonical searching

Adam-Hammo:
Partial fulfilment of #640. Candidate canonical searching is not an actively used feature (#278).

See also https://covidence.slack.com/archives/C099F6PR8R2/p1776911734874049

### DIFF
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -328,6 +328,7 @@ class FeatureFlags(BaseModel):
     """Feature flags for the application."""
 
     enable_percolation: bool = True
+    enable_canonical_candidate_search: bool = True
 
 
 class DedupCandidateScoringConfig(BaseModel):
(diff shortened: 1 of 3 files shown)


app/core/config.py
```diff
@@ -328,6 +328,7 @@ class FeatureFlags(BaseModel):
     """Feature flags for the application."""
 
     enable_percolation: bool = True
+    enable_canonical_candidate_search: bool = True
```
Adam-Hammo, line 331: This is maybe more correctly named `enable_full_deduplication`, open to thoughts. When `False` only the shortcut dedup will effectively run.

Adam-Hammo: It's a choice of whether we name according to what it guards in the code, or what codepath it effectively leads to downstream.

jaybea: I keep the naming to the guarded path for now. And I think `enable_full_deduplicaition` doesn't accurately reflect the situation yet, it may down the line, I acknowledge. I'd keep it "canonical" :)

The placeholder determinator funnels everything to UNSEARCHABLE anyway, there's isn't really a "full-dedup" to toggle.

jaybea (approved):
Couldn't really find any issues with this, nicely done!

PR #653, opened 2026-04-29
Loosen the SHACLs

Adam-Hammo:
Removes mincount constraints on hasContext, hasOutcome and hasEffectEstimate. Maintains that a Finding cannot be completely empty (as might be the case in completely uncoded works, eg as seen in ESEA).

Thread for context: https://covidence.slack.com/archives/C0810ARPCRG/p1777351294845379

This allows us to create a linked data enhancement for ESSA data in https://github.com/destiny-evidence/vocabulary-mapping-robot/issues/37.


jack-o-the-green (approved):
LGTM

Issue #645, opened 2026-04-24
Cutover consumers to Keycloak from Azure


Adam-Hammo:
Once #635 has been propagated up to production, we can begin migrating consumers to Keycloak authentication instead of Azure.

Off the top of my head:
- demonstrator UI
- ad-hoc users of the API
  - this will require docs changes
- ad-hoc users of the repository UI
- incremental updater

All new accounts should use Keycloak over Azure.

Issue #644, opened 2026-04-24
Cutover usage of app and api DNS from FQDN


Adam-Hammo:
Per #639, we have DNS for the repository API and UI:
`api.dev.evidence-repository.org`
`app.dev.evidence-repository.org`
`api.staging.evidence-repository.org`
`app.staging.evidence-repository.org`
`api.evidence-repository.org`
`app.evidence-repository.org`

We should work to move consumers to these, so we can block off the FQDNs. Off the top of my head:
- demonstrator UI
- ad-hoc usage of API (update `docs/procedures/oauth.rst` and teams communication)
- ad-hoc usage of UI (teams communication)
- all robots

Adam-Hammo:
Note a fun implication I discovered: front door has a timeout of 2 minutes but the FQDNs support 4. We should probably check apps aren't regularly making calls between 2 and 4 mins before cutting them over.

PR #662, opened 2026-05-07
Document dns and keycloak

Adam-Hammo:
Stepping stone to #644 and #645.

Three parts:
- SDK update: 
  - `OAuthMiddleware` is now a routing class that supports Azure & Keycloak. Backward-compatible for Azure users.
  - `OAuthClient` is now directly instantiable with `env` - this is a massive simplification :)
    - Eg `results=OAuthClient(env="production").search(query="cats AND dogs")`
- Documentation update
- CLI update: uses the SDK to create API clients, piggy-backing on the auth setup there.

Docs look like: [API Authentication - DESTINY Repository](https://github.com/user-attachments/files/27463721/API.Authentication.DESTINY.Repository.v1.0.documentation.pdf)

It's not a pre-requisite, but as a relatively fast follow, we should do https://github.com/destiny-evidence/destiny-shared-infra/issues/58


docs/procedures/oauth.rst
```diff
@@ -9,70 +9,123 @@ API Authentication
...
+        Client->>Auth Server: Request token (client credentials)
+        Auth Server-->>Client: Return access token
         Client->>API: API request with Authorization: Bearer <token>
         API-->>Client: Return requested data
 
 Provisioning
 ------------
 
-In order to obtain a token from Azure, you will need to be enrolled in our tenant ``JT_AD``. Please reach out if you need access.
+In order to obtain a token from the DESTINY authentication server, you will need to be enrolled in our tenant ``JT_AD``. Please reach out if you need access.
```
danielbreves: Is this still true?
> you will need to be enrolled in our tenant ``JT_AD``

Adam-Hammo: Nice catch

libs/sdk/src/destiny_sdk/client.py
```diff
@@ -646,23 +752,40 @@ class OAuthClient:
...
+        :param env: The Destiny environment. When set, fills in defaults for
+            ``base_url`` and ``auth`` if either is omitted.
+        :type env: _ENV_TYPE | None
+        :param base_url: The base URL for the Destiny Repository API. Required
+            unless ``env`` is provided, in which case it defaults to the API URL
+            for that environment.
+        :type base_url: HttpUrl | str | None
+        :param auth: The middleware for authentication. If not provided and
+            ``env`` is given, an :class:`OAuthMiddleware` configured for that
+            environment is used. If neither is provided, only unauthenticated
```
danielbreves: > If neither is provided, only unauthenticated requests can be made.

It doesn't look like we allow not providing either.

libs/sdk/src/destiny_sdk/client.py
```diff
@@ -617,6 +627,100 @@ def auth_flow(
...
+            self._inner = AzureOAuthMiddleware(
+                azure_client_id=azure_client_id,
+                azure_application_id=azure_application_id,
+                azure_login_url=azure_login_url,
+                azure_client_secret=azure_client_secret,
+                use_managed_identity=use_managed_identity,
+            )
+        else:
+            if env and not client_id:
+                client_id = f"destiny-auth-client-{env}"
```
danielbreves, line 730: Aren't these public clients, which shouldn't support the client credentials flow?

libs/ad_hoc/scripts/keycloak_user_migration/import_to_keycloak.py
```diff
@@ -0,0 +1,101 @@
...
+
+print(f"Processing {len(rows)} users{' (DRY RUN)' if args.dry_run else ''}")
+
+for row in rows:
+    email = row["email"].strip().lower()
+    group = row["group"]
+    sso = row["google_sso"].lower() == "true"
+
+    existing = api.get("/users", params={"email": email, "exact": "true"}).json()
+    if existing:
```
danielbreves:
```suggestion
    existing = api.get("/users", params={"email": email, "exact": "true"})
    existing.raise_for_status()
    if existing.json():
```

libs/sdk/src/destiny_sdk/client.py
Adam-Hammo: Correct, client credentials will need to supply their own client instead of using the `env` shortcut - that's noted in the example here: https://github.com/destiny-evidence/destiny-repository/pull/662/changes#diff-fb9579e45d2ada14c0b310f5c0006451f8dd32773cd1cf6480d23503106686bcR109

libs/sdk/src/destiny_sdk/client.py
Adam-Hammo: Good catch - we don't really have unauthenticated endpoints too so I'll update the docstring over the validation

libs/sdk/src/destiny_sdk/client.py
danielbreves: Should we raise if `client_secret` is passed in this scenario?

Adam-Hammo: Good call, doing now

libs/sdk/src/destiny_sdk/client.py
```diff
@@ -646,23 +776,41 @@ class OAuthClient:
 
     def __init__(
         self,
-        base_url: HttpUrl | str,
+        env: _ENV_TYPE | None = None,
+        base_url: HttpUrl | str | None = None,
```
danielbreves, line 781: Nit: I was wondering if it was worth putting `env` last so `OAuthClient(url, auth)` still work, but I didn't find any uses of it across Destity, so feel free to ignore.

danielbreves (approved):

cli/config.py
```diff
@@ -1,42 +0,0 @@
...
-import os
-from typing import Literal
-
-from pydantic import HttpUrl
-from pydantic_settings import BaseSettings, SettingsConfigDict
-
-from app.core.config import Environment
-
-
-class Settings(BaseSettings):
```
danielbreves, line 12: Should we also delete https://github.com/destiny-evidence/destiny-repository/blob/main/docs/cli/config.rst?

cli/register_robot.py
```diff
@@ -1,45 +1,33 @@
...
 
-from app.core.config import Environment
-from cli.auth import CLIAuth
-
-from .config import get_settings
+from cli.client import ApiArgumentParser
 
 
 def register_robot(
-    env: Environment, robot_to_register: destiny_sdk.robots.RobotIn
```
danielbreves, line 18: Should we remove " This will load the required environment file." from https://github.com/destiny-evidence/destiny-repository/blob/main/docs/cli/robot-registration.rst?

danielbreves (approved):
A couple of doc suggestions. Looks great otherwise!

danielbreves (approved):

PR #671, opened 2026-05-08
Full text enhancement structure

### LINKED ISSUE #666: Full Texts: Data Structure
Unlike other enhancement types, the SDK model will differ from the domain model. The SDK model uses URLs, while the domain model uses our existing `BlobStorageFile` (a pointer to a blob file). The anti-corruption layer should come up clutch here to bridge the gap, creating signed URLs on-demand.

This will require some small surgery in `app.domain.references.models.models` as the `EnhancementContent` discriminated union currently comes from the SDK. We will need to define our own domain `EnhancementContent` that combines the SDK models with the domain `FullTextEnhancement` model.

```python
# SDK
class FullTextEnhancement(BaseModel):
    enhancement_type: Literal[EnhancementType.FULL_TEXT]
    file_url: HttpUrl
    <metadata>

# domain
class FullTextEnhancement(DomainBaseModel):
    enhancement_type: Literal[EnhancementType.FULL_TEXT]
    blob: BlobStorageFile
    <metadata>

EnhancementContent = Annotated[
    sdk.BibliographicMetadataEnhancement
    | sdk.AbstractContentEnhancement
    | sdk.AnnotationEnhancement
    | sdk.LocationEnhancement
    | sdk.ReferenceAssociationEnhancement
    | sdk.LinkedDataEnhancement
    | sdk.RawEnhancement
    | domain.FullTextEnhancement,
    Field(discriminator="enhancement_type"),
]
```

## Metadata
We may not need, or be able to supply, all of this - open to opinions!

Field | Type | Description
-- | -- | --
`byte_size` | `int \| None` | Size of the file in bytes. Not required on import, computed during ingestion.
`sha256_checksum` | `str \| None` | SHA-256 hash of the file contents. Not required on import, computed during ingestion.
`mime_type` | `str` | IANA media type of the file (e.g. application/pdf). Defaults to application/pdf.
`version` | `DriverVersion \| None` | Manuscript version per the DRIVER scheme: published, accepted, submitted, or other.
`is_oa` | `bool \| None` | Whether this full text is available under an open access arrangement.
`license` | `str \| None` | The license governing this full text (e.g. cc-by, cc0, publisher-specific identifier).
`source` | `str \| None` | Name of the source the full text was obtained from (e.g. arxiv, pubmed).
`source_url` | `HttpUrl \| None` | Canonical URL the full text was originally retrieved from, distinct from any transient signed URL. Akin to a `LocationEnhancement`, so not sure on including this.
`retrieved_at` | `datetime \| None` | Timestamp at which the full text was retrieved.

It would be kinda nice to link a `FullTextEnhancement` to its source `LocationEnhancement` where possible, but we mint IDs at import time, so this seems hard to do from an ingestion POV unfortunately without getting fancy.

Adam-Hammo:
Closes #666

This adds SDK and domain models. These are equivalent, except the SDK model contains a URL while the domain model contains a `BlobStorageFile`.

Also makes the `BlobStorageFile` serializing/parsing natively pydantic so that it is stored in `enhancement.content` the same way we store it elsewhere. (and renames from `_sql` to `_uri`).

The SDK was not being included in mypy type-checking, it is now - as such there are a couple of minor typing changes in the SDK.

### DIFF
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -364,6 +364,19 @@ class FullTextEnhancement(DomainBaseModel):
         description="The timestamp when this full text was retrieved.",
     )
 
+    @property
+    def fingerprint(self) -> str:
+        """
+        The unique fingerprint of this full text enhancement.
+
+        Excludes retrieved_at: refetching the same file shouldn't produce
+        a different fingerprint.
+        """
+        return json.dumps(
+            self.model_dump(mode="json", exclude={"retrieved_at"}, exclude_none=True),
+            sort_keys=True,
+        )
+
 
 EnhancementContent = Annotated[
     destiny_sdk.enhancements.BibliographicMetadataEnhancement
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -349,7 +349,21 @@ class FullTextEnhancement(DomainBaseModel):
     license: str | None = Field(
         default=None,
         description="The publishing license for this full text.",
-        examples=["cc0", "cc-by"],
+        examples=[
+            "apache-2-0",
+            "cc-by",
+            "cc-by-nc",
+            "cc-by-nc-nd",
+            "cc-by-nc-sa",
+            "cc-by-nd",
+            "cc-by-sa",
+            "gpl-v2",
+            "gpl-v3",
+            "isc",
+            "mit",
+            "other-oa",
+            "public-domain",
+        ],
     )
     source: str | None = Field(
         default=None,
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -3,17 +3,20 @@
 import datetime
 import json
 from enum import StrEnum, auto
-from typing import Any, Literal, Self
+from typing import Annotated, Any, Literal, Self
 from uuid import UUID
 
 import destiny_sdk
 
-# Explicitly import these models for easy use in the rest of the codebase
-from destiny_sdk.enhancements import EnhancementContent, EnhancementType  # noqa: F401
+# Explicitly import these models for easy importing in the rest of the codebase
+# This allows us to abstract them later if we need (as has already happened with
+# EnhancementContent!)
+from destiny_sdk.enhancements import EnhancementType
 from destiny_sdk.identifiers import ExternalIdentifier, ExternalIdentifierType
 from pydantic import (
     BaseModel,
     Field,
+    HttpUrl,
     PositiveInt,
     TypeAdapter,
     field_validator,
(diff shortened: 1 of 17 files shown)


pyproject.toml
```diff
@@ -88,6 +88,7 @@ version = "2.0.0"
 
 [tool.mypy]
 explicit_package_bases = true
+mypy_path = "libs/sdk/src"
```
Adam-Hammo, line 91: This was required in order to validate the new discriminated union in `models.py` - there are a couple of previously-unnoticed mypy nits addressed in this PR as the SDK wasn't being strictly type-checked.

libs/sdk/src/destiny_sdk/enhancements.py
```diff
@@ -41,7 +41,7 @@ class EnhancementType(StrEnum):
     RAW = auto()
     """A free form enhancement for arbitrary/unstructured data."""
     FULL_TEXT = auto()
-    """The full text of the reference. (To be implemented)"""
```
mootpointer, line 44: Best deletion.

libs/sdk/src/destiny_sdk/enhancements.py
```diff
@@ -519,6 +519,74 @@ def fingerprint(self) -> str:
         )
 
 
+class FullTextEnhancement(BaseModel):
+    """An enhancement for storing a link to the full text and its metadata."""
```
mootpointer: We probably want to include a note in the docstring which shouts about the fact that the FullTextEnhancement is _not_ for the sharing of copyrighted material without a license. It is to maintain files for validation and to allow permitted Text and Data Mining (TDM) activities.

libs/sdk/tests/unit/test_enhancements.py
```diff
@@ -381,6 +381,74 @@ def test_linked_data_enhancement_non_string_context():
...
+        destiny_sdk.enhancements.FullTextEnhancement()
+
+
+def test_full_text_enhancement_invalid_file_url():
+    with pytest.raises(ValidationError):
+        destiny_sdk.enhancements.FullTextEnhancement(file_url="not-a-url")
+
+
+def test_full_text_enhancement_rejects_http_file_url():
+    """file_url must be https, not http."""
```
mootpointer, line 423: Is HTTPS an assumption while we're the only ones doing the importing? For blob storage it makes sense, but if someone's pulling from a random site, we can't guarantee HTTPS and I'm not sure how much value we get from enforcing it.

Adam-Hammo: Good point. I actually relax this constraint in #680 anyway, because it was making e2e testing tricky, so it's good to know that might be correct!

libs/sdk/src/destiny_sdk/enhancements.py
Adam-Hammo: Good call, have added

app/domain/references/models/models.py
```diff
@@ -298,6 +301,100 @@ class IdentifierLookup(GenericExternalIdentifier):
...
+        ),
+    )
+    is_oa: bool | None = Field(
+        default=None,
+        description=(
+            "If this full text is Open Access. "
+            "May be left as null if this is unknown."
+        ),
+    )
+    license: str | None = Field(
```
danielbreves, line 349: OpenAlex has a [defined list](https://developers.openalex.org/api-reference/licenses#licenses-overview). Should we make this an enum?

tests/unit/domain/references/models/test_models.py
```diff
@@ -48,6 +56,50 @@ def test_reference_create_result_error_str_multiple():
...
+
+    The SDK ships its own FullTextEnhancement with file_url: HttpUrl. The
+    domain variant uses BlobStorageFile. If someone reorders the union or
+    accidentally lists the SDK type in EnhancementContent, this round trip
+    would fail.
+    """
+    full_text = FullTextEnhancementFactory.build()
+    enhancement = EnhancementFactory.build(content=full_text)
+
+    restored = Enhancement.model_validate(enhancement.model_dump())
```
danielbreves, line 71: Would it be worth adding a test for the json mode round trip?

e.g.
```
dumped = enhancement.model_dump(mode="json")
assert isinstance(dumped["content"]["blob"], str)

restored = Enhancement.model_validate(dumped)
assert isinstance(restored.content, FullTextEnhancement)
assert restored.content.blob == full_text.blob
```

danielbreves (approved):

app/domain/references/models/models.py
Adam-Hammo: These are only open access licenses - are there cases where we'd have closed access papers with a license?

danielbreves: Probably I guess? We could include an `other-ca` option and maybe another `str` field to capture `other_license`? It might be overkill, I just thought it could be something people may want to filter by?

Adam-Hammo: Fair enough - do you think we should apply these to `LocationEnhancement.license` too? 

(Might need to check to make sure we haven't imported any non-listed licenses, so maybe not this pass)

Adam-Hammo: I think I'm opening a follow-up for this one. I feel equally cagey about having two license fields in enhancements that mean different things, and can't confidently say what licenses we already have imported in locations. I have however added all of the listed licenses as examples to hopefully nudge data producers in the right direction.

https://github.com/destiny-evidence/destiny-repository/issues/684

Please let me know if you disagree!

danielbreves: Do they mean different things? In any case, it can definitely be deferred. 👍

Adam-Hammo: They were aligned in my mind, but you're probably right to question. Will add that nuance to the ticket.

Issue #619, opened 2026-04-01
`python-jose` causing unfixable downstream security vulnerability alerts


harryjmoss:
I've recently noticed [this security vulnerability](https://github.com/destiny-evidence/fetch-everything-robot/security/dependabot/2) cropping up across a couple of repositories, that in itself seems to be a dependency of `python-jose`.

I see there's [this mitigation in place](https://github.com/destiny-evidence/destiny-repository/blob/c604d65d1794de847426ecb1cb496f6f9aac47ae/libs/sdk/pyproject.toml#L52) already for the repository, but this doesn't automatically get applied to downstream packages that include the destiny SDK as a dependency.

Since it [looks like python-jose has been abandoned](https://github.com/mpdavis/python-jose/) (no updates since May 2025) and there's [no response to the PR intended to fix this issue](https://github.com/mpdavis/python-jose/pull/400), I wondered if we'd be able to switch over to [pyjwt](https://github.com/jpadilla/pyjwt) for the repository? This is already included as a downstream dependency from `msal`, and from a quick skim it seems at least possible to migrate to from `python-jose`. 

Happy to take a time-boxed run at this myself if there's consensus that it's worth pursuing.

PR #672, opened 2026-05-10
Update authlib and joserfc

jack-o-the-green:
Update to versions that have patched existing security vulnerabilities. 

Should fix critical authlib dependabot alert.

### DIFF
--- a/libs/sdk/pyproject.toml
+++ b/libs/sdk/pyproject.toml
@@ -36,7 +36,7 @@ license = "Apache-2.0"
 name = "destiny_sdk"
 readme = "README.md"
 requires-python = ">=3.12, <4"
-version = "0.11.4"
+version = "0.11.5"
 
 [project.optional-dependencies]
 labs = []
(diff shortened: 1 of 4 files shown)


libs/sdk/pyproject.toml
```diff
@@ -23,7 +23,7 @@ authors = [
     {email = "tim.repke@pik-potsdam.de", name = "Tim Repke"},
 ]
 dependencies = [
-    "authlib>=1.4.0,<2",
+    "authlib>=1.6.11,<2",
```
Adam-Hammo, line 26: We'll need to bump the SDK version as well I think to propagate this change downstream.

Also adding as related context, but I don't think necessarily in scope of this PR: https://github.com/destiny-evidence/destiny-repository/issues/619

jack-o-the-green: Nice catch, have bumped!

Ahh yeah this was an interesting one! We're still getting those vulnerabilities because the ecdsa is installed regardless when you install python-jose even though we're using the cryptography backend so the ecdsa vulnerability isn't relevant. 

I do think we'll want to migrate away from python-jose when we get the chance though.

Adam-Hammo (approved):
LGTM

PR #676, opened 2026-05-11
Full text signed urls

### LINKED ISSUE #666: Full Texts: Data Structure

Adam-Hammo:
Follow on to #671 to implement outbound signed URL generation. 

This includes a couple of improvements:
- Reworked signed URL cache. The network dependency lives on the user delegation key, not the actual URL signing.
  - There was already half a delegation key cache implementation in place 🙈 
  - This is realistically a requirement, we don't want to be making `n` calls for these keys when exposing `n` full texts
- Tightened the blob dependency in the anti corruption layer. This reduces the surface area the anti corruption layer has access to to just the URL signing.

Commit history is a bit messy sorry, I pulled this out of #671

### DIFF
--- a/app/domain/references/routes.py
+++ b/app/domain/references/routes.py
@@ -435,9 +435,12 @@ async def lookup_references(
     identifier_lookups = anti_corruption_service.identifier_lookups_from_sdk(
         identifiers
     )
-    return await anti_corruption_service.references_to_sdk(
-        await reference_service.get_references_from_identifiers(identifier_lookups)
-    )
+    return [
+        await anti_corruption_service.reference_to_sdk(reference)
+        for reference in await reference_service.get_references_from_identifiers(
+            identifier_lookups
+        )
+    ]
 
 
 @enhancement_request_automation_router.post(
--- a/app/domain/references/routes.py
+++ b/app/domain/references/routes.py
@@ -354,7 +354,7 @@ async def search_references(
         if search_result.hits
         else []
     )
-    return anti_corruption_service.two_stage_reference_search_result_to_sdk(
+    return await anti_corruption_service.two_stage_reference_search_result_to_sdk(
         search_result, references
     )
 
@@ -376,7 +376,7 @@ async def get_reference(
 ) -> destiny_sdk.references.Reference:
     """Get a reference by id."""
     reference = await reference_service.get_reference(reference_id)
-    return anti_corruption_service.reference_to_sdk(reference)
+    return await anti_corruption_service.reference_to_sdk(reference)
 
 
 class IdentifierLookupQueryParams(BaseModel):
@@ -435,12 +435,9 @@ async def lookup_references(
     identifier_lookups = anti_corruption_service.identifier_lookups_from_sdk(
         identifiers
     )
-    return [
-        anti_corruption_service.reference_to_sdk(reference)
-        for reference in await reference_service.get_references_from_identifiers(
-            identifier_lookups
-        )
-    ]
+    return await anti_corruption_service.references_to_sdk(
+        await reference_service.get_references_from_identifiers(identifier_lookups)
+    )
 
 
 @enhancement_request_automation_router.post(
--- a/app/domain/references/routes.py
+++ b/app/domain/references/routes.py
@@ -100,7 +100,7 @@ def reference_anti_corruption_service(
     blob_repository: Annotated[BlobRepository, Depends(blob_repository)],
 ) -> ReferenceAntiCorruptionService:
     """Return the reference anti-corruption service."""
-    return ReferenceAntiCorruptionService(blob_repository=blob_repository)
+    return ReferenceAntiCorruptionService(sign_url=blob_repository.get_signed_url)
 
 
 def robot_anti_corruption_service() -> RobotAntiCorruptionService:
--- a/app/domain/references/services/anti_corruption_service.py
+++ b/app/domain/references/services/anti_corruption_service.py
@@ -25,16 +25,21 @@ from app.domain.references.models.models import (
 )
 from app.domain.service import GenericAntiCorruptionService
 from app.persistence.blob.models import BlobSignedUrlType
-from app.persistence.blob.repository import BlobRepository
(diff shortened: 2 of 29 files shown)


app/domain/references/services/anti_corruption_service.py
```diff
@@ -22,16 +24,21 @@
...
 from app.persistence.es.persistence import ESSearchResult
 
 
 class ReferenceAntiCorruptionService(GenericAntiCorruptionService):
     """Anti-corruption service for translating between Reference domain and SDK."""
 
-    def __init__(self, blob_repository: BlobRepository) -> None:
-        """Initialize the anti-corruption service."""
-        self._blob_repository = blob_repository
+    def __init__(self, sign_url: URLSigner) -> None:
```
danielbreves, line 34: Could we update our fixture as well: https://github.com/destiny-evidence/destiny-repository/blob/6352f87ce930be40ae0c0b53394e09acbf1840fc/tests/unit/domain/references/models/test_models.py#L105

app/domain/references/services/anti_corruption_service.py
```diff
@@ -112,11 +120,36 @@ def external_identifier_from_sdk(
...
+                content = await self.full_text_enhancement_content_to_sdk(
+                    enhancement.content
+                )
+                return destiny_sdk.references.Enhancement(
+                    **enhancement.model_dump(exclude={"content"}),
+                    content=content,
+                )
             return destiny_sdk.references.Enhancement.model_validate(
                 enhancement.model_dump()
             )
```
danielbreves: Potentially a tad simpler and safer?
```suggestion
            dumped = enhancement.model_dump()
            if enhancement.content.enhancement_type == EnhancementType.FULL_TEXT:
                dumped["content"] = (await self.full_text_enhancement_content_to_sdk(
                    enhancement.content
                )).model_dump()
            return destiny_sdk.references.Enhancement.model_validate(dumped)
)
```

app/domain/references/services/anti_corruption_service.py
```diff
@@ -321,14 +354,19 @@ def identifier_lookups_from_sdk(
...
+    async def two_stage_reference_search_result_to_sdk(
         self,
         search_result: ESSearchResult,
         references: list[Reference],
     ) -> destiny_sdk.references.ReferenceSearchResult:
         """Convert a search result and retrieved references to the SDK model."""
         try:
             hit_order = {hit.id: i for i, hit in enumerate(search_result.hits)}
+            sdk_references = [
+                await self.reference_to_sdk(reference)
```
danielbreves, line 361: Could we use `asyncio.gather` here?

app/domain/references/routes.py
```diff
@@ -436,7 +436,7 @@ async def lookup_references(
         identifiers
     )
     return [
-        anti_corruption_service.reference_to_sdk(reference)
+        await anti_corruption_service.reference_to_sdk(reference)
```
danielbreves, line 439: Could we use `asyncio.gather` here?

danielbreves (approved):

app/domain/references/services/anti_corruption_service.py
Adam-Hammo: Yes, but I'm not sure we actually get much benefit from it and I'm wary of implying that we do. Signed URL generation is CPU-bound, not network bound

app/domain/references/routes.py
Adam-Hammo: Same as https://github.com/destiny-evidence/destiny-repository/pull/676#discussion_r3237796607

Issue #661, opened 2026-05-06
Support unbounded searching in the persistence layer


Adam-Hammo:
Our current pagination implementation uses the simple `from` and `size` parameters from Elasticsearch, which cap at 10,000 results.

To view more results, we need to support [`search_after`](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/paginate-search-results#search-after). This is incompatible with `from/size` and needs to be forked in query construction.

As in the linked example, we should always provide `"id"` as the final sort parameter as a deterministic tiebreaker.

**Early design decision**
The Elasticsearch DSL does expose an `[iterate()](https://elasticsearch-dsl.readthedocs.io/en/stable/async_api.html#elasticsearch_dsl.AsyncSearch.iterate)` helper to do all of this. However, that closes the door to any extensibility, notwithstanding: custom pages, synchronous API-based iterating, PIT error handling, limits, knowing total size before the end of the iteration. Hence we implement ourselves.

**Scope**
Provide a generic interface in `app/persistence/es/repository.py` to create a [`PIT`](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-open-point-in-time) and iterate over an arbitrary number of documents. 
We should still allow a custom limit to be passed in.
This should use `[track_total_hits](https://www.elastic.co/docs/solutions/search/the-search-api#track-total-hits)`.

My initial thoughts are this could be called something like the below signature (adapted from `search_with_query_string`, and of course not necessarily prescriptive):
```python
async def scan_with_query_string(  # noqa: PLR0913
    self,
    query: str,
    page_size: int = 20,  # must be <=10,000
    limit: int | None = None,
    fields: Sequence[str] | None = None,
    sort: list[str | dict[str, Any]] | None = None,  # assert ["id"] in this list, default to `["id"]` if None
    filter_clauses: Sequence[Query] | None = None,
    *,
    parse_document: bool = False,
) -> AsyncGenerator[ESSearchResult, None]:
```
Note that none of the ES internals (sort values, PIT, etc) are exposed outside of this function at this point (yet!).
The Elasticsearch DSL may have [useful helper methods](https://elasticsearch-dsl.readthedocs.io/en/stable/async_api.html#elasticsearch_dsl.AsyncSearch.search_after).

**Optional scope for discussion**
[...]

PR #678, opened 2026-05-12
Add reference JSONL download endpoint (#674)

### LINKED ISSUE #674: Allow export of jsonl files of references from destiny-repository
## Context 

To implement excel exports as a first pass, we need to be able to download jsonl files of references from destiny-repository. 

## Technical Scope
* Implement a `references/download/` endpoint on destiny repository that triggers the creation of a jsonl file of all references included in a search 
    * The API should match the `/references/search/` API but it should not include parameters for page size/page number. Instead the endpoint will handle getting all matching references itself. 
    * The reference structure should also match the search endpoint
    * The endpoint should return a 202 ACCEPTED on successful call, and create the file as a background task. 
* Implement an endpoint that can be polled to check the status of the provided download file. This should return a pending state until the file is ready, and which point it should return a signed download url. 
* The endpoints do not need a new authentication scope, they can use the same authentication scope as search. 

## Notes
We will not fix the 10,000 reference search limit for this work, that will be handled separately in https://github.com/destiny-evidence/destiny-repository/issues/661.

danielbreves:
## Summary

Closes destiny-evidence/destiny-repository#674.

- New `POST /references/download/` queues a Taskiq job that produces a JSONL of every reference matching the search filters; returns `202 ACCEPTED` with a pending job id.
- New `GET /references/download/{id}/` is polled until status is `completed`, at which point it returns a signed download URL. Failures surface via `status: failed` + an `error` message.
- Endpoint mirrors `/references/search/` filters (`q`, `annotation`, `start_year`, `end_year`, `sort`) minus pagination; each JSONL line matches the same `destiny_sdk.references.Reference` shape as the search response.
- Reuses the existing `REFERENCE_READER` scope.
- Modelled on `EnhancementRequest`: new `ReferenceDownload` domain/SQL model + Alembic migration `88264d67fc82`, anti-corruption mapping that resolves the signed URL on completion, and a streaming upload via the existing `FileStream` + `_get_jsonl_deduplicated_references` plumbing.
- `SearchService.search_with_query_string` now takes an optional `page_size` (defaults to 20). The download path uses `page_size=1000`, cutting a full 10k export from ~500 ES requests to ~10. User-facing `/references/search/` is unchanged.

The 10k ES `max_result_window` ceiling is inherited from the search path and is explicitly out of scope here — tracked separately in destiny-evidence/destiny-repository#661.

## Test plan

- [x] Unit test for paging (`test_collect_reference_download_ids_pages_at_1000`) — verifies `page_size=1000`, page increment, and early-exit on partial page.
- [x] Integration test happy path (`test_request_reference_download_happy_path`) — POST returns 202 pending, task runs via `InMemoryBroker`, GET returns completed with `result_url` + `n_references`.
- [x] Integration test failure path (`test_request_reference_download_marks_failed_on_error`) — exception during collection lands the job in `failed` with the error surfaced.
- [x] Integration test pending state (`test_get_reference_download_pending_has_no_url`) — a freshly inserted pending row exposes no signed URL.
- [x] Manual end-to-end smoke against a real search once deployed to a dev env (verify SAS URL downloads a JSONL whose line count equals `n_references`).


Adam-Hammo:
Just to note, not to block - depending on ordering, we're gonna have a few considerations between some in-flight full text PRs and this:
- Signed URL interface change in the anti-corruption layer
- `BlobStorageFile` interface change
- Need to add access control to this service
  - We'll need to find a way to pass it to the task. ~Probably a follow-up PR~ We can just achieve this by adding `entitlements` to the function signature.

I'm happy to handle these :)

app/domain/references/routes.py
```diff
@@ -359,9 +371,108 @@ async def search_references(
...
+
+
+@download_router.get(
+    "/{reference_download_id}/",
+    description=(
+        "Get the status of a reference download job. Once `status` is "
+        "`completed`, the response includes a signed `result_url` for the "
+        "produced JSONL file, and `truncated: true` if the matching set "
+        "exceeded the 10,000-result cap (the file contains only the first "
+        "10,000 matches). The URL is re-signed on each call, so an expired "
```
Adam-Hammo: Should we string build with `SearchService.MAX_RESULT_WINDOW`?

libs/sdk/src/destiny_sdk/references.py
```diff
@@ -73,3 +74,50 @@ class ReferenceSearchResult(SearchResultMixIn, BaseModel):
...
+    )
+    n_references: int | None = Field(
+        default=None,
+        description="The number of references included in the produced file.",
+    )
+    truncated: bool = Field(
+        default=False,
+        description=(
+            "Whether the matching result set exceeded the 10,000-result cap. "
+            "When true, the JSONL contains only the first 10,000 matches."
```
Adam-Hammo: Can we remove reference to 10,000 to avoid needing to update SDK models alongside API functionality?

app/domain/references/routes.py
```diff
@@ -359,9 +371,108 @@ async def search_references(
...
+    reference_download = await reference_service.get_reference_download(
+        reference_download_id
+    )
+    return await anti_corruption_service.reference_download_to_sdk(reference_download)
+
+
+# NB it's important these occur before defining `/references/{reference_id}/` route
 # to avoid route conflicts. Order matters for FastAPI route matching.
 reference_router.include_router(search_router)
+reference_router.include_router(download_router)
```
Adam-Hammo: A couple of nits on the route: 
- should download**s** be plural?
- should the search route include a `/search/` resource, in case we want to download by some other criteria in the future?
  - `/references/downloads/search/?...` and `/references/downloads/{id}/`?

And a particularly annoying one: could/should it be called `exports` instead throughout? That more implies to me the level of work being done.

app/domain/references/service.py
```diff
@@ -1227,6 +1230,211 @@ async def search_references(
...
+            )
+            if reference_download.start_year or reference_download.end_year
+            else None
+        )
+
+        page_size = 1000
+        max_results = SearchService.MAX_RESULT_WINDOW
+        max_page = max_results // page_size
+
+        seen_ids: dict[UUID, None] = {}
```
Adam-Hammo: Is there a reason not to use a `set`?

app/domain/references/service.py
```diff
@@ -1227,6 +1230,211 @@ async def search_references(
...
+                page_size=page_size,
+                annotations=reference_download.annotation_filters,
+                publication_year_range=publication_year_range,
+                sort=reference_download.sort,
+            )
+            if first_total is None:
+                first_total = search_result.total
+            if not search_result.hits:
+                break
+            # ES from/size can repeat an id across pages if the index changes mid-walk.
```
Adam-Hammo: Tagging #661 which can mitigate this with PIT IDs

app/domain/references/service.py
```diff
@@ -1227,6 +1230,211 @@ async def search_references(
...
+        reference_ids: list[UUID],
+        blob_repository: BlobRepository,
+    ) -> tuple[BlobStorageFile, int]:
+        """Stream matching references to blob storage as JSONL."""
+        chunk_size = settings.upload_file_chunk_size_override.get(
+            UploadFile.REFERENCE_DOWNLOAD,
+            settings.default_upload_file_chunk_size,
+        )
+        # A reference id with no SQL row produces no JSONL line, so the file
+        # can have fewer lines than `reference_ids` has entries.
```
Adam-Hammo: How does this occur in practice?

app/domain/references/service.py
```diff
@@ -1227,6 +1230,211 @@ async def search_references(
...
+        publication_year_range = (
+            PublicationYearRange(
+                start=reference_download.start_year,
+                end=reference_download.end_year,
+            )
+            if reference_download.start_year or reference_download.end_year
+            else None
+        )
+
+        page_size = 1000
```
Adam-Hammo: Is there any reason to not just do 10,000?

app/domain/references/service.py
```diff
@@ -1227,6 +1230,211 @@ async def search_references(
...
+            ],
+        )
+        result_file = await blob_repository.upload_file_to_blob_storage(
+            content=file_stream,
+            path="reference_downloads",
+            filename=f"{reference_download_id}.jsonl",
+        )
+        return result_file, n_emitted
+
+    async def run_reference_download(
```
Adam-Hammo: For another example of why we might consider `export` naming - we're not downloading anything here

Adam-Hammo (commented):
Looking good, mostly naming things!

libs/sdk/src/destiny_sdk/references.py
Adam-Hammo: Also we'll need to bump the SDK version (you may have been deliberately holding off)

app/domain/references/service.py
danielbreves: Not really, I started with this to include support for paginating through, but given we're revisiting the 10k limit later it's probably premature optimisation.

app/domain/references/service.py
danielbreves: It shouldn't in practice since we don't remove references, removed now we're fetching it all in one go.

app/domain/references/service.py
```diff
@@ -1227,6 +1230,182 @@ async def search_references(
             sort=sort,
         )
 
+    @sql_unit_of_work
+    async def request_reference_export(
```
jack-o-the-green: Do we think exports deserves to be it's own service?

jack-o-the-green: I know we handle reference ingestion in this file, so happy to be convinced either way

app/domain/references/models/models.py
```diff
@@ -398,6 +411,54 @@ def n_references(self) -> int:
         return len(self.reference_ids)
 
 
+class ReferenceExport(DomainBaseModel, SQLAttributeMixin):
```
jack-o-the-green: `nitpick` do we think we should be calling these `SearchExport` instead. Given it's based on the search interface?

libs/sdk/src/destiny_sdk/references.py
```diff
@@ -73,3 +74,51 @@ class ReferenceSearchResult(SearchResultMixIn, BaseModel):
...
+        description=(
+            "Whether the matching result set exceeded the server's result-window "
+            "cap. When true, the JSONL contains only the first window's worth of "
+            "matches."
+        ),
+    )
+    error: str | None = Field(
+        default=None,
+        description="Error encountered while producing the export.",
+    )
```
jack-o-the-green, line 126: Might be out of scope for this PR but do we want to add a function to our client for doing an export? We support search there.

jack-o-the-green: It's not necessary to get these exports working given the evidence-repo-ui doesn't using the sdk, so happy to leave off.

jack-o-the-green (commented):
Mostly musing from me, feel free to ignore.

app/domain/references/service.py
danielbreves: I agree actually, good idea! 👍

jack-o-the-green (approved):
LGTM - thanks for breaking the service out. 

Wanted to note that I think there will be migration collisions between this and https://github.com/destiny-evidence/destiny-repository/pull/686/changes, so will need to update the down revision of whichever migration goes in second. 

Also also full text access control will need to be added to the export endpoint.

app/domain/references/models/models.py
danielbreves: Mmm interesting, now I'm wondering if the routes should be the other way around too - `/references/search/exports/` POST + `/references/search/exports/{id}/`

jack-o-the-green: Doesn't sound unreasonable to me `/references/exports` I'd expect to take a list of UUIDs really, so this feels like a good change.

PR #679, opened 2026-05-12
Blob repository copy function

### LINKED ISSUE #668: Full Texts: Ingestion
Importing processes provide `FullTextEnhancement`s in the incoming `.jsonl` file. The `file_url` in this case points to the full text hosted by them (much like the `.jsonl` itself).

When the incoming `Reference` is parsed, the file is downloaded from the importer's blob and stored on the repository's blob storage. If `mime_type`, `byte_size` and `sha256_checksum` were provided, these are used to verify the copy, otherwise they are calculated and saved. The `FullTextEnhancement` is then stored with a `BlobStorageFile` pointing at the repository's copy of the file.

Adam-Hammo:
First part of #668 - adds the ability for `BlobRepository` to copy files, streaming them through memory to calculate the checksum and size.

This required:
- a new `stream_chunks` method to stream raw data
  - `stream_file` has been moved to the parent class and calls `stream_chunks`
- a new `RemoteBlobClient` class for fetching remote blobs
- moves `content_type` handling out of `BlobStorageFile` to allow it to be explicitly passed as well as inferred
- a couple of security mitigations, see also https://github.com/destiny-evidence/destiny-repository/issues/668#issuecomment-4438371217

### DIFF
--- a/app/persistence/blob/client.py
+++ b/app/persistence/blob/client.py
@@ -31,6 +31,7 @@ class GenericBlobStorageClient(ABC):
         self,
         content: FileStream | BytesIO | AsyncIterator[bytes],
         file: BlobStorageFile,
+        content_type: str | None = None,
     ) -> None:
         """
         Upload a file to the blob storage.
@@ -39,6 +40,10 @@ class GenericBlobStorageClient(ABC):
         :type content: FileStream | BytesIO | AsyncIterator[bytes]
         :param file: The file to upload.
         :type file: BlobStorageFile
+        :param content_type: Optional MIME type to attach to the uploaded
+            object. If not provided, implementations infer it from
+            ``file.filename``.
+        :type content_type: str | None
         """
 
     @trace_blob_client_generator(tracer)
--- a/app/persistence/blob/client.py
+++ b/app/persistence/blob/client.py
@@ -93,6 +93,7 @@ class GenericBlobStorageClient(ABC):
         self,
         file: BlobStorageFile,
         interaction_type: BlobSignedUrlType,
+        content_disposition: str | None,
     ) -> str:
         """
         Generate a signed URL for the file in blob storage.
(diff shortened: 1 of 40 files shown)


app/persistence/blob/clients/azure.py
```diff
@@ -65,46 +71,47 @@ def __init__(
...
         blob_client = self.blob_service_client.get_blob_client(
             container=file.container, blob=f"{file.path}/{file.filename}"
         )
+        content_settings = ContentSettings(
+            content_type=content_type or infer_content_type(file.filename)
+        )
         try:
             if isinstance(content, FileStream):
-                await blob_client.upload_blob(content.stream(), overwrite=True)
+                await blob_client.upload_blob(
```
jack-o-the-green: Can we just collapse this to 

```
await blob_client.upload_blob(
    content.stream() if isinstance(content, FileStream) else content,
    overwrite=True,
    content_settings=content_settings,
)
```

app/persistence/blob/client.py
```diff
@@ -29,20 +29,43 @@ class GenericBlobStorageClient(ABC):
...
+        """
+        Stream a file as raw byte chunks from the blob storage.
+
+        :param file: The file to stream.
+        :type file: BlobStorageFile
+        :return: An async generator that yields byte chunks from the file.
+        :rtype: AsyncGenerator[bytes, None]
+        """
+        # https://github.com/python/mypy/issues/5070
+        __here_be_dragons = False
```
jack-o-the-green, line 64: Aw I remember her 💚

app/persistence/blob/client.py
```diff
@@ -76,6 +102,11 @@ async def generate_signed_url(
         :type file: BlobStorageFile
         :param interaction_type: The type of interaction (upload or download).
         :type interaction_type: BlobSignedUrlType
+        :param content_disposition: Override for the Content-Disposition response
+            header served when the signed URL is fetched. Set to ``"attachment"``
+            for untrusted content (e.g. full-text blobs) to neutralize inline
+            browser rendering. Ignored for ``UPLOAD``.
```
jack-o-the-green: Inordinately pedantic, but the implementations of this are "included for download" not "ignored for upload". So a third blob URL type would not get the content_disposition override, whereas this implies it would.

jack-o-the-green (approved):
Looks good to me, nitpicks only

PR #680, opened 2026-05-12
Full text ingestion plumbing

### LINKED ISSUE #668: Full Texts: Ingestion

Adam-Hammo:
Closes #668 

Adds full text ingestion handling to import and enhancement flows. 

- users submit a full text enhancement with URL point to their copy
- repository verifies and copies to local blob storage

That's about it. Big diff for a little change unfortunately, lots of testing.


infra/app/main.tf
```diff
@@ -387,6 +387,7 @@ resource "azurerm_servicebus_queue" "taskiq" {
   namespace_id = azurerm_servicebus_namespace.this.id
 
   partitioning_enabled = true
+  lock_duration        = "PT5M"
```
Adam-Hammo, line 390: With this change, import reference tasks may take longer to complete as they need to copy and verify the file.

We ideally don't want to set up auto lock renewers too much as we've had low-level errors with the azure service bus SDK when doing so.

Azure Service Bus default lock was 1 minute, this increases it to the maximum 5 minutes. I don't think the extra latency is a significant concern in a truly lost task.

Let's hope it's enough.

app/persistence/blob/repository.py
```diff
@@ -241,8 +241,8 @@ async def copy(
         self,
         source: BlobStorageFile,
         destination: BlobStorageFile,
-        content_type: str | None = None,
         max_bytes: int | None = None,
+        content_type: str | None = None,
```
Adam-Hammo, line 245: Merge conflict juggling 😩

app/domain/references/models/sql.py
```diff
@@ -307,6 +307,15 @@ def from_domain(cls, domain_obj: DomainEnhancement) -> Self:
         timestamps when converting from the domain. They're purely managed
         by the persistence model.
         """
+        if (
+            domain_obj.content.enhancement_type == EnhancementType.FULL_TEXT
+            and domain_obj.content.blob.is_remote
```
jack-o-the-green, line 320: Confirming - `is_remote` is implemented as `is http/https url`. This exception is correct here because we only generate signed urls when convering to an sdk model, so there's no chance of one of our signed urls hitting here? 

Just want to make sure there's not an instance of our signed url ending up here, because in that case it would be incorrect to say that the full text enhancement isn't in our storage.

app/domain/references/services/enhancement_service.py
```diff
@@ -64,6 +76,102 @@ def __init__(
...
+            computed value.
+        """
+        content = enhancement.content
+        if content.enhancement_type != EnhancementType.FULL_TEXT:
+            msg = (
+                f"store_full_text called with non-full-text enhancement "
+                f"{enhancement.id} (type {content.enhancement_type})"
+            )
+            raise ValueError(msg)
+        if not content.blob.is_remote:
```
jack-o-the-green, line 104: Thinking of horrible things I could do as a user. 

What if I pulled out existing references with full text enhancements and then re-imported them, passing through the signed urls for our previously stored full texts. That's technically a non-remote block but I don't think it would be caught by this and we'd probably make more copies of the full text. That's a pretty nasty usage pattern though!

app/domain/references/services/enhancement_service.py
```diff
@@ -64,6 +76,102 @@ def __init__(
...
+            msg = (
+                f"store_full_text called with non-remote blob for enhancement "
+                f"{enhancement.id}: {content.blob.to_uri()}"
+            )
+            raise FullTextIngestionError(msg)
+
+        source = content.blob
+        extension = mimetypes.guess_extension(content.mime_type) or ".bin"
```
jack-o-the-green, line 112: Why `.bin` as the particular default here?

app/domain/references/service.py
```diff
@@ -515,6 +519,27 @@ async def ingest_reference(
         reference_create_result.reference_id = reference.id
         trace_attribute(Attributes.REFERENCE_ID, str(reference.id))
 
+        kept_enhancements: list[Enhancement] = []
```
jack-o-the-green: This function is a pretty big boy now, wondering if we can extract the storage of full text enhancements to a private method that is called here?

jack-o-the-green (commented):
Looking good, minor comments

app/domain/references/services/enhancement_service.py
Adam-Hammo: It's the Python stdlib `mimetypes` for `application/octet-stream`, which in turn is the [default for "arbitrary binary data"](https://datatracker.ietf.org/doc/html/rfc2046#section-4.5.1).

app/domain/references/services/enhancement_service.py
Adam-Hammo: You're right that there is nothing stopping someone from re-pointing our own storage account files in ingestion, and it would be considered `remote` and reingested. 

We could consider detecting this based off the URL hostname and finding the original blob to point to instead, this is probably follow-up work though and I personally tend towards YAGNI. There are other considerations with pointing two references to the same blob too - if we delete one someday we have to track the number of enhancements still pointing to it before knowing if we can delete the blob.

app/domain/references/models/sql.py
Adam-Hammo: Correct - this is a defensive guard to ensure that any full texts we store now belong to our storage, not remotely. When we store the file in our own blob, we switch the blob to `azure/minio` pointers instead of the URLs.

> so there's no chance of one of our signed urls hitting here?

The defensive check really is designed to catch this - if a future implementation creates that opportunity, the error is thrown.

app/domain/references/services/enhancement_service.py
jack-o-the-green: Yeah that's fair - happy to ignore this! 

I think possibly those user re-imports would be caught by our discarding of exact duplicates anyway.

jack-o-the-green (approved):
LGTM

PR #681, opened 2026-05-13
Full text access control

### LINKED ISSUE #669: Full Texts: Access Control
Add `Principal` and `Capability` primitives to the existing auth dependency. Capabilities are derived from the user's scopes. In our case, we'd give the `Principal` the `READ_FULL_TEXT` capability if it has the new scope (or, optionally, if it's a robot).

```python
class Capability(StrEnum):
    READ_FULL_TEXT = auto()

class Principal:
    subject_id: str           
    subject_type: "USER" | "SERVICE" | "ROBOT"
    capabilities: frozenset[Capability]
```

A new AccessControlService is dependency injected with capabilities and AntiCorruptionService.

```python
class AntiCorruptionService:
    def reference_to_sdk(self, ref) -> sdk.Reference:
        return sdk.Reference(...)


class AccessControlService:
    def __init__(self, anti_corruption_service, capabilities):
        self._anti_corruption_service = anti_corruption_service
        self._capabilities = capabilities

    def reference_to_sdk(self, ref) -> sdk.Reference:
        ref.enhancements = [
            e for e in ref.enhancements
            if e.content.enhancement_type != EnhancementType.FULL_TEXT
            or Capability.READ_FULL_TEXT in self._capabilities
        ]
        return self._anti_corruption_service.reference_to_sdk(ref)
```
(Probably with renaming one/both of the reference_to_sdk)

Adam-Hammo:
Closes #669

- Parsing of principal and deriving of entitlements at auth time
- An `AccessControlService` to do just-in-time redaction
  - New typing enforces that this is called before the anti-corruption layer as a safeguard
- The necessary role and scope in Azure, unassigned
  - Pairs with https://github.com/destiny-evidence/destiny-shared-infra/pull/35

Builds off #680


app/api/auth.py
```diff
@@ -77,6 +79,53 @@ class HMACClientType(StrEnum):
...
+    BYPASS = auto()
+
+
+class Entitlement(StrEnum):
+    """Enum describing the entitlements that can be granted to users or clients."""
+
+    FULL_TEXT = auto()
+
+
+_ROBOT_ENTITLEMENTS: frozenset[Entitlement] = frozenset({Entitlement.FULL_TEXT})
```
Adam-Hammo: TBC, maybe empty set.

jack-o-the-green: I'm cautious of granting robots blanket entitlement to full text. Currently we control all the robots, but eventually we'll have a bunch of them and if they all get full text access by default I can see an attack vector where someone just deploys a robot and has it download any full texts it receives. 

Likely an overbearing concern at this stage.

app/domain/references/services/access_control_service.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        Return the principal's redacted view of a reference.
+
+        Drops enhancements the principal is not entitled to read.
+        """
+        return RedactedReference(
+            reference.model_copy(
+                update={
+                    "enhancements": [
+                        enhancement
```
jack-o-the-green: This implementation feels too wide to me/the documentation is implying a broader implementation . If I'm reading it right this loop does the following 

For each enhancement include it if 
* It's not a full text enhancement or 
* the principal is entitled to view full text enhancements. 

If we add other enhancement type that should be redacted, these will automatically be included by both conditions on the loop and _not_redacted. 

Either I think we should call this "redact_full_text" and update the documentation to be specific to this case (easiest) or implement a more general redacting solution, probably by mapping enhancement type to entitlements.

app/domain/references/routes.py
```diff
@@ -305,6 +326,9 @@ async def search_references(
     anti_corruption_service: Annotated[
         ReferenceAntiCorruptionService, Depends(reference_anti_corruption_service)
     ],
+    access_control_service: Annotated[
```
jack-o-the-green, line 365: Just highlighting that we'll need to add the access control service to downloads/exports also in Daniel's existing PR.

app/domain/service.py
```diff
@@ -16,6 +17,14 @@ class GenericAntiCorruptionService:
     """
 
 
+class AccessControlService:
```
jack-o-the-green: Other services in this file are prefixed `Generic`, should this be too?

infra/app/auth.tf
```diff
@@ -52,6 +54,15 @@ resource "azuread_application" "destiny_repository" {
       user_consent_description   = "Allow you to view references"
       user_consent_display_name  = "Reference Reader"
     }
+    oauth2_permission_scope {
```
jack-o-the-green, line 59: Comment to remind us to confirm any auth_v2 changes needed.

jack-o-the-green: auth_v2 changes not needed, will raise separate PR to remove the unused files

app/domain/references/services/access_control_service.py
Adam-Hammo: I think the instinct on a separate `redact_full_text` is probably right. Some sort of generalised mapping makes assumptions that we only care about enhancement type - in reality we probably care about other signals down the road eg open access. Will implement

jack-o-the-green (approved):
LGTM

PR #686, opened 2026-05-14
Robot entitlement provisioning

### LINKED ISSUE #669: Full Texts: Access Control

Adam-Hammo:
Adds per-robot entitlement provisioning.

Using the existing endpoints, users with `robot.entitlement.writer` can grant/revoke entitlements for specific robots.


app/domain/robots/models/sql.py
```diff
@@ -30,6 +31,10 @@ class Robot(GenericSQLPersistence[DomainRobot]):
 
     owner: Mapped[str] = mapped_column(String, nullable=False)
 
+    entitlements: Mapped[list[str]] = mapped_column(
```
jack-o-the-green: I'm pretty sure you can do a Mapped[list[Entitlement]] here - similar to https://github.com/destiny-evidence/destiny-repository/blob/7a4849c7eac8f99cb1564bd8710bf78b12253dc3/app/domain/references/models/sql.py#L355

app/domain/robots/models/sql.py
```diff
@@ -49,6 +54,7 @@ def from_domain(cls, domain_obj: DomainRobot) -> Self:
             description=domain_obj.description,
             name=domain_obj.name,
             owner=domain_obj.owner,
+            entitlements=sorted(e.value for e in domain_obj.entitlements),
```
jack-o-the-green: What value do we get out of the sorting here?

jack-o-the-green (approved):
LGTM - minor comments on sql model

app/domain/robots/models/sql.py
Adam-Hammo: None at all!

PR #689, opened 2026-05-17
Add priority queue for export processing

### LINKED ISSUE #682: Run reference export jobs on a dedicated high-priority queue
[PR #678](https://github.com/destiny-evidence/destiny-repository/pull/678) queues `POST /references/exports/search/` jobs onto the shared Taskiq broker, so an export can sit behind a long-running import for hours. That's poor UX for an interactive "click export, wait for file" operation.

Consider a dedicated queue / higher priority for exports so they aren't head-of-line-blocked by batch work.

jack-o-the-green:
Closes #682 

## Context 
Exports used to run on the same task queue as all other tasks, meaning they could get caught behind large imports and not process for an extended period. As exports are a user triggered tasks we need appropriate response times and therefore need to process them ahead of other tasks. Enter priority queue. 

Azure Servicebus is first-in first-out and does not support message ordering, so this implements a separate priority queue. Routing to each queue is determined by set task priority. Exports are the only high priority task. 

## What's Changed
* Tasks now take an optional priority, defaulting to normal 
* Additional priority queue created in the same servicebus namespace 
* Works now poll priority queue prior to polling normal queue
    * Also reduced the max wait time on polling the normal queue to ensure export tasks start in a timely manner. 
* High priority tasks are sent to the priority queue by the service bus broker. 

Prioritization is implemented at the broker level based on set task priorities. 

## Want to test?

This is currently deployed in development, so you can trigger a download and see the tasks hitting the priority queue.

### DIFF
--- a/app/core/azure_service_bus_broker.py
+++ b/app/core/azure_service_bus_broker.py
@@ -32,6 +32,7 @@ from taskiq import AckableMessage, AsyncBroker, BrokerMessage
 from app.core.config import get_settings
 from app.core.exceptions import MessageBrokerError, MessageTooLargeError
 from app.core.telemetry.logger import get_logger
+from app.core.telemetry.taskiq import TaskPriority
 
 _T = TypeVar("_T")
 
@@ -83,8 +84,9 @@ class AzureServiceBusBroker(AsyncBroker):
             If None, the namespace parameter must be provided.
         :param namespace: The fully qualified namespace of the Service Bus.
             Used with DefaultAzureCredential if connection_string is None.
-        :param queue_name: queue that used to get incoming messages.
-        :param connection_kwargs: additional keyword arguments.
+        :param queue_name: queue used to get normal priority incoming messages with.
+        :param priority_queue_name: queue used to get high priority incoming
+            messages with.
         """
         super().__init__()
 
@@ -165,6 +167,26 @@ class AzureServiceBusBroker(AsyncBroker):
         if self.credential:
             await self.credential.close()
 
+    def _resolve_priority(self, message: BrokerMessage) -> TaskPriority:
+        """
+        Resolve the ``priority`` label on a broker message to a TaskPriority.
+
+        Unrecognised or unparseable values log a warning and fall back to
+        ``TaskPriority.NORMAL`` so an unknown sender can never block routing.
+        """
+        raw_priority = parse_val(int, message.labels.get("priority"))
+        if raw_priority is None:
+            return TaskPriority.NORMAL
+        try:
+            return TaskPriority(raw_priority)
+        except ValueError:
+            logger.warning(
+                "Unknown priority value, defaulting to NORMAL",
+                priority=raw_priority,
+                task_id=message.task_id,
+            )
+            return TaskPriority.NORMAL
+
     async def kick(self, message: BrokerMessage) -> None:
         """
         Send message to the queue.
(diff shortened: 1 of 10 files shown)


app/core/azure_service_bus_broker.py
```diff
@@ -190,134 +233,151 @@ async def kick(self, message: BrokerMessage) -> None:
...
-                        else False
-                    ):
-                        data = gzip.decompress(data)
-
-                    ackable = AckableMessage(
-                        data=data,
-                        ack=ack_message,
+                async with self._receive_lock:
+                    batch_messages = await self.receiver.receive_messages(
+                        max_wait_time=2
```
jack-o-the-green: Would love thoughts on whether we should move max_wait_time to a configurable variable. Haven't done here as minimizing changes.

danielbreves: It could be worth making it configurable so we can change quickly, but with a default so we don't need to actually configure it until it's needed?

app/core/azure_service_bus_broker.py
```diff
@@ -140,31 +156,59 @@ async def shutdown(self) -> None:
...
             await self.credential.close()
 
+    def _resolve_priority(self, message: BrokerMessage) -> TaskPriority:
+        """
+        Resolve the ``priority`` label on a broker message to a TaskPriority.
+
+        Unrecognised or unparseable values log a warning and fall back to
+        ``TaskPriority.NORMAL``.
+        """
+        raw_priority = parse_val(int, message.labels.get("priority"))
```
danielbreves: Should we log a warning if `message.labels.get("priority")` is set, but not parseable? e.g. `"HIGH"`.

app/core/azure_service_bus_broker.py
```diff
@@ -190,134 +233,151 @@ async def kick(self, message: BrokerMessage) -> None:
...
+    def _build_ackable(
+        self,
+        sb_message: ServiceBusReceivedMessage,
+        receiver: ServiceBusReceiver,
+    ) -> AckableMessage:
+        """
+        Wrap a received Service Bus message as an AckableMessage.
+
+        Captures ``receiver`` in the ack closure so completion and lock
+        renewal re-registration target queue message was received from.
```
danielbreves:
```suggestion
        renewal re-registration target the queue the message was received from.
```

app/core/azure_service_bus_broker.py
```diff
@@ -140,31 +156,59 @@ async def shutdown(self) -> None:
...
+            return TaskPriority.NORMAL
+
     async def kick(self, message: BrokerMessage) -> None:
         """
         Send message to the queue.
 
         This function constructs a service bus message and sends it with the
         appropriate metadata and routing.
 
+        Messages with TaskPriority.HIGH label are sent to the priority queue;
```
danielbreves:
```suggestion
        Messages with TaskPriority.HIGH label are sent to the priority queue.
```

app/core/telemetry/taskiq.py
```diff
@@ -39,6 +53,8 @@ async def queue_task_with_trace(
     :type args: object
     :param long_running: Whether the task is long-running and needs lock renewal.
     :type long_running: bool
+    :param priority: Priority of the class.
```
danielbreves:
```suggestion
    :param priority: Priority of the task.
```

app/core/azure_service_bus_broker.py
```diff
@@ -190,134 +235,151 @@ async def kick(self, message: BrokerMessage) -> None:
...
             raise MessageBrokerError(detail="Call startup before starting listening.")
 
         while True:
             try:
-                # Receive a batch of messages
                 async with self._receive_lock:
-                    batch_messages = await self.receiver.receive_messages(
-                        max_wait_time=10
+                    priority_batch = await self.priority_receiver.receive_messages(
+                        max_wait_time=settings.message_broker_queue_max_wait
```
danielbreves: Looks like the `max_wait_time` setting may be flipped?
```suggestion
                        max_wait_time=settings.message_broker_priority_queue_max_wait
```

app/core/azure_service_bus_broker.py
```diff
@@ -190,134 +235,151 @@ async def kick(self, message: BrokerMessage) -> None:
...
-                        else False
-                    ):
-                        data = gzip.decompress(data)
-
-                    ackable = AckableMessage(
-                        data=data,
-                        ack=ack_message,
+                async with self._receive_lock:
+                    batch_messages = await self.receiver.receive_messages(
+                        max_wait_time=settings.message_broker_priority_queue_max_wait
```
danielbreves:
```suggestion
                        max_wait_time=settings.message_broker_queue_max_wait
```

danielbreves (changes requested):

app/core/azure_service_bus_broker.py
jack-o-the-green: 🤦 🤦 🤦 

Good catch! Clearly wasn't awake this morning.

danielbreves (approved):

PR #692, opened 2026-05-19
Add option to pass document id subset to index repair

Adam-Hammo:
Allows the reindexing of a subset of document IDs

Expedited for handling of https://covidence.slack.com/archives/C0810ARPCRG/p1779150745575739

image


jaybea (changes requested):
The code is clear and the API shape feels right. A few blockers that should at least be considered before this lands (mainly trying to future proof it against the day that I need to use it and mess up my params), plus some smaller polish notes after that if any feel worth it.

## Blockers

**1. A single nonexistent UUID looks like it would silently fail the whole subset task.** It looks to me like `bulk_sql_to_es` (`app/domain/references/services/synchronizer_service.py:123`) calls `get_by_pks(reference_id_chunk, preload=…)` without overriding `fail_on_missing`. The default at `app/persistence/sql/repository.py:260` is `True`, so one typo or recently-deleted ID would raise `SQLNotFoundError` *after* the route has already returned 202. The operator's most likely workflow ("paste 200 IDs from a Slack thread, hit the endpoint") seems like exactly the failure-prone case, and the failure mode wouldn't be visible from the API. Maybe think about pre-validating the IDs at the route (422 on any missing), or alternatively threading `fail_on_missing=False` into the subset path and having the task log/report skipped IDs. Curious which feels right to you. A test for `[valid_id, missing_id]` could be reassuring either way.

**2. Subset repair looks like it would silently no-op on the IDs an operator is most likely to send.** As far as I can tell the two sync paths have asymmetric semantics:
- `sql_to_es` (single, `synchronizer_service.py:88-91`): non-canonical-like + has canonical → `delete_by_pk(reference.id, fail_hard=False)` then recurses into the canonical.
- `bulk_sql_to_es` (subset, `synchronizer_service.py:127-129`): `if reference.is_canonical_like: yield ...`, which silently skips non-canonical refs with no delete.

The docstring at line 100 sez ("This does not handle any deletions, purely upserting"). That feels fine for the corpus-wide repair (canonical-only by construction), but for the "I see a stale ES doc, repair it" workflow this PR documents, it looks to me like the most likely input is a stale ES doc ID, which is precisely the case that would get silently skipped. I wonder if routing this through the per-ID `sql_to_es` path (so it gets delete semantics) would feel right, or whether surfacing the upsert-only restriction in the route docstring and response message would be enough.

**3. `{"document_ids": null}` looks like it triggers a full reindex.** Empirical probe against the route:

```
{"document_ids": null}              -> 202, full reindex queued
{}                                  -> 202, full reindex queued
no body                             -> 202, full reindex queued
{"document_ids": []}                -> 422 (min_length=1 works as intended)
{"document_ids": [valid_uuid]}      -> 202, subset reindex (correct)
```

It looks like `Annotated[list[UUID], Body(...)] = None` accepts null and falls through to the no-subset branch on `app/system/routes.py:177`. On the prod corpus that could be hours of unintended work from a client templating bug. The OpenAPI schema also advertises the field as non-null `list[UUID]` while the runtime accepts null, so the contract and the behavior seem to disagree. Maybe think about either rejecting null explicitly when the key is present (422), or going the other way and committing to "null means no subset" with `list[UUID] | None` in the signature, with the `None` dropped from `examples=` on `routes.py:149`.

## Smaller polish notes

**4. Duplicated task body.** `repair_reference_index_subset` (`tasks.py:286-304`) is byte-for-byte identical to the tail half of `repair_reference_index_for_chunk` (`tasks.py:271-283`) once you strip the partition lookup; both end at `await reference_service.index_references(reference_ids)`. Maybe a shared `_repair_subset_inner(reference_ids)` helper could DRY them up, though happy to leave as-is.

**5. Error-type asymmetry for "subset unsupported."** Route raises `InvalidPayloadError` → 422 (`routes.py:163`); `IndexManager.repair_index(document_ids=…)` raises `NotFoundError` → 404 (`index_manager.py:171`) on the same condition. The IndexManager branch is unreachable from HTTP today, but it's a defensive backstop for direct callers, and `NotFoundError` reads more like a capability/configuration issue to me than a missing entity. Wondering if aligning them feels worth it.

**6. `max_length=settings.es_reference_repair_chunk_size` couples the request size limit to the worker chunk size.** They might want to be the same number today, but they feel conceptually different. Tuning worker chunking later could silently change the API contract. Maybe worth thinking about a separate `es_reference_repair_subset_max_ids` (defaulting to the chunk size).

**7. `examples=[[uuid7(), uuid7()], None]` (`routes.py:149-152`).** Wondering if dropping the `None` makes sense either way (see #3); the inline `uuid7()` calls also evaluate at import time, so the OpenAPI docs ship fixed UUIDs forever. Pretty minor; fixed string literals might be cleaner.

**8. Auth-gate test for the subset shape.** `test_repair_auth_failure` covers the rebuild path but not the new subset body. Could be worth sending a subset request without a token and asserting 401, just for cheap regression coverage.

#4-#8 don't feel like dealbreakers on their own. I think #1-#3 should seriously be considered.

jaybea:
Pretty cool, I tested some mojibake fixes locally (inspired by PR #699) with the ref ES re-index - Showed you can't find docs with mojibake in the title (cause you can't type the invisible last byte), then used pure SQL to fix it:

```sql
UPDATE enhancement
SET content = jsonb_set(
  content,
  '{title}',
  to_jsonb(convert_from(convert_to(content->>'title', 'LATIN1'), 'UTF8'))
)
WHERE id = '...';
``` 

then reindex just the changed refs and show that you CAN now find them with ES search!

Adam-Hammo:
Thanks @jaybea!

1. Good pick-up, definitely don't want to fail silently. I'll update to check at API call time and return 404 if there's a missing ID, like we do elsewhere. I think it's important to let the caller know the ID doesn't exist as it might imply their method of getting IDs to reindex was wonky.
2. My opinion is the actual syncing mechanism is a different concern to this PR. #693 addresses some of these deficiencies. For the first couple of passes, we'll be using this to update stale canonical references and reindex the ESEA community as LDE projections change, so I don't think it's an immediate concern.
3. This was by design to preserve existing functionality in a backwards-compatible way. I see your point that it could be accidentally omitted though, I propose we make it an explicit `repair_all` or similar flag as well. There's no existing programmatic users so we can break this contract.
4. ~Good call~ This abstraction actually ended up being a bit superfluous, it really just wraps `index_references()`
5. I'm not sure which way we would align this - one is explicitly an API-level exception, while the other is lower level. I think NotFoundError = 404 is perhaps a false equivalence, if a non-API process calls the index manager they can handle that in whichever way, and InvalidPayloadException wouldn't make sense.
6. I see this variable as "what's the largest number of references we want to reindex in one task", which would align their meaning in the two uses. Maybe a variable rename would address this?
7. Fair, will need changing with 3 anyway
8. I'm not sure I understand this, isn't auth applied at the route level?

jaybea:
Thanks @Adam-Hammo.

1. Agreed, 404 is better than 422 here. Detecting wonkiness in gathering ID's is what I'd want to see in response.
2. Sounds good, we'll defer to #693.
3. `repair_all' or similar seems like a great way to solve this.
5. You're right, they're serving different concernes.
6. I see your point, we can just keep this as-is.
8. Your right, existing auth test already exercises sufficiently.

jaybea:
Wow, fast turnaround!

I reran my mojibake ES search failure tests and targeted reindexing still works great. Also bad UUID's when present were surfaced, no accidental full reindexes when I retried those, batch size rename clarification is a nice touch.

Good to go I think!

jaybea (approved):
Good to go!

PR #694, opened 2026-05-20
Move entitlements into a core-level file

Adam-Hammo:
`Entitlements` lived at `api/auth.py`. When imported by the `Robot` SQL model (as we store entitlements on robots), this causes `Settings` to eagerly load.

When performing an alembic migration, the `Settings` fails loading as we don't provide the required configuration. The root of the problem though was the SQL models reaching into API auth concepts. This PR moves the entitlements into a separate `core.entitlements` file.


jack-o-the-green (approved):
LGTM

Issue #697, opened 2026-05-20
eppi_importer latin-1 default mojibakes UTF-8 EPPI exports


jaybea:
## Summary

`libs/eppi_importer/import_from_eppi.py` defaults `--input-codec` to
`latin-1` (since aade7fac, 2025-12-08). UTF-8 EPPI exports — including the
EEF 2025-12-03 export — are misdecoded byte-for-byte, producing silent
mojibake for every non-ASCII character encoded as a multi-byte UTF-8
sequence. The mojibake'd strings survive unchanged through the ingest
pipeline into the DB.

## How to reproduce

**Source bytes.** In `eef_export-2025-12-03.00002.json` at byte offset
1895095, the title's apostrophe is encoded as the UTF-8 sequence
`\xe2\x80\x99` (U+2019).

**Local DB.** The reference for DOI `10.1111/j.1468-0297.2008.02247.x`
(imported from the EEF 2025-12-03 export) stored title bytes `... 73 c3
a2 c2 80 c2 99 20 74 72 ...` — `Teachers` + U+00E2, U+0080, U+0099 +
` tr`. That's the latin-1 reinterpretation of the source UTF-8 bytes,
with each input byte becoming a separate codepoint, then re-encoded to
UTF-8 for storage. The C1 control codepoints (U+0080, U+0099) often
render invisibly, which is why this passed UI review.

**Prod confirmation (2026-05-20).** Reference
`68db42c5-e353-4572-99e3-95dc99a04cd0` (same DOI) carries the identical
mojibake byte signature `c3 a2 c2 80 c2 99`. A parallel OpenAlex-sourced
reference for the same paper is encoded correctly, isolating the issue to
the EPPI ingestion path.

## Root cause

Commit aade7fac ("Decode files with latin-1 encoding", 2025-12-08) flipped
the default from `utf-8` to `latin-1` to silence a decode error on the EEF
export. `latin-1`'s 256→256 byte-to-codepoint mapping is total, so it can
never raise — but for actually-UTF-8 input, every multi-byte sequence is
misdecoded into its constituent bytes as separate codepoints.

The underlying decode error being worked around is a real but distinct
issue: the EEF export contains CESU-8 surrogate pairs (e.g.
`\xed\xa0\xb5\xed\xb1\x9b` for U+1D45B 𝑛, MATHEMATICAL ITALIC SMALL N) at
offset 5045365, which strict UTF-8 correctly rejects. CESU-8 represents
[...]

PR #699, opened 2026-05-20
Restore utf-8 default in eppi_importer to fix mojibake on imports

jaybea:
## Summary

- Restores `--input-codec` default to `utf-8` (was flipped to `latin-1`
  in aade7fac, 2025-12-08).
- Adds `errors="replace"` so the CESU-8 surrogate pairs in EPPI exports
  degrade to U+FFFD rather than failing the import.
- Adds a regression test covering the U+2019 apostrophe byte pattern
  from the EEF 2025-12-03 export (offset 1895095).

See #697 for the full bug analysis and the lossy-vs-lossless codec
tradeoff. This PR ships the minimum viable fix that closes the immediate
regression; the broader CESU-8 fidelity question (custom codec vs
surrogatepass vs ftfy) stays open in the Issue.

Existing mojibake'd records in prod are bijectively repairable (see
the Issue's "Repair note"); a one-shot fix script for affected rows is
a follow-up, not part of this PR.

## Validation performed

- [x] `uv run pytest libs/eppi_importer/tests/test_import_from_eppi.py`
      passed against the fix.
- [x] `uv run pyright libs/eppi_importer/import_from_eppi.py
      libs/eppi_importer/tests/test_import_from_eppi.py` — 0 errors.
- [x] `uv run pre-commit run --files <changed files>` clean.
- [x] Ran importer on the actual EEF 2025-12-03 export and confirmed
      U+2019 in the title round-trips as UTF-8 bytes `\xe2\x80\x99`,
      not the byte-doubled `\xc3\xa2\xc2\x80\xc2\x99`.

Addresses the regression portion of #697; broader codec strategy stays
open there.

### DIFF
--- a/libs/eppi_importer/import_from_eppi.py
+++ b/libs/eppi_importer/import_from_eppi.py
@@ -38,8 +38,8 @@ def main() -> None:
         "--input-codec",
         "-ic",
         type=str,
-        default="latin-1",
-        help="The codec to decode the input file with, defaults to 'latin-1'",
+        default="utf-8",
+        help="The codec to decode the input file with, defaults to 'utf-8'",
     )
 
     arg_parser.add_argument(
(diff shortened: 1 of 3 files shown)


Adam-Hammo:
If we wanted to, would we able to do the same thing at the vocab robot to create clean LDEs from mojibaked raw enhancements?

jaybea:
Yes, I believe so. We could bump the robot version, update it to detect mojibake in the raw, (look for those invisible C1 bytes), and if found correct it (it's reversible). Then queue all the ESEA refs and re-process.

Adam-Hammo (approved):

PR #705, opened 2026-05-22
Naive facet count

### LINKED ISSUE #698: Add naive search facet count API endpoint
Flow-on from https://github.com/destiny-evidence/evidence-repo-ui/issues/25.

⚠️ This is an intentionally naive implementation: facet counts reflect the full filter set (including the facet's own selections), so unselected siblings of selected facets will return co-occurrence counts (typically ~0) rather than alternative counts. See #703 for correct semantics. The endpoint must surface this clearly in its OpenAPI description.

Add an API endpoint to display per-facet counts of `linked_data_concepts` of a particular search.

This is a naive implementation that does not correctly handle OR sibling relationships. See #703 for follow-up work to handle this. The endpoint's documentation should clearly state this limitation.

This is not a generalised aggregation endpoint (eg crosstabbing). My weak opinion is that these sort of metrics/visualisations would not live under the `/search/` resource of the API, whereas these counts would.

# Proposal

## API interface

Add `/search/facets/` endpoint that supersets the `search` endpoint's parameters, adding a `?facet=concepts` parameter also (later can support countries, years).

## Concept Counting (naive)

Perform a simple term aggregation on linked_data_concepts to count the concepts:

```
GET /reference/_search?size=0
{
    "query": {
        "query_string": "..."
    },
    "size": 0,
    "aggs": {
      "concepts": {
        "terms": {
           "size": 1000
        }
      }
    }
}
```

## Data Structure

The `facets` model will look like:

```
{
  facets: {
    concepts: [
      {uri: ..., count: ...},
      ...
    ],
    countries: [
      {country: ..., count: ...},
      ...
    ],
    publication_years: [
      {start_year: ..., end_year: ..., count: ...},
      ...
    ]
  }
}
```

Only concepts will be done in this ticket - the rest are included by way of example.

## Considerations
- The vocabulary will need to be provided alongside the search, as it requires traversal. This should perhaps be folded into #695 while we're changing the API?
  - We would need to do similar if we wanted to do child term explosion in the future.
- Any concept filters included in the query string instead of the API parameters will incorrectly count. We could detect and reject these queries if `count_facets=True`

Adam-Hammo:
Closes #698 

- New endpoint to perform naive concept counting
- Related SDK models
- Shared query parameters for search-related endpoints

image

### DIFF
--- a/app/domain/references/routes.py
+++ b/app/domain/references/routes.py
@@ -362,20 +362,16 @@ def parse_search_query(
     )
 
 
-def parse_sort(
-    sort: Annotated[
-        list[str] | None,
-        Query(
-            description="A list of fields to sort the results by. "
-            "Prefix a field with `-` to sort in descending order. "
-            "If omitted, will sort by relevance score descending. "
-            "Multiple sort fields can be provided and will be applied "
-            "in the order given. Sort fields cannot be `text` fields.",
-        ),
-    ] = None,
-) -> list[str] | None:
-    """Parse the sort query parameter."""
-    return sort
+SortParam = Annotated[
+    list[str] | None,
+    Query(
+        description="A list of fields to sort the results by. "
+        "Prefix a field with `-` to sort in descending order. "
+        "If omitted, will sort by relevance score descending. "
+        "Multiple sort fields can be provided and will be applied "
+        "in the order given. Sort fields cannot be `text` fields.",
+    ),
+]
 
 
 @search_router.get(
@@ -405,7 +401,7 @@ async def search_references(
         ReferenceAccessControlService, Depends(reference_reader_access_control_service)
     ],
     query: Annotated[SearchQuery, Depends(parse_search_query)],
-    sort: Annotated[list[str] | None, Depends(parse_sort)],
+    sort: SortParam = None,
     page: Annotated[
         int,
         Query(
@@ -501,7 +497,7 @@ async def request_search_export(
     ],
     entitlements: Annotated[frozenset[Entitlement], Depends(reference_reader_auth)],
     query: Annotated[SearchQuery, Depends(parse_search_query)],
-    sort: Annotated[list[str] | None, Depends(parse_sort)],
+    sort: SortParam = None,
 ) -> destiny_sdk.references.SearchExportRead:
     """Queue a search export job and return its id and pending status."""
     search_export = await search_export_service.request_search_export(
--- a/app/domain/references/routes.py
+++ b/app/domain/references/routes.py
@@ -54,6 +54,7 @@ from app.domain.references.models.models import (
     PendingEnhancementStatus,
     PublicationYearRange,
     ReferenceIds,
+    SearchQuery,
 )
 from app.domain.references.service import ReferenceService
 from app.domain.references.services.access_control_service import (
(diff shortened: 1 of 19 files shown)


app/domain/references/routes.py
```diff
@@ -365,20 +404,8 @@ async def search_references(
...
-    ],
-    annotations: Annotated[
-        list[AnnotationFilter] | None,
-        Depends(parse_annotation_filters),
-    ],
-    publication_year_range: Annotated[
-        PublicationYearRange | None,
-        Depends(parse_publication_year_range),
-    ],
+    query: Annotated[SearchQuery, Depends(parse_search_query)],
```
danielbreves, line 403: Nice!

libs/sdk/src/destiny_sdk/references.py
```diff
@@ -78,6 +78,30 @@ class ReferenceSearchResult(SearchResultMixIn, BaseModel):
...
+    concept: str = Field(description="The concept URI.")
+    count: int = Field(
+        description="Number of matching references tagged with this concept.",
+    )
+
+
+class ReferenceFacetResult(BaseModel):
+    """Facet counts for a reference search."""
+
+    facets: dict[FacetType, list[ConceptFacetCount]] = Field(
```
danielbreves: Is this likely to support other facets? Or should we just make it: `concepts: list[ConceptFacetCount] | None = None` and have new facets as new fields?

app/domain/references/services/search_service.py
```diff
@@ -82,35 +87,67 @@ def _build_annotation_query_string_filter(
...
-        self,
-        query_string: str,
-        page: int = 1,
-        page_size: int = 20,
-        annotations: list[AnnotationFilter] | None = None,
-        publication_year_range: PublicationYearRange | None = None,
-        sort: list[str] | None = None,
-    ) -> ESSearchResult:
-        """Search for references matching the query string."""
+    _FACET_FIELDS: ClassVar[dict[FacetType, str]] = {
```
danielbreves: Nit: could we move this to the top of the class?

app/domain/references/routes.py
```diff
@@ -339,6 +340,44 @@ def parse_annotation_filters(
...
+) -> SearchQuery:
+    """Bundle the shared search parameters into a domain SearchQuery."""
+    return SearchQuery(
+        query_string=q,
+        annotation_filters=annotation_filters,
+        publication_year_range=publication_year_range,
+    )
+
+
+def parse_sort(
```
danielbreves: Nit: could this be a simple type alias? i.e. `SortParam = Annotated[list[str] | None, Query(description=...)]`

danielbreves (approved):
Some minor comments. Looks good otherwise!

libs/sdk/src/destiny_sdk/references.py
Adam-Hammo: My thinking was countries, which aren't keyed by URIs, and perhaps also publication year buckets, which would be keyed by a tuple of start and end year. I'm making early assumptions about decisions we're not making yet though so feel free to push back!

Adam-Hammo: Ah I think I also see what you're saying on a higher level now too - I felt a bit of tension in the return typing. It felt weird being able to specify `facets=concepts` but having a response model that also has optional `countries`, `publication_years`, etc. 

This is pretty nitty-gritty so happy to go back to a named base-level class like you suggest?

Adam-Hammo: Went ahead with your suggestion. The behaviour should be clear enough to the caller!

Issue #637, opened 2026-04-13
Consider merging multiple LinkedDataEnhancements in projection


danielbreves:
## Context

The current `ReferenceSearchFieldsProjection` uses winner-takes-all for linked data enhancements — the highest-priority LDE overwrites all others. This differs from annotations, which accumulate by scheme.

Each `LinkedDataEnhancement` is scoped by its `vocabulary_uri`, which is a natural grouping key (analogous to annotation schemes). If a reference has LDEs from multiple vocabularies, only the highest-priority one is projected today.

## Proposal

Accumulate LinkedDataEnhancements by taking each RawEnhancement (as RawEnhancements have different sources), and then preserving the most recent LinkedDataEnhancements for vocabularies used for that reference (last-per-vocabulary wins).

For example

* If we have two RawEnhancements, one from EEF and one from ESSA, each with an associated LinkedDataEnhancement for the ESEA vocabulary, we preserve both LinkedDataEnhancements regardless of vocabulary version (as they are built from different sources and may contain different data/disagreements). 
* If we have one RawEnhancement, and two associated LinkedDataEnhancements for _two different_ vocabularies, we preserve both LinkedDataEnhancements as they are for different vocabularies. 
* If we have one RawEnhancement, and two associated LinkedDataEnhancements for the _same_ vocabulary, we take the LinkedDataEnhancement with the latest vocabulary version. 
    * We need to discuss whether we should prioritise the most recently created, or check for the most recent vocabulary version. The former allows us to roll back to a previous version by creating a new LinkedDataEnhancement with it, the latter does not. We may want to allow a `deprecated` or similar field that removes enhancements from projection. 

We then project each LinkedDataEnhancement independently via `LinkedDataProjectionService`, and union the flat outputs:

- `linked_data_concepts` — union of concept URIs (already globally scoped by URI, no conflict)
- `linked_data_evaluated_properties` — union of property URIs (same)
- `linked_data_labels` — deduplicated union (text search surface, not identifiers — duplicates across vocabularies are harmless but unnecessary)

## Additionally

Remove the `sorted()` calls on the projected linked data fields. Sorting provides no search benefit — ES treats `keyword` arrays as unordered sets, and `text` arrays are analysed independently per element. The sort was added for human readability of `_source` but adds unnecessary computation.

Adam-Hammo:
Should we consider stripping out the version from `vocabulary_uri` for grouping, so we can supersede vocabulary versions without preserving old data?

[...]

PR #708, opened 2026-05-25
Project country codes from linked data enhancements (#696)

### LINKED ISSUE #696: Make reference country searchable in destiny-repository
## Context 

Searching by country is high need for our stakeholders who want to use faceted search. 

As we are only projecting taxonomy concept schemes into elasticsearch, we are not projecting country as a searchable field, so it is not possible to construct a search query that targets country. 

## Technical Scope 

* Add a country mapping to our elasticsearch index that can take multiple values and is populated by our linked data enhancement projection
    * This will be the same country codes that the taxonomy currently uses. We should consider whether the country part of the taxonomy eventually comes into core, otherwise searching could be more tricky. 
    * It may make sense to project both the coded value and the supporting text.  
* Do an elasticsearch migration to apply the mapping 
* Refresh/reindex of all documents included in the jacobs domain to populate the search field on the index
* Update our search documentation (if necessary) to show how to search by country
* Add some tests to confirm searching by country is possible.

## Notes
The following elasticsearch query will get you a document in prod that has a linked data enhancement.
```
Get reference/_doc/019db8c6-245a-76ca-85c4-4717d59b2b18
```

The linked data enhancement contains the country

```
  "country": [
    {
      "@type": "StringCodingAnnotation",
      "status": "evrepo:coded",
      "codedValue": {
        "@type": "xsd:string",
        "@value": "KE"
      },
      "supportingText": "Kenya"
    }
  ],
```

Which is displayed in the evidence repo because it parses the full investigation for display 
https://data.evidence-repository.org/esea/references/019db8c6-245a-76ca-85c4-4717d59b2b18

danielbreves:
Closes #696.

## Summary

- Adds `LinkedDataProjection.countries` and a new `linked_data_countries` Keyword field on the reference index, projected from `esea:country` StringCodingAnnotations.
- Registers country properties in a `_COUNTRY_PROPERTIES` set so future vocabularies opt in by URI.
- ISO codes normalised to uppercase on write; `notReported` / `notApplicable` country annotations excluded.

## Test plan

- [x] `uv run pytest tests/unit/domain/references/services/test_linked_data_projection_service.py` — 13 pass, including 4 new country tests (extraction, uppercase normalisation, `notReported` exclusion, and a regression that other StringCodingAnnotations like `participants` are not projected).
- [x] `uv run pytest tests/unit/persistence/es/test_repository.py::test_linked_data_field_search` against local ES — 11 pass, including 3 new country search cases (`KE`, `GH`, negative `ZZ`).
- [ ] After merge: run the ES mapping migration and reproject the Jacobs-domain corpus to populate `linked_data_countries` on existing documents.

### DIFF
--- a/app/domain/references/models/es.py
+++ b/app/domain/references/models/es.py
@@ -356,18 +356,18 @@ class ReferenceSearchFieldsMixin(InnerDoc):
             annotations=search_fields.annotations,
             evaluated_schemes=search_fields.evaluated_schemes,
             inclusion_destiny=search_fields.destiny_inclusion_score,
-            linked_data_concepts=sorted(linked_data_projection.concepts)
+            linked_data_concepts=list(linked_data_projection.concepts)
             if linked_data_projection
             else None,
-            linked_data_labels=sorted(linked_data_projection.labels)
+            linked_data_labels=list(linked_data_projection.labels)
             if linked_data_projection
             else None,
-            linked_data_evaluated_properties=sorted(
+            linked_data_evaluated_properties=list(
                 linked_data_projection.evaluated_properties
             )
             if linked_data_projection
             else None,
-            linked_data_countries=sorted(linked_data_projection.countries)
+            linked_data_countries=list(linked_data_projection.countries)
             if linked_data_projection
             else None,
         )
--- a/app/domain/references/models/es.py
+++ b/app/domain/references/models/es.py
@@ -334,6 +334,12 @@ class ReferenceSearchFieldsMixin(InnerDoc):
     )
     """Property URIs for all evaluated linked data dimensions."""
 
+    linked_data_countries: list[str] | None = mapped_field(
+        Keyword(required=False),
+        default=None,
+    )
+    """ISO country codes from LinkedDataEnhancements, for exact-match filtering."""
+
     @classmethod
     def from_projections(
         cls,
@@ -359,6 +365,9 @@ class ReferenceSearchFieldsMixin(InnerDoc):
             linked_data_evaluated_properties=sorted(
                 linked_data_projection.evaluated_properties
             )
+            if linked_data_projection
+            else None,
+            linked_data_countries=sorted(linked_data_projection.countries)
             if linked_data_projection
             else None,
         )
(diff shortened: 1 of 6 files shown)


jack-o-the-green (approved):
Looks good to me! 

Interested in whether you think it would be valuable to do follow up work to add a country parameter to search that handles conversion of country names to iso codes?  

We can handle that on the UI for now, but I think it would be nicer for users to see a `&country="Kenya"` in their search rather than `q=phonics AND linked_data_countries:KE`

app/domain/references/models/es.py
```diff
@@ -361,6 +367,9 @@ def from_projections(
             )
             if linked_data_projection
             else None,
+            linked_data_countries=sorted(linked_data_projection.countries)
```
jack-o-the-green: We talked about taking the `sorted(` calls out of here in #637, can we add this change in now since we're touching the file?

jaybea (approved):
This looks great!

A couple of inane nits, feel free to ignore!

nit: Docstring on `project()` is now stale - also extracts countries. Something like:

```python
"""Extract concepts, labels, evaluated properties, and countries."""
```

nit: The PR description mentions excluding both `notReported` and `notApplicable`, but only `notReported` is asserted. The code path (`status not in (EVREPO.coded, None)`) covers both - worth a parametrize so the contract is explicit:

```python
@pytest.mark.parametrize(
    "status",
    ["evrepo:notReported", "evrepo:notApplicable"],
)
async def test_excludes_uncoded_country(self, projector, status):
    ...
```

danielbreves:
> Interested in whether you think it would be valuable to do follow up work to add a country parameter to search that handles conversion of country names to iso codes?

> We can handle that on the UI for now, but I think it would be nicer for users to see a &country="Kenya" in their search rather than q=phonics AND linked_data_countries:KE

Yea, I think that makes sense given this is powering an end user interface and not purely API users. We could use https://github.com/IndEcol/country_converter.

jack-o-the-green:
> Yea, I think that makes sense given this is powering an end user interface and not purely a API users. We could use https://github.com/IndEcol/country_converter.

Noice, I will make a follow on ticket to put in the bucket of "one day when we have time"

PR #709, opened 2026-05-25
Project World Bank regions from country codes (#707)

### LINKED ISSUE #707: Group countries into World Bank regions for filter selection
Parent: destiny-evidence/evidence-repo-ui#78

## Context

The country filter currently exposes individual ISO 3166-1 alpha-2 codes (see #696 for the backend country projection). Stakeholders want to filter by World Bank region (e.g. "Sub-Saharan Africa") in addition to selecting individual countries.

## Suggested approach

Project region as a sibling Keyword field on the reference index so the UI filter is a flat facet, mirroring how country itself was added:

- Static ISO → WB region lookup applied inside `LinkedDataProjectionService` alongside country extraction.
- Emitted as a new `linked_data_country_regions` Keyword field on `ReferenceSearchFieldsMixin`.
- One ES mapping migration + reproject — piggybacks the migration already pending for country.
- UI (destiny-evidence/evidence-repo-ui) renders a region selector in the country filter that queries `linked_data_country_regions:"Sub-Saharan Africa"`.

Region list: the seven World Bank regions (https://datahelpdesk.worldbank.org/knowledgebase/articles/906519).

danielbreves:
Closes #707. Stacks on top of #708.

## Summary

- Adds `LinkedDataProjection.country_wb_regions` and a new `linked_data_country_wb_regions` Keyword field, derived from `linked_data_countries` via a static ISO → WB region table (`app/domain/references/services/world_bank_regions.py`).
- Stores WB region IDs (e.g. `SSF`).
- `app/utils/refresh_world_bank_regions.py` rewrites the static dict in place from the live WB API; `git diff` is the drift report.

## Test plan

- [x] `uv run pytest tests/unit/domain/references/services/test_linked_data_projection_service.py` — 16 pass, including 3 new region tests (dedupe across multi-country findings, distinct regions across continents, unknown ISO omitted).
- [x] `uv run pytest tests/unit/persistence/es/test_repository.py::test_linked_data_field_search` — 13 pass, including 2 new region search cases (`SSF` matches, `SAS` doesn't).
- [x] `uv run python -m app.utils.refresh_world_bank_regions` — script runs against live WB API and produces a byte-identical file (idempotent when in sync).
- [ ] After merge of #708 and this: run the ES mapping migration and Jacobs-domain reproject to populate both new fields on existing documents.


jack-o-the-green (approved):
LGTM - nice quick solution! Same comment as on #708 for removing unnecessary `sorted()` from projection.

Issue #577, opened 2026-03-02
Add ObservedResult and EffectEstimate parsing from outcome fields.


jack-o-the-green:
### Background
Observed results and EffectEstimates are stored in the Outcomes list in each references RawEnhancement. Sometimes they are stored in the same outcome, and other times not. 

To determine whether we are looking at an ObservedResult or an EffectEstimate, we need to work backwards from the `Data[INT]Desc` fields. We can extract mean, standard deviation, and sample size from here to create ObservedResults. There are many other possible values that constitute different types of EffectEstimates. EffectEstimates can also be stored in specific fields on the Outcome, typically when the `Data[INT]` fields contain ObservedResults. 

It is also fine for an EffectEstimate to be logged without an ObservedResult. 

### Technical Scope 

* Extract ObservedResults by parsing outcomes
   * You'll need to work out which `Data[INT]Desc` possibilities map to which ObservedResults properties in evrepo-core.ttl
* Extract EffectEstimates by parsing outcomes
   * You'll need to work out which `Data[INT]Desc` possibilities map to which EffectEstimate properties in evrepo-core.ttl _and_ how they can be used to identify the `effectSizeMetric` which are properties that contain defined concepts in our lightweight ontology. 
   * The EEF data does not appear to code for `estimateSource` so we'll leave this out. 

### What this ticket doesn't cover 

Don't worry about turning them into jsonld for the time being, the majority of that work is being done in #576 

### Notes

Useful references (from prod)
* `b7a34d6d-f11e-4852-9030-d060a6bfdf8f` (nice one to get started)
*  `7d6f1092-1976-4903-ab17-330b79787f7d` (p-value and t-value in data descriptions)

https://covidence.atlassian.net/wiki/spaces/Platforms/pages/1055326209/Mapping+EEF+data+to+ESEA+Vocabulary#ObservedResult-and-EffectEstimate

Example jsonld reference https://github.com/destiny-evidence/ad-hoc-mappers/pull/1/changes#r2875889427

PR #711, opened 2026-05-25
Fix linked_data_concepts missing EffectEstimate concept properties

jaybea:
## Summary

The linked-data projection walked only `evrepo:codedValue` triples, so concept references that appear bare on `EffectEstimate` (`effectSizeMetric`, `estimateSource`) never reached the `linked_data_concepts` or `linked_data_evaluated_properties` ES fields. The EffectSizeMetric scheme in the UI FilterDrawer rendered empty as a result.

This adds a second projection pass that picks up unwrapped concept references (concept values that appear at a property's slot without a `CodingAnnotation` envelope around them). The set of properties is discovered from the vocabulary itself via SPARQL (`owl:ObjectProperty` whose `rdfs:range` is, or resolves to, a `skos:ConceptScheme`), rather than encoded as a hardcoded registry. Currently this surfaces `effectSizeMetric` and `estimateSource`; any future unwrapped concept property added to the vocab will project automatically with no DR code change.

## Implementation notes

- The two traversals stay separate. `codedValue` properties carry status/provenance and need scheme-to-property indirection via the OWL restriction graph; unwrapped properties point straight at concepts and the predicate URI IS the scheme property. The asymmetry mirrors a deliberate vocab decision (taxonomy-builder #108): EffectEstimate carries class-level provenance, so its concept refs are unwrapped rather than CodingAnnotation-wrapped.
- The discovery SPARQL uses a UNION to handle two equivalent vocab patterns. Pattern A (production today): `rdfs:range` IS a `skos:ConceptScheme` directly. Pattern B (conventional SKOS, used in the test fixture): `rdfs:range` is a class `subClassOf skos:Concept`, with concept instances joined to the scheme via `skos:inScheme`. Both shapes are legitimate; the query returns the same `(property, scheme)` rows from either.
- Cached as a new `unwrapped_concept_properties: dict[str, str]` field on `_LoadedVocabulary` alongside `scheme_to_property`, populated once per vocab URI on first access (same pattern as the existing lookups).
- `estimateSource` is unpopulated in current EEF-derived LDEs by design (vocabulary-mapping-robot omits it per its issue #577), which is why one of the new tests is synthetic rather than fixture-driven.

## Validation performed

- All 13 projection unit tests pass; 4 new regression tests cover: discovery of concept-typed properties against the deployed-shape vocab fixture, discovery against a synthetic graph using the direct-`ConceptScheme` range pattern, `effectSizeMetric` projection against the existing real-data LDE fixture, and both predicates together via synthetic JSON-LD.
- Ruff and ruff format clean on the touched files.
- evidence-repo-ui FilterDrawer renders the Effect Size Metric scheme with vocab-resolved labels; with `q=music`, selecting 7 metrics narrows results from 64 to 46, confirming the filter parameter wires through ES end-to-end. Per-concept counts are not yet implemented in the UI, so verification used result-count delta rather than facet aggregations.

## Backfill

Existing LDEs need a reindex to populate `linked_data_concepts`. This can ride along with the already-scheduled country/regions reindex.

### DIFF
--- a/app/domain/references/services/linked_data_projection_service.py
+++ b/app/domain/references/services/linked_data_projection_service.py
@@ -28,7 +28,7 @@ class _LoadedVocabulary:
     concept_labels: dict[str, str]
     concept_schemes: dict[str, str]
     scheme_to_property: dict[str, str]
-    unwrapped_concept_properties: dict[str, str]
+    unwrapped_concept_properties: set[str]
 
 
 class LinkedDataProjectionService:
@@ -90,7 +90,7 @@ class LinkedDataProjectionService:
                 if prop_uri is not None:
                     evaluated_properties.add(prop_uri)
 
-        _project_unwrapped_concept_properties(
+        self._project_unwrapped_concept_properties(
             data_graph, vocab, concepts, labels, evaluated_properties
         )
 
@@ -119,6 +119,34 @@ class LinkedDataProjectionService:
                         if code:
                             countries.add(code)
         return countries
+
+    @staticmethod
+    def _project_unwrapped_concept_properties(
+        data_graph: Graph,
+        vocab: _LoadedVocabulary,
+        concepts: set[str],
+        labels: set[str],
+        evaluated_properties: set[str],
+    ) -> None:
+        """
+        Project values discovered by ``_build_unwrapped_concept_properties``.
+
+        Without a CodingAnnotation wrapper there is no provenance, so any present
+        value is treated as coded.
+        """
+        for prop_uri_str in vocab.unwrapped_concept_properties:
+            predicate = URIRef(prop_uri_str)
+            for _, _, value in data_graph.triples((None, predicate, None)):
+                if not isinstance(value, URIRef):
+                    continue
+                if (value, RDF.type, SKOS.Concept) not in vocab.graph:
+                    continue
+                concept_uri = str(value)
+                concepts.add(concept_uri)
+                label = vocab.concept_labels.get(concept_uri)
+                if label is not None:
+                    labels.add(label)
+                evaluated_properties.add(prop_uri_str)
 
     # -- vocabulary resolution with derived-lookup caching ---------------------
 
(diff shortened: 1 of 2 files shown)


app/domain/references/services/linked_data_projection_service.py
```diff
@@ -157,3 +170,32 @@ def _build_scheme_to_property(graph: Graph) -> dict[str, str]:
...
+) -> None:
+    """
+    Project EffectEstimate concept properties that point at SKOS concepts directly.
+
+    These properties (``effectSizeMetric``, ``estimateSource``) carry a bare
+    ``@id`` rather than a CodingAnnotation wrapper, so they don't produce
+    ``codedValue`` triples and are missed by the main traversal. Any present
+    value is treated as coded, since these references have no provenance fields.
+    """
+    for predicate in _EFFECT_ESTIMATE_CONCEPT_PROPERTIES:
```
Adam-Hammo: Is there a reason not to project every concept of this shape? I might be missing a particular case

app/domain/references/services/linked_data_projection_service.py
```diff
@@ -21,6 +21,7 @@ class _LoadedVocabulary:
     concept_labels: dict[str, str]
     concept_schemes: dict[str, str]
     scheme_to_property: dict[str, str]
+    unwrapped_concept_properties: dict[str, str]
```
Adam-Hammo: Could this just be a set?

app/domain/references/services/linked_data_projection_service.py
```diff
@@ -157,3 +165,63 @@ def _build_scheme_to_property(graph: Graph) -> dict[str, str]:
...
+        return {str(row.prop): str(row.scheme) for row in results}
+
+
+def _project_unwrapped_concept_properties(
+    data_graph: Graph,
+    vocab: _LoadedVocabulary,
+    concepts: set[str],
+    labels: set[str],
+    evaluated_properties: set[str],
+) -> None:
```
Adam-Hammo: It feels a little strange to have this outside the class whereas `_build_unwrapped_concept_properties` is a staticmethod on the class - unsure which direction to take that.

Adam-Hammo (commented):
Looking good - noting that there might be a fun conflict with #714 which moves some of the vocab preprocessing but we'll cross that bridge when it comes to it!

app/domain/references/services/linked_data_projection_service.py
jaybea: You're right, let's stay consistent.

app/domain/references/services/linked_data_projection_service.py
jaybea: Good catch, yes!

Adam-Hammo (approved):
Looks good!

Issue #703, opened 2026-05-21
Handle sibling relationships in search facet counting


Adam-Hammo:
Follow-on from #698.

#698 introduced naive search facet counts, and #695 introduced linked data concepts as explicit filters.

Search facet counting does not yet handle sibling relationships, where applying the filter is an OR relationship, not an AND relationship. Selecting a sibling filter increases the number of results, selecting an unrelated filter decreases the number of results.

We need to implement the below.

### Worked example

**Taxonomy**
```
Scheme: Topics (hierarchical)
  Biology
    Botany
    Zoology
    Microbiology
  Chemistry

Scheme: Region (flat)
  Africa
  Asia
  Europe
```

**Query**
`GET /search/?q=<text>&concept=Botany,Zoology&concept=Africa&facets=concepts`

**Process**
- Traverse the vocabulary (already cached in the repository as a `rdflib` graph) to find the siblings of any selected concepts. We find two concept groups: 
  - `Topics->Biology`, with `Botany` and `Zoology` selected and `Microbiology` needing to be counted
  - `Region`, with `Africa` selected and `Asia` and `Europe` needing to be counted
- For each concept group, construct a "search broadening" aggregation that:
  - Applies any selected concepts from outside that group
  - Counts the concepts within that group
- Construct another aggregation that counts all of the other "search narrowing" concepts
- Finally, apply the concepts filter to the results as a `post_filter`. This filters the hits and not the aggregation (as we did our own filtering in the aggregations).

**Final constructed query**
```json
{
  "query": { "query_string": { "query": "<text and non-concept filters>" } },
  "post_filter": {
    "bool": { "must": [
      { "terms": { "linked_data_concepts": ["Botany", "Zoology"] } },
      { "terms": { "linked_data_concepts": ["Africa"] } }
    ]}
  },
  "aggs": {
    "facet_group_0": {
      "filter": { "terms": { "linked_data_concepts": ["Africa"] } },
      "aggs": { "concepts": {
        "terms": {
          "field": "linked_data_concepts",
          "include": ["Botany", "Zoology", "Microbiology"],
          "size": 1000
        }
      }}
    },
    "facet_group_1": {
      "filter": { "terms": { "linked_data_concepts": ["Botany", "Zoology"] } },
      "aggs": { "concepts": {
        "terms": {
          "field": "linked_data_concepts",
          "include": ["Africa", "Asia", "Europe"],
          "size": 1000
        }
      }}
    },
    "unselected": {
[...]

PR #713, opened 2026-05-26
Add linked concepts to search

### LINKED ISSUE #695: Add backend linked_data_concepts structured field parameter to search
## Context

Right now to filter search results by linked_data_concept, we need to construct large queries in the `q` field of the `/references/search` endpoint of destiny-repository. For our evidence-repo-ui, this results in query splitting to redetermine active filters, which while possible, is messy and more difficult especially as the complexity of our fields grow. 

## Technical Scope

* Add a structured search parameter `concept` similar to `AnnotationFilter` for filtering search results by `linked_data_concepts` https://github.com/destiny-evidence/destiny-repository/blob/main/app/domain/references/routes.py#L374. Referred to as `LinkedDataConceptFilter` after this. 
   * LinkedDataConceptFilter will need to allow OR matches between concepts contained in the same filter e.g. `concept=https://vocab.esea.education/EducationLevelScheme/C00003 OR https://vocab.esea.education/EducationLevelScheme/C00004` 
   * LinkedDataConceptFilter should then be AND matches between so `concept=A OR B&concept=C`
   * Should accept fully qualified URIs 
* Search documentation updated 

## Worth considering

- Should we add a `vocabulary_uri` parameter as well that can expand concept URIs from their compacted form (eg `esea:EducationLevelScheme/C00003`) using the `context.json`, creating shorter URLs? Optional, but required if concept URIs come in the compacted form. If we do this, I'd suggest doing it all at once, to avoid churn in the API contract.
  - Worth noting we need similar for #703 

## Out of scope 
Not doing label search for the time being, so no search by "Journal Article" etc

Adam-Hammo:
Closes #695.

Adds `linked_data_concepts` to all `search`-related endpoints. Each parameter instance is `AND-ed`, comma-separated concepts in a single instance are `OR-ed` (see `search.rst` for examples).

This also moves the string-building of the filters, in favour of adding them as structural . Necessarily, some of `search_service` has moved to a `repository` now. Integration tests in `test_search.py` provide a robust surface to verify the refactor. String building was brittle and needing to be removed for #703 anyway, so it felt right to do it before adding more.


app/domain/references/models/models.py
```diff
@@ -1116,6 +1136,13 @@ class SearchQuery(BaseModel):
         default=None,
         description="Publication year range to AND with the query string.",
     )
+    linked_data_concept_filters: list[LinkedDataConceptFilter] = Field(
```
jack-o-the-green, line 1121: Thinking we might want to move the overlap of the SearchQuery and SearchExport models to a shared base model - I don't believe there's an instance where we'd want the supported search fields to differ.

app/domain/references/repository.py
```diff
@@ -253,6 +270,109 @@ def __init__(self, client: AsyncElasticsearch) -> None:
             ReferenceDocument,
         )
 
+    def _build_publication_year_clause(
```
jack-o-the-green, line 273: I like the translation of the domain SearchQuery model into clauses living here 👌

app/domain/references/routes.py
```diff
@@ -340,6 +342,46 @@ def parse_annotation_filters(
         raise ParseError(detail=str(exc)) from exc
 
 
+def parse_linked_data_concept_filters(
```
jack-o-the-green, line 345: I'm getting cagey about the length of the routes file, these feel like appropriate candidates to move somewhere else eventually. We might need to go `routers/search.py` etc.

Not for this PR, just as a callout.

tests/factories.py
```diff
@@ -415,9 +416,24 @@ def enhancements(self, create, extracted, **kwargs):  # noqa: ANN001, ANN003, AR
...
-def to_indexable(reference: Reference) -> ReferenceSearchProjection:
-    """Convert a Reference to an ReferenceSearchProjection for ES test indexing."""
+def to_indexable(
+    reference: Reference,
+    linked_data_concepts: list[str] | None = None,
+) -> ReferenceSearchProjection:
+    """
+    Convert a Reference to an ReferenceSearchProjection for ES test indexing.
+
+    ``linked_data_concepts`` short-circuits the RDF projection pipeline and
```
jack-o-the-green, line 426: ❤️

jack-o-the-green (approved):
LGTM - one thought on domain models.

app/domain/references/models/models.py
Adam-Hammo: 🧠

PR #714, opened 2026-05-26
Correct sibling facet counts

### LINKED ISSUE #703: Handle sibling relationships in search facet counting

Adam-Hammo:
Closes #703 

Adds sibling-aware facet count aggregations to the repository. Four main themes:
- Move vocabulary preprocessing from projection to the shared cache so they can be reused when searching too
- Add `ReferenceESRepository` methods that generically handle sibling facet counting
- Add `search_service` plumbing that uses the vocabulary to generate concept sibling groups
- Add `vocabulary` query param to the facet endpoint

Validation:
- `vocabulary` is required if `concepts` is in both `facets` and the `query`. Then if we enter down that path:
  - All filtered concepts must be part of the vocab
  - All OR-grouped concepts must be siblings
  - Siblings cannot be in different AND-groups

NB these only apply to the facet count endpoint, not the search endpoint.

### DIFF
--- a/app/domain/references/repository.py
+++ b/app/domain/references/repository.py
@@ -375,7 +375,10 @@ class ReferenceESRepository(
 
         Non-empty ``grouping`` builds per-group filtered terms aggs plus an
         ``unselected`` agg, with concept filters applied via ``post_filter``
-        so they don't constrain the aggregations.
+        so they don't constrain the aggregations. Currently
+        :class:`FacetType.CONCEPTS` is the only facet, so this path only
+        returns concept buckets — extending to other facets is left to the
+        next implementer.
 
         :raises ESQueryError: If a group's sibling set exceeds ``max_buckets``.
         """
@@ -403,17 +406,7 @@ class ReferenceESRepository(
         concepts_buckets: list[ESFacetBucket] = []
         for spec in aggs:
             concepts_buckets.extend(results[spec.name])
-        out: dict[FacetType, list[ESFacetBucket]] = {
-            FacetType.CONCEPTS: concepts_buckets,
-        }
-        other_facets = tuple(f for f in facets if f is not FacetType.CONCEPTS)
-        if other_facets:
-            out.update(
-                await self._aggregate_facets_naive(
-                    query, other_facets, max_buckets=max_buckets
-                )
-            )
-        return out
+        return {FacetType.CONCEPTS: concepts_buckets}
 
     async def _aggregate_facets_naive(
         self,
@@ -447,8 +440,8 @@ class ReferenceESRepository(
         for i, group in enumerate(grouping.groups):
             other_clauses = tuple(
                 self._build_linked_data_concept_clause(other.source_filter)
-                for other in grouping.groups
-                if other is not group
+                for j, other in enumerate(grouping.groups)
+                if j != i
             )
             include = tuple(sorted(group.siblings_including_selected))
             aggs.append(
--- a/app/domain/references/repository.py
+++ b/app/domain/references/repository.py
@@ -330,12 +330,7 @@ class ReferenceESRepository(
     def _build_filter_clauses_excluding_concepts(
         self, query: SearchQuery
     ) -> list[Query]:
-        """
-        Build the year + annotation clauses, omitting concept filters.
-
-        The sibling-aware facet path applies concept filters via ``post_filter``
-        rather than ``bool.filter`` so that aggregations can ignore them.
-        """
+        """Year + annotation clauses only; the sibling-aware path AND-s concepts."""
         clauses: list[Query] = []
         if query.publication_year_range and (
             clause := self._build_publication_year_clause(query.publication_year_range)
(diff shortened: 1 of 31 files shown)


app/domain/references/services/search_service.py
```diff
@@ -59,10 +68,87 @@ async def aggregate_facets(
...
+            )
+        for i, group in enumerate(groups):
+            for other in groups[i + 1 :]:
+                overlap = (
+                    group.siblings_including_selected
+                    & other.siblings_including_selected
+                )
+                if overlap:
+                    msg = (
+                        "Two concept filters share a sibling set. Overlap: "
```
danielbreves, line 150: Could we add a test for this case?

app/domain/references/repository.py
```diff
@@ -353,25 +365,124 @@ async def aggregate_facets(
...
+            groups = sibling_groups_by_facet.get(facet)
+            if not groups:
+                continue
+            results[facet] = await self._aggregate_facet_sibling_aware(
+                query, facet, groups, max_buckets=max_buckets
+            )
+
+        return results
+
+    async def _aggregate_facet_sibling_aware(
```
danielbreves, line 410: This felt a bit dense. I wonder if it could be broken up a bit further or maybe we could just add some more inline comments to make it easier to follow?

app/domain/references/repository.py
```diff
@@ -355,25 +368,167 @@ async def aggregate_facets(
...
-            max_buckets=max_buckets,
+            self.default_search_fields,
+            self._build_filter_clauses(query, exclude_facet=facet),
+        )
+        return (
+            AsyncSearch(using=self._client, index=self._persistence_cls.Index.name)
+            .extra(size=0)
+            .query(composed)
+            .source(includes=[])
+            .post_filter(Bool(filter=list(group_clauses)))
```
danielbreves: `post_filter` only applies to hits, which is 0 here, so I think this is no-op? The tests seem to pass without it.

app/domain/references/routes.py
```diff
@@ -520,11 +525,24 @@ async def count_facets_for_search(
             description="One or more facet types to count.",
         ),
     ],
+    vocabulary: Annotated[
```
danielbreves, line 544: Could we validate/restrict this to `.evidence-repository.org`?

app/domain/references/repository.py
Adam-Hammo: Nice catch, this is left over from the original proposal where searching and counting was done together

app/domain/references/repository.py
```diff
@@ -355,25 +368,159 @@ async def aggregate_facets(
...
+        Run sibling-aware aggregation for one facet.
+
+        Each group's selection becomes a Terms clause. Aggs are wrapped in
+        ``filter`` aggs that AND in the *other* groups' selections — OR within
+        a group (multi-URI Terms); AND between groups.
+        """
+        field = self._FACET_FIELDS[facet]
+        group_clauses = [Terms(**{field: list(g.selected)}) for g in groups]
+
+        # Build search query wthout facet filter clauses
```
danielbreves: How about this?

```suggestion
        # Build search query excluding the facet's own filter so its agg includes sibling counts
```

danielbreves (approved):
Great work, LGTM!

Issue #255, opened 2025-08-15
Investigate asyncio resources not being closed


Adam-Hammo:
Noticed in honeycomb. Currently alert trigger has asyncio errors excluded, we should instead avoid the errors in the application and remove the exclusion.

See: https://ui.honeycomb.io/destiny-evidence/environments/production/datasets/destiny-repository-task-production/result/7wqtbbF2FHU?expd_col=_hny.ts&expd_fwd=false&expd_id=FVEtjfVDG69&expd_page=25&tab=explore

Eg:
```
{
  "batch_enhancement_request_id": "f45db96d-42c7-4387-8956-27a02c3c395c",
  "event": "Unclosed connector\nconnections: ['deque([(<aiohttp.client_proto.ResponseHandler object at 0x7e07d40118b0>, 1301895.89144)])']\nconnector: <aiohttp.connector.TCPConnector object at 0x7e07af62dd60>",
  "level": "error",
  "logger": "asyncio",
  "task_name": "app.domain.references.tasks:collect_and_dispatch_references_for_batch_enhancement"
}
```

```
{
  "batch_enhancement_request_id": "f45db96d-42c7-4387-8956-27a02c3c395c",
  "event": "Unclosed client session\nclient_session: <aiohttp.client.ClientSession object at 0x7e07af62d070>",
  "level": "error",
  "logger": "asyncio",
  "task_name": "app.domain.references.tasks:collect_and_dispatch_references_for_batch_enhancement"
}
```

(Also observed on validate&import enhancement task)

Adam-Hammo:
We also fail to close `BlobServiceClient` and `DefaultAzureCredential` in the application lifecycle at `/app/persistence/blob/clients/azure.py`. This is at startup time, so doesn't churn much.

PR #715, opened 2026-05-26
Convert blob client manager to a process-wide singleton

Adam-Hammo:
Addresses #255 

Async resources in the blob repository were being opened anew on each request and not being closed properly. 

This PR moves async resource management to the application lifecycle level, in line with our other persistence implementations (SQL & ES).


jaybea (changes requested):
I think app/run_task.py closes DB and ES but not blob.

Confirmed by running a blob-touching task through the real run_task(): after its finally closes DB and ES, the process-wide registry still holds an open client (httpx is_closed=False); calling close_blob_clients() releases it.

The job *is* short-lived so the OS reclaims the socket on exit (it's not an accumulating leak),  but the the client isn't cleanly closed, and I believe on the Azure path the managed-identity credential's pool never flushes through an async close - it just dies with the process.

Maybe time for async `init_persistence` and `close_persistence`? With the caveat that blobs client are lazy-initialised on first use, so they wouldn't have an `.init` call like SQL/ES, but they'd have a `await close_blob_clients()` on `close_persistence()`.

Or just fix `run_task()` in this PR with the one-liner.

I couldn't find any direct tests for thew new registry or teardown. Might be nice to have:
* get() returns the same instance for repeated calls of one location and distinct instances per location.
* aclose() calls aclose() on each held client, clears the registry, and logs (does not raise) when one client's close fails.
* AzureBlobStorageClient.aclose() closes both the BlobServiceClient and the managed-identity credential.
* The lifecycle hooks (main.py, tasks.py, and run_task.py once fixed) call close_blob_clients().

Migh consider holding `_lock` during `_BlobClentRegistry.aclose(). It's a long-shot, but some `get()` might get in during shutdown and store a client after the snapshot. Can't really happen at this point, but maybe a little insurance against a future use.

Tiny nit: `close_blob_clients()` is a module-level function wrapping `_registry_aclose(), whereas SQL/ES expose the manager object directly (`db_manager.close`, `es_manager.close()`). Minor asymetry between blob and SQL/ES. But if there's a good reason to not follow SQL/ES that's fine.

Adam-Hammo:
Thanks @jaybea, good points! Re `close_blob_clients()`, this decision is because there's multiple backend clients and so it made sense to me to abstract them so we don't need to propagate future blob implementations to the lifecycle handlers. I've implemented the rest.

jaybea (approved):
Changes look good. A couple of possible gaps to fix at your discretion:

I reverted the `run_task` fix. The leak came back and the unit suite still passed, so nothing that I could find is asserting a regression for the missing `close_blob_clients()`.

On the `test_blob_registry_get_reuses_client_per_location`, the fake returns one shared sentinel per location, so the assertions hold even if get() stops caching (I checked, a non-caching get() still passes it). This seems to be asserting the correctness of the routing, not the reuse.

jaybea:
@Adam-Hammo this lifecycle regression suite is sweet! Iron-clad!

PR #716, opened 2026-05-26
Add CLI to invoke subset reference index repair

danielbreves:
## Summary

- Adds `cli/invoke_reference_repair.py` — a CLI to trigger subset repair (`POST /v1/system/indices/reference/repair/` with `document_ids`) for a file of reference IDs, batched to the server's 1000-ID limit. Mirrors the existing `invoke_reference_deduplication.py` pattern (CLIAuth, `--env`).
- Supports `--batch-size`, `--start-batch` (resume after a partial failure), and `--dry-run`.
- Adds `cli/.env.production.example` and a `.gitignore` exception so `cli/.env.*.example` templates can be committed.

Motivation: backfilling new ES fields on the subset of references that have linked-data enhancements, without a corpus-wide `repair_all`.

## Test plan

- [ ] `uv run python -m cli.invoke_reference_repair --reference-id-file ids.txt --env <env> --dry-run` prints the batch plan without sending requests.
- [ ] A live run against staging queues `Repair index subset` tasks (visible in Honeycomb) and returns 202 per batch.

### DIFF
--- a/cli/invoke_reference_repair.py
+++ b/cli/invoke_reference_repair.py
@@ -2,22 +2,16 @@
 
 # ruff: noqa: T201
 import argparse
-from collections.abc import Iterator
 from pathlib import Path
 
 import httpx
 from fastapi import status
 
 from app.core.config import Environment
+from app.utils.lists import list_chunker
 from cli.auth import CLIAuth
 
 from .config import get_settings
-
-
-def _batched(items: list[str], size: int) -> Iterator[list[str]]:
-    """Yield slices of length up to ``size``."""
-    for i in range(0, len(items), size):
-        yield items[i : i + size]
 
 
 def invoke_reference_repair(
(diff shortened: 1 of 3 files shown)


cli/invoke_reference_repair.py
```diff
@@ -0,0 +1,133 @@
...
+from app.core.config import Environment
+from cli.auth import CLIAuth
+
+from .config import get_settings
+
+
+def _batched(items: list[str], size: int) -> Iterator[list[str]]:
+    """Yield slices of length up to ``size``."""
+    for i in range(0, len(items), size):
+        yield items[i : i + size]
```
Adam-Hammo: Nit: we have one of these already in `app.utils.lists`

Adam-Hammo (approved):

Adam-Hammo (approved):

PR #717, opened 2026-05-27
Add country and World Bank region facet aggregations

### LINKED ISSUE #710: Add country aggregations
## Context
After #705 and evidence-repo-ui#80 and we want to add aggregations to the CountryFilter to match https://claude.ai/design/p/4df77d9c-8dec-45f9-9a19-65cbce3b7cab?file=Fast+follow+prototype.html 

We want to add country as a structured parameter to both our search and aggregations endpoints to facilitate this. 

## Technical Scope
* Add a structured parameter to `/references/search` to allow searching by country  
    * Current scope is that this should only support ISO-2 codes as these are what is projected by our enhancments 
    * Use your judgement to decide (a) how to document this limitation, or (b) implement country name conversion for friendlier search requests.  
* Add the same structured parameter to the endpoint implemented in #698 and update the response to also include country aggregations
    * This will be used to populate the CountryFilter counts for each country.   

## Out of scope
This ticket does not include adding ability to search or aggregate by world bank regions.

danielbreves:
Closes #710

Adds `countries` and `country_wb_regions` facet types to `/references/search/facets/`, counting the already-indexed `linked_data_countries` (ISO-2) and `linked_data_country_wb_regions` (WB region IDs) keyword fields. Mirrors the existing `concepts` facet — naive term aggregation, no new query parameters.

WB region aggregation was out of scope in the issue but folded in here since it's the identical wiring and easier to review together. The structured country *search* parameter remains split out into a separate ticket.

### Note on #714
This is the naive aggregation. It overlaps the facet methods #714 reworks, so the second of the two to land rebases. `test_concept_filter_does_not_drop_country_facet` guards against #714's concept sibling-aware path silently dropping a co-requested country facet.

### DIFF
--- a/app/domain/references/services/anti_corruption_service.py
+++ b/app/domain/references/services/anti_corruption_service.py
@@ -418,6 +418,20 @@ class ReferenceAntiCorruptionService(GenericAntiCorruptionService):
                     )
                     for bucket in buckets
                 ]
+            elif facet is FacetType.COUNTRIES:
+                fields["countries"] = [
+                    destiny_sdk.references.CountryFacetCount(
+                        country=bucket.key, count=bucket.count
+                    )
+                    for bucket in buckets
+                ]
+            elif facet is FacetType.COUNTRY_WB_REGIONS:
+                fields["country_wb_regions"] = [
+                    destiny_sdk.references.CountryWBRegionFacetCount(
+                        country_wb_region=bucket.key, count=bucket.count
+                    )
+                    for bucket in buckets
+                ]
             else:
                 msg = f"facets_to_sdk has no SDK mapping for FacetType.{facet.name}"
                 raise NotImplementedError(msg)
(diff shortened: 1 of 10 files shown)


libs/sdk/src/destiny_sdk/references.py
Adam-Hammo, line 1: We'll need to bump the SDK version

app/domain/references/services/anti_corruption_service.py
```diff
@@ -418,6 +418,20 @@ def facets_to_sdk(
                     )
                     for bucket in buckets
                 ]
+            elif facet is FacetType.COUNTRIES:
+                fields["countries"] = [
```
Adam-Hammo, line 422: Mega-nit that I should have done in the first place - can we use eg `destiny_sdk.references.FacetType.COUNTRIES` to key?

Adam-Hammo (approved):
Looks good!

app/domain/references/services/anti_corruption_service.py
Adam-Hammo: Actually I see this is a false equivalence. Carry on.

PR #719, opened 2026-05-28
Add OR-based facet counting for countries and regions

### LINKED ISSUE #718: Structured search parameter for country
## Context 

De-scoped from initial deliverable 

## Technical scope 

destiny-repo
* Add a structured parameter to /references/search to allow searching by country
     * Current scope is that this should only support ISO-2 codes as these are what is projected by our enhancments
    * Use your judgement to decide (a) how to document this limitation, or (b) implement country name conversion for friendlier search requests.

evidence-repo-ui
* Update the searching to use the structured country filter parameter

Adam-Hammo:
Implements correct counting for country and region facets. On the way, #718 had to be implemented in order to know which filters to exclude.

- Adds aggregation path where the entire field is siblings (simpler than concepts)
- Configures countries and regions to use this path
- Adds countries and regions query parameters to search


app/domain/references/repository.py
```diff
@@ -496,8 +538,8 @@ def _attach_unselected_agg(
         max_buckets: int,
     ) -> str:
         """Attach the ``unselected`` agg: field values outside any group's siblings."""
-        all_grouped_uris = frozenset().union(
-            *(g.siblings_including_selected for g in groups)
+        all_grouped_uris: frozenset[str] = frozenset().union(
+            *(g.siblings_including_selected or () for g in groups)
```
jaybea: Nit: The `or ()` here is unreachable I believe. `_attach_unselected_agg` is only called at this line under `if all(... is not None)`, so groups are never None here, and the set always contains the selection so is never empty.

I know, it just feels better to include as belt-and-braces, so feel free to ignore.

app/domain/references/services/search_service.py
```diff
@@ -141,9 +177,8 @@ async def _resolve_concept_sibling_groups(
             )
         for i, group in enumerate(groups):
             for other in groups[i + 1 :]:
-                overlap = (
-                    group.siblings_including_selected
-                    & other.siblings_including_selected
+                overlap = (group.siblings_including_selected or frozenset()) & (
```
jaybea: Both of these `or frozenset()`'s never fire, this function only builds concept groups whose sibling set is always real.

jaybea (approved):
Excellent PR. Took me a while to get my head around the exclusions, when to AND and when to OR, and how those translate into ES queries.

I checked out the branch and ran it against a freshly reindexed local corpus: country and WB-region facets work end-to-end (`US` -> `NAC`, region buckets `NAC=2095, ECS=589, EAS=234, ...`), and sibling-aware universal mode behaves correctly - filtering `country=US` still returns the full 101-country distribution with the selected value intact.

Of note: it looks like the WB-region universal-mode facet counting is untested. Existing tests cover parsing, projection, field searchability, and unfiltered region facets, but not the new path where a region filter is applied while requesting the region facet. This is the same as the country sibling-aware test and should probably get service/repository/integration coverage. Runtime behavior  looks good, this is maybe a coverage gap, not a bug.

Also a couple of inline dead-code nits that you can consider.

app/domain/references/repository.py
Adam-Hammo: These exist to nudge mypy in the right direction as the field itself is nullable - I'll replace with a sturdier check of group shape.

PR #721, opened 2026-06-02
Allow robot.writer role to write robots

Adam-Hammo:
Latent problem exposed by keycloak (Azure automatically attached the scope on all requests through the client)


jaybea (approved):
Nice catch.

PR #726, opened 2026-06-03
Add second pass to reindex duplicates in bulk sync

### LINKED ISSUE #693: Re-index canonical references when duplicates are modified
Arises from https://covidence.slack.com/archives/C0810ARPCRG/p1779150745575739.

When enhancements are added to duplicate references, they are not re-indexed in Elasticsearch. This is because the synchronizer service silently drops non-canonical references (in order to avoid duplicate entries). Only canonical references are allowed to be indexed.

```python
for reference in references:
    if reference.is_canonical_like:
        yield await self._to_indexable(reference)
```

We should either:
- Update the synchronizer service to fetch and reindex the deduplicated projection of the canonical reference when it sees a duplicate
  - Gut feel is we should prefer this approach.
  - Note that I believe `sql_to_es` behaves correctly, only `bulk_sql_to_es` needs updating. 
- Pass in the canonical reference from `tasks.validate_and_import_robot_enhancement_batch_result`

We should also consider adding the "delete if duplicate" handling to `bulk_sql_to_es` (currently explicitly omitted per the docstring).

Adam-Hammo:
Closes #693 by adding a second pass that reindexes the canonicals of any duplicates requested for reindexing.

### DIFF
--- a/app/domain/references/services/synchronizer_service.py
+++ b/app/domain/references/services/synchronizer_service.py
@@ -116,17 +116,37 @@ class ReferenceSynchronizer(GenericSynchronizer[Reference]):
             AsyncGenerator[ReferenceSearchProjection, None]
         ):
             """Generate references for indexing."""
-            for reference_id_chunk in list_chunker(
-                ids,
-                chunk_size,
-            ):
+            indexed_ids: set[UUID] = set()
+            redirect_ids: set[UUID] = set()
+
+            for reference_id_chunk in list_chunker(ids, chunk_size):
                 references = await self.sql_uow.references.get_by_pks(
                     reference_id_chunk,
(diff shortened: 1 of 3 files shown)


jack-o-the-green (approved):
LGTM

Only question would be how expensive is double indexing a reference and how often would we actually expect it to happen? The code would be quite a bit simpler if we accepted the occasional double canonical index as we could avoid the second list chunk block.

Adam-Hammo:
Yeah fair point, definitely not often - in the times where it's most likely to occur, they'd more than likely be in different tasks anyway and so not collapsed here. Will simplify

jack-o-the-green (approved):
LGTM

PR #727, opened 2026-06-03
Add cross-facet endpoint and functionality

### LINKED ISSUE #722: Add cross-facet aggregation endpoint
# Context

This issue provides backend support for evidence maps https://github.com/destiny-evidence/evidence-repo-ui/issues/93 by adding an endpoint to perform cross-tabulation of facets.

# What we need

- A new endpoint `/references/search/cross-facets/`
- Reuses the same filters as searching and facet counting.
  - Must also take in `vocabulary` in order to restrict the concepts to the required schemes.
- Accepts exactly two axes, each one of (`countries`, `country_wb_regions` or any `concept_scheme` URI). Mixed types are allowed.
- Returns non-zero cells.
- Returns total results as well for UI display.

# Counting semantics

These differ from `/references/search/facets/` (net effect: simpler!):

1. **Each cell is a strict intersection** - records matching row ∧ column ∧ all panel filters ∧ query string. No sibling broadening (yay).
2. **Panel filters apply to the whole matrix.** If a filter overlaps an axis (e.g. region filtered to one value with that scheme on an axis), it still applies.
3. **None of /facets/'s filter-grouping rules apply here** - without broadening, multiple AND'd filters just narrow, which is fine. This isn't an immediate UI use-case but worth getting right.

# Worked example

```
GET /references/search/cross-facets/
    ?q=malaria
    &row=https://vocab.../scheme/WHORegion
    &column=https://vocab.../scheme/Themes
    &vocabulary=https://vocab.evidence-repository.org/published/01935.../vocabulary.ttl
    &start_year=2015
```
(`row`/`column` take a concept-scheme URI, or the literal `countries` / `country_wb_regions`.)

Which the server resolves to a single nested `terms` aggregation — concept axes scoped to their scheme's members via `include`:

```json
{
  "size": 0,
  "query": { "query_string": { "query": "..." } },
  "aggs": {
    "rows": {
      "terms": {
        "include": [
          "https://vocab.../WHORegion/AFR",
          "https://vocab.../WHORegion/EMR",
          "https://vocab.../WHORegion/EUR",
          "https://vocab.../WHORegion/AMR",
          "https://vocab.../WHORegion/SEAR"
        ],
        "size": 1000
      },
      "aggs": {
        "columns": {
          "terms": {
            "include": [
              "https://vocab.../Themes/Education",
              "https://vocab.../Themes/Health",
              "https://vocab.../Themes/Governance",
              "https://vocab.../Themes/Environment",
              "https://vocab.../Themes/Economics"
            ],
            "size": 1000
          }
        }
      }
    }
  }
}
```

A mixed-type map (e.g. concept rows × `country_wb_regions` columns) is the same shape with the column agg pointed at `linked_data_country_wb_regions` and `include` omitted (the field domain is the full code set).

**Response:** a `total`, plus a flat list of non-zero `{row, column, count}` cells - exact shape is the implementer's choice.

Adam-Hammo:
Closes #722 

- Adds endpoint to provide cross-facets for a given `row` and `column`, which can be countries, regions or concept scheme IDs
- Treats every concept in a scheme as a sibling - each will get its own entry in the axis regardless of hierarchical depth
  - NB there will be facet searching follow-ups to mirror this behaviour over searching/counting - see also https://github.com/destiny-evidence/evidence-repo-ui/pull/98#discussion_r3323286633
- Returns a model with every non-zero cell, keyed by row and column, and overall total

### DIFF
--- a/app/domain/references/routes.py
+++ b/app/domain/references/routes.py
@@ -640,6 +640,81 @@ async def count_facets_for_search(
     return anti_corruption_service.facets_to_sdk(buckets_by_facet)
 
 
+@search_router.get(
+    "/cross-facets/",
+    status_code=status.HTTP_200_OK,
+    responses={
+        status.HTTP_400_BAD_REQUEST: {
+            "description": "Bad Query String",
+            "model": APIExceptionContent,
+        }
+    },
+    description=(
+        "Cross-tabulate two axes over the references matching the search, for "
+        "evidence maps. Accepts the same filters as `/references/search/`, plus a "
+        "`row` and `column` axis.\n\n"
+        "When an axis is a scheme  URI, supply `?vocabulary=`.\n\n"
+        "**Counting.** Each returned cell is a strict intersection: references "
+        "matching its row value, its column value, all panel filters and the query "
+        "string. Only non-zero cells are returned.\n\n"
+        "Cells may sum to more than `total` because a reference can carry multiple "
+        "values on an axis. `total` is exact. A matrix whose row x column bucket "
+        f"count would exceed {settings.es_cross_facet_max_cells:,} is rejected."
+    ),
+)
+async def cross_tabulate_facets(
+    reference_service: Annotated[ReferenceService, Depends(reference_service)],
+    anti_corruption_service: Annotated[
+        ReferenceAntiCorruptionService, Depends(reference_anti_corruption_service)
+    ],
+    query: Annotated[SearchQuery, Depends(parse_search_query)],
+    row: Annotated[
+        str,
+        Query(
+            min_length=1,
+            description=(
+                "The row axis: `countries`, `country_wb_regions`, or a concept-scheme "
+                "URI."
+            ),
+            examples=["country_wb_regions"],
+        ),
+    ],
+    column: Annotated[
+        str,
+        Query(
+            min_length=1,
+            description=(
+                "The column axis: `countries`, `country_wb_regions`, or a "
+                "concept-scheme URI."
+            ),
+            examples=["https://vocab.evidence-repository.org/scheme/Themes"],
+        ),
+    ],
+    vocabulary: Annotated[
+        HttpUrl,
+        AfterValidator(_validate_vocab_host),
+        Query(
+            description=(
+                "Vocabulary URI scoping concept-scheme axes to their members. "
+                "Required when `row` or `column` is a scheme URI. "
+                f"Host must be a subdomain of `{settings.vocabulary_host}`."
+            ),
+            examples=[
+                "https://vocab.evidence-repository.org/published/01935f56-2c8e-7000-8000-000000000001/vocabulary.ttl"
+            ],
+        ),
+    ] = None,
+) -> destiny_sdk.references.ReferenceCrossFacetResult:
+    """Cross-tabulate two axes for references matching the query."""
+    cells, total = await reference_service.aggregate_cross_facet(
+        query,
+        row,
+        column,
+        vocabulary_uri=str(vocabulary) if vocabulary else None,
+    )
+    return anti_corruption_service.cross_facet_to_sdk(cells, total)
+
+
 @exports_router.post(
     "/",
     status_code=status.HTTP_202_ACCEPTED,
--- a/app/domain/references/routes.py
+++ b/app/domain/references/routes.py
@@ -651,14 +651,15 @@ async def count_facets_for_search(
     },
     description=(
         "Cross-tabulate two axes over the references matching the search, for "
-        "evidence maps. Accepts the same filters as `/references/search/`, plus a "
-        "`row` and `column` axis.\n\n"
-        "When an axis is a scheme  URI, supply `?vocabulary=`.\n\n"
+        "evidence maps. Accepts the same filters as `/references/search/`, plus the "
+        "two `axes`.\n\n"
+        "When an axis is a concept-scheme URI, supply `?vocabulary=`.\n\n"
         "**Counting.** Each returned cell is a strict intersection: references "
-        "matching its row value, its column value, all panel filters and the query "
-        "string. Only non-zero cells are returned.\n\n"
+        "matching both its axis values, all panel filters and the query string. "
+        "Only non-zero cells are returned, sorted by descending count.\n\n"
         "Cells may sum to more than `total` because a reference can carry multiple "
-        "values on an axis. `total` is exact. A matrix whose row x column bucket "
+        "values on an axis (and to less, as a reference matching the filters need "
+        "not have a value on either axis). `total` is exact. A matrix whose bucket "
         f"count would exceed {settings.es_cross_facet_max_cells:,} is rejected."
     ),
 )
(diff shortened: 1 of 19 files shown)


app/domain/references/services/search_service.py
```diff
@@ -108,6 +123,94 @@ async def aggregate_facets(
...
+    ) -> tuple[list[CrossFacetCell], ESSearchTotal]:
+        """
+        Cross-tabulate two axes over references matching ``query``.
+
+        Each axis is a literal axis or a concept-scheme URI (scoped to its members
+        via ``vocabulary_uri``). Returns the non-zero cells and the exact grand total.
+        """
+        scheme_members: dict[str, frozenset[str]] | None = None
+        if vocabulary_uri and (
+            self._literal_axis_facet(row_token) is None
```
jack-o-the-green: `nitpick` Would be nice to reverse the logic here for readability, even if we just wrapped the `_literal_axis_facet(token)` in a `_is_concept_scheme` or similar.

app/domain/references/routes.py
```diff
@@ -640,6 +640,81 @@ async def count_facets_for_search(
...
+        }
+    },
+    description=(
+        "Cross-tabulate two axes over the references matching the search, for "
+        "evidence maps. Accepts the same filters as `/references/search/`, plus a "
+        "`row` and `column` axis.\n\n"
+        "When an axis is a scheme  URI, supply `?vocabulary=`.\n\n"
+        "**Counting.** Each returned cell is a strict intersection: references "
+        "matching its row value, its column value, all panel filters and the query "
+        "string. Only non-zero cells are returned.\n\n"
```
jack-o-the-green: I'm curious as to why we're only returning non-zero cells. I can see it reducing the size of our response but it feels like it would be better for consumers to receive the entire table shape? Happy to be wrong here. 

On some further reading it looks like we're not returning table shapes anyway, but highest count combinations first (so any zero results would all go at the bottom anyway). In this case I am wondering whether we want to move away from `row` and `column` as parameters as I feel there's some implied structure in these terms? I'd be happier if we called them `axes` 

Feel free to discount this comment, just a bunch of musing about what language implies.

app/domain/references/routes.py
```diff
@@ -640,6 +640,81 @@ async def count_facets_for_search(
     return anti_corruption_service.facets_to_sdk(buckets_by_facet)
 
 
+@search_router.get(
+    "/cross-facets/",
```
jack-o-the-green, line 644: `nitpick` endpoint doesn't match endpoint described in ticket #727 (this is fine but maybe worth updating in ticket clarity)

docs/procedures/search.rst
```diff
@@ -254,6 +254,36 @@ Limitations
...
 Each facet returns at most ``ES_AGGREGATION_MAX_BUCKETS`` buckets (default 1000). For concept facets, if a sibling group's vocabulary set exceeds this the request is refused rather than silently truncated.
 
+.. _cross-facets-procedure:
+
+API Cross-Facet Counts
+^^^^^^^^^^^^^^^^^^^^^^^
+
+The cross-facets endpoint at `/v1/references/search/cross-facets/` cross-tabulates two axes over the references matching the search, for evidence maps. It accepts the same filter parameters as `/v1/references/search/` plus a ``row`` and a ``column`` axis.
+
+Each axis is the literal ``countries``, the literal ``country_wb_regions``, or a concept-scheme URI; mixed types are allowed. When an axis is a scheme URI, supply ``vocabulary=`` — its concepts scope that axis (via the SKOS membership of the scheme, at every depth).
```
jack-o-the-green: `nitpick` can we make this 

```
Each axis is one of 
    * A concept-scheme URI 
    * The literal countries
    * The literal country_wb_regions
```

and then anywhere we say `scheme URI` we should update that to `concept-scheme URI`.

docs/procedures/search.rst
```diff
@@ -254,6 +254,36 @@ Limitations
...
+        &column=https://vocab.evidence-repository.org/scheme/Themes
+        &vocabulary=https://vocab.evidence-repository.org/vocabulary/v1
+        &start_year=2015
+
+Counting is simpler than the facets endpoint: each returned cell is a strict intersection - references matching its row value, its column value, all panel filters and the query string. Only non-zero cells are returned.
+
+Returns
+""""""""""
+
+Returns a :class:`ReferenceCrossFacetResult <libs.sdk.src.destiny_sdk.references.ReferenceCrossFacetResult>`: an exact ``total`` plus the non-zero ``{row, column, count}`` cells, sorted by descending count. Cells may sum to more than ``total`` because a reference can carry multiple values on an axis (e.g. several countries), so it contributes to multiple cells.
```
jack-o-the-green: This is the only place that `sorted by descending count` is referenced, it's implemented here right? https://github.com/destiny-evidence/destiny-repository/pull/727/changes#diff-4bc9de6d060cf03d0428526832b176c3b395472038f676d2b65ce05106f635e3R611. Maybe good to call out on the endpoint description also.

tests/integration/test_cross_facets.py
```diff
@@ -0,0 +1,333 @@
...
+            "q": "*",
+            "row": TOPICS_SCHEME,
+            "column": REGION_SCHEME,
+            "vocabulary": primed_vocab,
+        },
+    )
+    assert response.status_code == status.HTTP_200_OK, response.text
+    body = response.json()
+    assert body["total"] == {"count": 5, "is_lower_bound": False}
+    assert body["cells"] == [
```
jack-o-the-green, line 196: We should update the test fixture so that it demonstrates the row mixing in the responses that happens with descending order cell sorting. If you add another reference document for `[BOTANY, ZOOLOGY, AFRICA]` then the output of this will be

```
  {"row": BOTANY, "column": AFRICA, "count": 4},
  {"row": ZOOLOGY, "column": AFRICA, "count": 2},
  {"row": MICROBIOLOGY, "column": EUROPE, "count": 1},
  {"row": ZOOLOGY, "column": ASIA, "count": 1},
```

which demonstrates the row mixing. The current implementation just happens to look like it could imply alphabetical sorting for row followed by column as well `Botany -> Microbiology -> Zoology`, then within Zoology `Africa -> Asia` all have alphabetical precedence.

app/domain/references/routes.py
Adam-Hammo: To me it significantly reduces the size of the response - if we cross HPV's countries with study design, we get ~6,000 entries, realistically most of which will be zero. We also know the consumer knows the full scheme, as it passes in the vocabulary uri, so it can reconstruct gaps if it wants to. 

In my opinion, returning the list instead of a table shape is a bit nicer for general consumption - it's simpler, streamable, orderable and more generalisable. If we want to reintroduce zero cells, consumers could then choose to stop reading the response after it sees the first zero.

Axes I see the point for - will make that change.

jaybea (approved):
First off, excellent implementation. Ran it through its paces and stepped through:

Literal x literal, countries x WB regions: total 13,395 on my corpus, 154 result cells - US x NAC 2020, GB x ECS 184, ...
Scheme x literal with a published vocabulary, OutcomeScheme x countries: 336 cells, top OutcomeScheme/C00122 x US 2019, ...
A filtered run, countries x WB regions with country=US: total narrowed 13,395 to 2,020, 11 cells, and the surviving non-US rows are co-occurrences only - rule 2 working as specified.
Very cool!

Some should-fixes, at your discretion:

I couldn't find any cross-facet test that includes a panel filter (like run 3 above), so rules 2 and 3 are implicit rather than pinned for regression - the divergence from /facets/ is carried entirely by _build_filter_clauses(query) having no exclude_facet at app/domain/references/repository.py:575 (vs :459 on the facets path), and a future harmonising refactor would pass the whole suite. Cheapest pin: a country= filter on the literal-axes case. Could also add multiple AND'd country filters (rule 3) and a scheme axis with an overlapping concept= filter to be super-thorough.

`vocabulary_fetch_exception_handler` maps upstream failures to client 400s, which the client can't act on. Granted this is a long-shot occurrence since vocab artifacts are cached by versioned URI with no expiry, so it only fires in cache-miss windows: a cold process after a deploy or scale-out coinciding with a vocab-service problem, or the first fetches of a newly published vocabulary version. Those rare cases are also why a clear status code would help when it does fire - a trickle of 400s right after a deploy reads as client bugs rather than an upstream dependency being down. I acknowledge it may be rare enough to leave as is.

Super-nit: the cell guard is off by the count of the parent buckets - the nested aggregation materialises the row buckets too, so the worst case is row.size * (column.size + 1) against search.max_buckets, not row.size * column.size. One-line fix in _validate_cross_facet_cell_count, or soften the config comment to best-effort.

Just an FYI, you've probably thought of this already: the static guard bounds the nominal matrix, but terms aggs\regations only materialise observed buckets, and real matrices are sparse (154 of 2,048 nominal in my run 1). If a larger vocabulary lands on an axis someday (DESTINY's 500+ terms? HPV?), a cheap marginal pre-pass - two flat terms aggs under the same query - could shrink each scheme axis's include to observed values and make the guard adaptive, with no API change.

One scope note: the new exception handler registers app-wide, so vocabulary fetch failures on the existing /facets/ endpoint also move from unhandled 500 to 400. An improvement, just flagging since it lands silently in a cross-facets PR.

jack-o-the-green (approved):
LGTM

PR #729, opened 2026-06-03
Group facet count siblings by concept schemes

### LINKED ISSUE #728: Treat entire concept schemes as siblings
Currently sibling groups when searching/counting facets are treated as only those next to each other in the concept scheme hierarchy. This does not actually represent our coding structure, where data mapped to a parent concept should be OR-joinable with its children.

See also https://github.com/destiny-evidence/evidence-repo-ui/pull/98#discussion_r3323286633 where this was picked up.

**Ticket scope**
Back-compatible change to the facet count endpoints (to still allow the UI to do the depth-sensitive sibling counting while awaiting updating). Assess feasibility of the back-compatibility, if too destructive we can consider going for broke and releasing in parallel. 
A separate UI ticket will update the searching patterns.

Adam-Hammo:
Closes #728 

Fixes facet counting to group by concept scheme instead of direct siblings in hierarchical concept schemes:
- Adjusts query parameter validation
- Changes `SiblingGroup` building accordingly (simpler!)

Based off #727 in order to piggy-back off the new `VocabClient._build_scheme_members()`.

To be released in lockstep with https://github.com/destiny-evidence/evidence-repo-ui/pull/114.
- I considered backwards compatibility, but figured given low usage in ANZ hours and general tight timelines we could just get away with this one.


jaybea (approved):
The PR absolutely does it job, drops prodution line count (Woohoo!), and is correct in all ways I could come up with.

2 pre-existing nits, that only surface for a mention here because scheme membership resolution accuracy is getting so critical with facets. And further caveat that yes, our integration suite covers the shapes our current, carefully crafted vocabularies use:

1. `hasTopConcept` *ONLY* path in `_build_scheme_members` isn't isolated by any test. The Apple Pear fixture has two paths, `skos:hasTopConcept` (via Fruits) AND `skos:inScheme`, so the "Flat scheme via skos:hasTopConcept" assertion in `test_scheme_members_covers_every_membership_shape` would still pass if the `hstTopConcept` handling were deleted, the `inScheme` declarations would carry it. Easy fix here is drop the redundant `inScheme` from Apple/Pear.
2. The `broader` fallback union across multiple parents never fires in tests:
```
for parent in parents_by_concept.get(concept, ()):
    inherited |= schemes_for(parent, seen)

``` 
Quantum declares an explicit `skos:inScheme ex:Sciences`, so `schemes_for` early-returns and never follows its two `broader` parents. The only fallback test (`test_descendant_without_inscheme_inherits_via_broader`) is a single-parent chain.

PR #734, opened 2026-06-04
Prepare repo for HPV data

### LINKED ISSUE #731: Prepare repo for HPV data
Part of the HPV ingestion epic [#725](https://github.com/destiny-evidence/destiny-repository/issues/725).

## Why

To unblock UI development against the real backend stack with HPV data, two things need to land together:

1. HPV references available in the local dev DB.
2. The projection service must discover `evrepo:hasAppliedConcept` — today's unwrapped-property SPARQL at [`linked_data_projection_service.py:218-236`](app/domain/references/services/linked_data_projection_service.py#L218-L236) excludes `?range = skos:Concept`, so the property is never discovered and `linked_data_concepts` stays empty for HPV.

## Prerequisite

The HPV staging vocab needs to be bootstrapped with `evrepo:LinkedDataEnhancement`, `evrepo:Investigation`, `skos:Concept` (as `owl:Class`), `evrepo:hasInvestigation`, and `evrepo:hasAppliedConcept` (domain `Investigation`, range `skos:Concept`, cardinality `multiple`) — so the HPV LDEs in the seed can point at a valid vocab.

## Part A — Discover `hasAppliedConcept` in the projection

`_build_unwrapped_concept_properties` needs to handle the case where an unwrapped concept property's `rdfs:range` is the bare `skos:Concept` class. One suggested approach is a third UNION branch in the SPARQL:

```sparql
UNION
{
    FILTER(?range = skos:Concept)
    ?concept a skos:Concept ;
             skos:inScheme [] .
}
```

Branches 1 and 2 would stay untouched (branch 2's `FILTER(?range != skos:Concept)` keeps the *specific subclass* case separate from the new *generic concept* case). Other approaches welcome.

Either way, the change should be regression-safe for ESEA — ESEA declares no `owl:ObjectProperty` with `rdfs:range skos:Concept`, so discovery against the ESEA vocab is unchanged.

## Part B — Seed HPV references

Add HPV references to [`.db_seed/local.sql`](.db_seed/local.sql), alongside the existing seed data. Mirror the layout of the existing ESEA `linked_data` example at [`local.sql:67-96`](.db_seed/local.sql#L67-L96) — bibliographic / abstract / linked_data enhancements per reference, with external identifiers and an optional canonical duplicate-decision row.

The HPV LDE shape is documented by the fixtures in [evidence-repo-ui#112](https://github.com/destiny-evidence/evidence-repo-ui/pull/112) — the `content.data` JSON-LD blob is the canonical shape to mirror; reference IDs and other surrounding metadata can be picked fresh for the seed (no need to match the fixture UUIDs). Bibliographic data can be drawn from [HPV Map v3 Prelim Data.json](https://github.com/user-attachments/files/28500488/HPV.Map.v3.Prelim.Data.json) — each fixture's file name matches a real EPPI reference. The `linked_data` enhancement's `vocabulary_uri` should point at the HPV staging vocab URL.

⚠️ Seed writes bypass `LinkedDataValidationService` — SHACL doesn't run on these inserts, so HPV's flat shape doesn't need to pass the existing ESEA-shaped shapes today. SHACL coverage for HPV is tracked in [#723](https://github.com/destiny-evidence/destiny-repository/issues/723); until that lands, the bypass is what makes this approach work.

## Verify

- `./recreate-local.sh` completes; HPV references appear in pg and ES after reindex.
- Search returns HPV refs.
- Concept facet returns non-empty HPV concept counts.
- ESEA behaviour unchanged.

## Out of scope

- SHACL coverage for HPV — tracked in [#723](https://github.com/destiny-evidence/destiny-repository/issues/723).
- vocabulary-mapping-robot work — covered by [vocabulary-mapping-robot#48](https://github.com/destiny-evidence/vocabulary-mapping-robot/issues/48).
- Production rollout — local dev only.

danielbreves:
Closes #731. Mostly to unblock UI development against the real backend with HPV data.

- **Discover \`evrepo:hasAppliedConcept\`.** Adds a third UNION branch to \`_build_unwrapped_concept_properties\` that matches \`?prop rdfs:range skos:Concept\` directly. Existing ESEA discovery (range = scheme / range = ConceptClass subclass) is unchanged.
- **Seed 6 HPV references.** Bibliographic + abstract + linked_data enhancements mirroring the ESEA pattern, with reference IDs and \`hasAppliedConcept\` arrays taken from [evidence-repo-ui#112](https://github.com/destiny-evidence/evidence-repo-ui/pull/112) verbatim and bibliographic data from \`HPV Map v3 Prelim Data.json\`. Vocab URI points at the staging HPV taxonomy (\`.ttl\`).

> [!NOTE]
> The seeded \`hasAppliedConcept\` arrays come from PR #112 (hand-curated) — they don't yet reflect a programmatic mapping from EPPI codings. The existing EPPI codes on \`HPV Map v3 Prelim Data.json\` records were tagged against an older codeset and have zero overlap with the new HPV Taxonomy v1 codeset (\`381669\`) that [vocabulary-mapping-robot#49](https://github.com/destiny-evidence/vocabulary-mapping-robot/pull/49) targets. Once records are re-coded against \`381669\`, this seed should be regenerated programmatically.


jack-o-the-green (approved):
LGTM

And nice that you've restricted the discovery of the property to vocabularies containing concept schemes 👌

.db_seed/local.sql
```diff
@@ -94,6 +100,24 @@ public	Toy Robot	e2277656-4930-4c5a-8718-c3eb422385bb	annotation	0.2.0	{"annotat
...
+public	Vocabulary Mapping Robot	019e8b08-21a0-76b8-8abe-5c6313499de5	linked_data	1.0.0	{"enhancement_type":"linked_data","vocabulary_uri":"https://vocab.staging.evidence-repository.org/published/019e86db-fa16-71e0-8d89-390f6168ff59/1.1/vocabulary.ttl","data":{"@context":"https://vocab.staging.evidence-repository.org/published/019e86db-fa16-71e0-8d89-390f6168ff59/1.1/context.jsonld","@type":"LinkedDataEnhancement","hasInvestigation":{"@type":"Investigation","hasAppliedConcept":["https://vocab.aliveevidence.org/hpv/Country/NG","https://vocab.aliveevidence.org/hpv/WorldBankRegion/HPVV0171","https://vocab.aliveevidence.org/hpv/StudyDesign/HPVV0148","https://vocab.aliveevidence.org/hpv/CountryClassification/HPVV000216","https://vocab.aliveevidence.org/hpv/StudyDesign/HPVV000257","https://vocab.aliveevidence.org/hpv/TargetPopulation/HPVV0041","https://vocab.aliveevidence.org/hpv/TargetPopulation/HPVV000211","https://vocab.aliveevidence.org/hpv/EquityDimension/HPVV0011","https://vocab.aliveevidence.org/hpv/WHORegion/HPVV0179","https://vocab.aliveevidence.org/hpv/UNICEFRegion/HPVV0178"]}}}	7c3f576d-4212-78a8-993c-ede3df6c4e61	2026-06-04 12:00:00.000000+00	2026-06-04 12:00:00.000000+00	\N
```
jack-o-the-green, line 105: Do we want to add fake raw enhancements that these are derived from? Likely not a huge value but thought I'd call it out. Not sure we want to hand-roll too many fake eppi exported data

danielbreves: Good call out. I think given these are not even using the right codeset and the UI doesn't use it I might leave it out for now, but probably something we could include when we have real data?

PR #737, opened 2026-06-05
Fix token expiry detection

Adam-Hammo:
Noticed by @harryjmoss - thank you!

The API emits `Token is expired` on expiry, and the client detects `Token *has* expired` to re-request a new token. This fixes and standardises to a shared variable to avoid future drift.


Adam-Hammo:
Pls ignore the TF failure, Azure API timed out across the board 👍 👍 👍

danielbreves (approved):

PR #741, opened 2026-06-09
Add search endpoint that returns ids

Adam-Hammo:
Adds a search endpoint to return all IDs, without pagination.

See https://github.com/futureevidence/ai-evidence-summariser/issues/1 for context. Endpoint flagged as experimental for now in case we land on a different approach long-term.

Virtually entirely made of existing work which is beautiful!

### DIFF
--- a/app/domain/references/routes.py
+++ b/app/domain/references/routes.py
@@ -576,7 +576,6 @@ async def search_references(
     )
 
 
-@experimental
 @search_router.get(
     "/ids/",
     status_code=status.HTTP_200_OK,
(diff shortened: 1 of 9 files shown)


app/domain/references/routes.py
```diff
@@ -576,6 +576,40 @@ async def search_references(
     )
 
 
+@experimental
```
danielbreves: It looks like `description=` overrides __doc__ so the decorator is noop. Also it would need to run first to mutate the docstring before the route registration.

danielbreves (approved):
Legend!

PR #744, opened 2026-06-10
Relax SHACL core shapes for HPV flat linked-data enhancements

### LINKED ISSUE #723: HPV: SHACL design + validation for HPV linked-data shape
## Why

The bundled SHACL shapes ([`app/static/evrepo-core-shapes.ttl`](app/static/evrepo-core-shapes.ttl)) currently require an Education-style hierarchy:

- `LinkedDataEnhancement` must have `hasInvestigation` ≥ 1
- `Investigation` must have `hasFinding` ≥ 1
- `Finding` must have at least one of `hasContext`, `hasOutcome`, `hasEffectEstimate`, `evaluates`, `comparedTo`, `hasArmData`

HPV references are flat (bibliographic + applied codes, no Findings per the design). HPV `linked_data` enhancements will be rejected by these core shapes regardless of which unwrapped flavour the wire-shape decision lands on.

This blocks:

- The vocabulary-mapping-robot's POSTs (linked_data enhancements get stripped on ingest if SHACL fails — see [`LinkedDataValidationService`](app/domain/references/services/linked_data_validation_service.py)).
- Any direct ingest of HPV linked_data enhancements.

## Design questions to settle before implementing

A few options — pick whichever fits the team's direction:

- **Make core shapes community-aware** — split the Investigation/Finding requirements into community-specific shape sets, apply based on the reference's community.
- **Relax core requirements** — make `hasInvestigation`/`hasFinding` optional at the core level; keep stricter shapes elsewhere for Education-specific validation.
- **Load per-vocabulary shapes from taxonomy-builder** — instead of bundling shapes here, publish them per project alongside the vocab + context.
- Some combination.

## Deliverables

- Pick a design direction; write it up briefly in the PR description.
- Update SHACL so a preliminary HPV linked_data enhancement (matching the wire shape from [evidence-repo-ui#111](https://github.com/destiny-evidence/evidence-repo-ui/issues/111)) passes validation.
- Tests in [`tests/unit/domain/references/services/test_linked_data_validation_service.py`](tests/unit/domain/references/services/test_linked_data_validation_service.py) covering both:
  - A valid HPV `linked_data` enhancement passes.
  - Existing Education shape coverage still works (no regressions).

danielbreves:
Closes #723

## Design direction

The bundled SHACL shapes required an Education-style hierarchy on every `LinkedDataEnhancement` (`Investigation` → `hasFinding` ≥ 1). HPV enhancements (per [evidence-repo-ui#112](https://github.com/destiny-evidence/evidence-repo-ui/pull/112)) nest an `Investigation` that carries `hasAppliedConcept` codes instead of findings, so the *only* failing constraint was `InvestigationShape`'s `hasFinding` ≥ 1.

Of the options in #723, this takes the **relax-the-core** route for now. The two community-aware alternatives were set aside:

- **Key off `vocabulary_uri`** — varies per environment, so not a stable selector.
- **Key off the reference's `domain-inclusion` annotation** — the robot enhancement-import path only has the enhancement in hand (the reference isn't loaded until after validation), so this would add a per-enhancement reference load on the hot path plus an ordering assumption.

**Note:** `sh:class skos:Concept` only resolves against the graph fetched from the enhancement's `vocabulary_uri`, so every applied-concept IRI must be defined in the referenced vocab.

### DIFF
--- a/app/static/evrepo-core-shapes.ttl
+++ b/app/static/evrepo-core-shapes.ttl
@@ -1,5 +1,6 @@
 @prefix evrepo: <https://vocab.evidence-repository.org/> .
 @prefix sh:     <http://www.w3.org/ns/shacl#> .
+@prefix skos:   <http://www.w3.org/2004/02/skos/core#> .
 
 # =============================================================================
 # SHACL Shapes for Evidence Repository Core Vocabulary
@@ -18,14 +19,17 @@ evrepo:LinkedDataEnhancementShape a sh:NodeShape ;
         sh:minCount 1 ;
     ] .
 
-# --- Investigation must have at least one Finding ---
+# --- Investigation may contain Findings (each must be a Finding) ---
 
 evrepo:InvestigationShape a sh:NodeShape ;
     sh:targetClass evrepo:Investigation ;
     sh:property [
         sh:path evrepo:hasFinding ;
         sh:class evrepo:Finding ;
-        sh:minCount 1 ;
+    ] , [
+        sh:path evrepo:hasAppliedConcept ;
+        sh:nodeKind sh:IRI ;
+        sh:class skos:Concept ;
     ] .
 
 # --- Finding must be non-empty (at least one of the structural links present) ---
--- a/tests/unit/domain/references/services/test_linked_data_validation_service.py
+++ b/tests/unit/domain/references/services/test_linked_data_validation_service.py
@@ -207,6 +207,61 @@ async def test_hpv_investigation_without_finding_conforms(
 
 
 @pytest.mark.asyncio
+async def test_investigation_with_multiple_findings_conforms(
+    service: LinkedDataValidationService,
+):
+    """An Investigation may hold more than one Finding."""
+    data = {
+        "@context": {"evrepo": EVREPO},
+        "@type": "evrepo:LinkedDataEnhancement",
+        "evrepo:hasInvestigation": {
+            "@type": "evrepo:Investigation",
+            "evrepo:hasFinding": [
+                {
+                    "@type": "evrepo:Finding",
+                    "evrepo:hasOutcome": {
+                        "@type": "evrepo:Outcome",
+                        "evrepo:name": "Reading comprehension",
+                    },
+                },
+                {
+                    "@type": "evrepo:Finding",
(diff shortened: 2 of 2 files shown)


app/static/evrepo-core-shapes.ttl
```diff
@@ -18,14 +19,17 @@ evrepo:LinkedDataEnhancementShape a sh:NodeShape ;
...
 
 evrepo:InvestigationShape a sh:NodeShape ;
     sh:targetClass evrepo:Investigation ;
     sh:property [
         sh:path evrepo:hasFinding ;
         sh:class evrepo:Finding ;
-        sh:minCount 1 ;
+    ] , [
+        sh:path evrepo:hasAppliedConcept ;
+        sh:nodeKind sh:IRI ;
```
jack-o-the-green, line 31: Just confirming we want the generalized IRI here instead of a URI?

tests/unit/domain/references/services/test_linked_data_validation_service.py
```diff
@@ -170,6 +195,89 @@ async def test_ontology_concept_types_resolved_via_graph_merge(
     assert result.conforms, f"Expected conforms=True, got errors: {result.errors}"
 
 
+@pytest.mark.asyncio
+async def test_hpv_investigation_without_finding_conforms(
```
jack-o-the-green, line 199: Having a look at the tests, there's no test data that contains multiple findings and I am also wondering about the impact of whether we're allowing findings + hasApplied Concepts in the same enhancement and whether we should add a test that documents that that as intentional (I think we do, and should allow both)

jack-o-the-green (approved):
LGTM - ignore testing gap comments in favour of speed if necessary (they should be quick to add though)

app/static/evrepo-core-shapes.ttl
danielbreves: I don't think there is a sh:URI https://www.w3.org/TR/shacl/#x4.1.3-sh:nodekind

jack-o-the-green: Oh my apologies! I saw the `URI` was bolded in the note https://www.w3.org/TR/rdf11-concepts/#dfn-iri and assumed it was a thing.

PR #746, opened 2026-06-11
Fix token refresh flow in oauthclient

Adam-Hammo:
Required to unblock DESTINY abstract loadings (and possibly will also be needed for the full text robot script!)

### DIFF
--- a/libs/sdk/pyproject.toml
+++ b/libs/sdk/pyproject.toml
@@ -36,7 +36,7 @@ license = "Apache-2.0"
 name = "destiny_sdk"
 readme = "README.md"
 requires-python = ">=3.12, <4"
-version = "0.12.5"
+version = "0.12.6"
 
 [project.optional-dependencies]
 labs = []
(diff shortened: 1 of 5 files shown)


jaybea (approved):
This does the job, but there might be a narrow fix that avoid a possible performance hit down the road (if we ever stream authenticated downloads?).

I validated the fix by removing it and `test_token_refresh_through_real_client` fails with exactly `ResponseNotRead`as desired.

### Suggestion (non-blocking): read the body on the 401 branch instead of the class flag

`requires_response_body = True` requires full buffering for *every* response before the caller seeit, including 2xx, including successful ones (a streaming dwonload with 200, or 206 for Partial Content)? It seems like this logic is only needed for the 401 case. I think instead of holding one chunk at a time, this would cause httpx to pull the whole thing into memory before we see it.

Maybe:

```python
if response.status_code == httpx.codes.UNAUTHORIZED:
    response.read()
    try:
        json_response: dict = response.json()
    ...
```

I tried this narrow approach and all 14 tests pass.

PR #747, opened 2026-06-11
Add EPPI import CLI and parser improvements for HPV ingest

jack-o-the-green:
## Summary

- `cli/post_import_file.py` - cli command to post an import file to Destiny. adds a configurable `--poll-interval`.
- `cli/request_enhancements.py` - cli command to request enhancements against a search query
`_parse_url_to_identifier` now resolves `doi.org`/`dx.doi.org` URLs to a `DOIIdentifier`, with dedup.
- If the `--include-eppi-id` flag is set, every reference's `ItemId` is now parsed first as an `OtherIdentifier`. As a result, references with an `ItemId` always have at least one identifier.
- Annotations formatted for search can be added
- Domain inclusion annotation is required when using `ItemId` as an identifier (an annotation with the scheme `domain-inclusion`


cli/post_import_file.py
```diff
@@ -0,0 +1,205 @@
...
+        response.raise_for_status()
+        import_batch = ImportBatchRead.model_validate(response.json())
+        print(import_batch)
+        if import_batch.status == "completed":
+            print("Import batch complete.")
+            break
+        print(f"Import batch not complete, sleeping for {poll_interval} seconds...")
+        time.sleep(poll_interval)
+    else:
+        print("Import batch did not complete in time.")
```
Adam-Hammo: Very nitty - can we give the command to call `get_import_batch_summary` here to assist follow-up?

cli/post_import_file.py
```diff
@@ -0,0 +1,205 @@
...
+    """Poll for completion and produce a summary of the import."""
+    print(f"Polling import batch {import_batch_id} for completion...")
+    for _ in range(5):
+        response = client.get(
+            f"/v1/imports/records/{import_record_id}/batches/{import_batch_id}/"
+        )
+        response.raise_for_status()
+        import_batch = ImportBatchRead.model_validate(response.json())
+        print(import_batch)
+        if import_batch.status == "completed":
```
Adam-Hammo: We could also break here on failed and partially failed

cli/post_import_file.py
```diff
@@ -0,0 +1,205 @@
...
+    """Parse the environment and import file details."""
+    parser = argparse.ArgumentParser(
+        description="Posts a single file to the Destiny API for import."
+    )
+    parser.add_argument(
+        "-e",
+        "--env",
+        type=Environment,
+        default=Environment.LOCAL,
+        required=True,
```
Adam-Hammo: Required and with a default feels like a contradiction

cli/post_import_file.py
Adam-Hammo: (And if I can be awful, we can compare to the enum itself)

Adam-Hammo (approved):
Looks good, non-blocking nits

cli/post_import_file.py
jack-o-the-green: Good catch - added a `get_terminal_statuses` to the model

cli/post_import_file.py
jack-o-the-green: I've removed the sys.exit(1) from this script entirely, it's not really a failure state if the batch just hasn't completed. Now you'll just get a log line if it's not done and then the batch summary regardless.

cli/request_enhancements_for_search.py
Adam-Hammo, line 1: Could we maybe call `request_enhancements_for_search` to safeguard against future sibling enhancement requesters?

Adam-Hammo (approved):
Looks good! One naming nit

PR #751, opened 2026-06-17
Add --created-after flag to request_enhancements_for_search cli tool

Adam-Hammo:
Allows time-based filtering of enhancement requests - for instance, when importing additional records to an existing community.

### DIFF
--- a/cli/request_enhancements_for_search.py
+++ b/cli/request_enhancements_for_search.py
@@ -12,6 +12,7 @@ enhancement-request endpoints are called directly via httpx.
 # ruff: noqa: T201
 import argparse
 import sys
+from datetime import UTC, datetime
 from uuid import UUID
 
 import httpx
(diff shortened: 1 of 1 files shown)


cli/request_enhancements_for_search.py
```diff
@@ -64,20 +65,63 @@ def create_enhancement_request(
...
+        parsed = datetime.fromisoformat(value)
+    except ValueError as exc:
+        msg = f"Invalid ISO-8601 datetime: {value!r}"
+        raise argparse.ArgumentTypeError(msg) from exc
+    return parsed if parsed.tzinfo else parsed.replace(tzinfo=UTC)
+
+
+def _uuid7_created_at(reference_id: UUID) -> datetime:
+    """Extract the creation time embedded in a UUID7's leading 48-bit ms timestamp."""
+    return datetime.fromtimestamp((reference_id.int >> 80) / 1000, tz=UTC)
```
danielbreves, line 94: Can we use [UUID.time](https://docs.python.org/3/library/uuid.html#uuid.UUID.time) here? Otherwise a comment could be helpful to explain how this works for us mere mortals 😅

cli/request_enhancements_for_search.py
```diff
@@ -166,8 +228,12 @@ def argument_parser() -> argparse.ArgumentParser:
             robot_id=args.robot_id,
             source=args.source,
             exclude_reference_ids=set(args.exclude_reference_ids),
+            created_after=args.created_after,
             dry_run=args.dry_run,
         )
+    except ValueError as exc:
```
danielbreves, line 234: Could we raise something more specific to let other Errors bubble up?

danielbreves (approved):
A couple of minor comments, looks good otherwise!

cli/request_enhancements_for_search.py
Adam-Hammo: Of course! 🤦

Adam-Hammo:
I'm actually going to hold off merging this @danielbreves - realised that there's a quiet failure case where if a record gets deduplicated against an older UUID7 canonical, it will silently be skipped. This feels hard to guard against safely in the CLI without exploding the amount of code or complicating the arg docs.

Still, was handy for running the new HPV enhancements and perhaps is useful to be aware of!

PR #757, opened 2026-06-19
Handle large file streaming

Adam-Hammo:
Streaming over chunk boundaries could cause an issue where a multi-byte character is undecodable - eg https://ui.honeycomb.io/destiny-evidence/environments/production/result/reZXzLiksw6/trace/GM3YuSHu5Z3?fields[]=s_name&fields[]=s_serviceName&span=ae5fb029d3eaa1b6

### DIFF
--- a/app/persistence/blob/client.py
+++ b/app/persistence/blob/client.py
@@ -1,5 +1,6 @@
 """Generic class for a blob storage client."""
 
+import codecs
 from abc import ABC, abstractmethod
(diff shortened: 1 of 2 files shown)


danielbreves (approved):

Adam-Hammo:
Won't merge until next week - have mitigated any impact with the current job by reducing batch (and hence file) size. Should be below the 4MB chunk boundary now.

PR #767, opened 2026-06-30
Add ris search export functionality

### LINKED ISSUE #764: Add RIS export option
Add the option to export RIS files from the repository, akin to the jsonl export.

Adam-Hammo:
Closes #764
- Adds `RIS` projection and rendering
- Adds `export_format` option to search export process
- Also includes some minio config that makes it possible to access signed URLs outside the docker network for local testing :)

Notes:
- The RIS format is not well documented. Different databases support it in different ways. I've tried to strike a relatively verbose best-efforts middle ground but we may find we'll need to tweak over time from user requests.
- Authorship is passed straight through, our stored format is inconsistent by import source. For better or worse RIS doesn't specify a format here anyway.
- Abstract is omitted by default.
- `Venue → RIS type` is a best-effort map. Eg repositories/preprints map to `GEN` rather than optimistically claiming `JOUR`.
  - We can get a lot better here (and likely in other fields) by looking at coded fields. I think this is beyond the current scope.

Just by nature of bibliographic density, records with OpenAlex bibliographies will present more nicely.

Sample export:
(extension changed so github doesn't think i'm tryna hack y'all)
[evidence-repository-references-hpv-20260630.ris.txt](https://github.com/user-attachments/files/29491208/evidence-repository-references-hpv-20260630.ris.txt)


app/domain/references/routes.py
jaybea: Maybe change to "produced export file" (How's that for a nitpick!)

app/persistence/blob/clients/minio.py
jaybea, line 154: url might not get set above?

app/domain/references/models/projections.py
```diff
@@ -241,6 +246,159 @@ def get_canonical_candidate_search_fields(
...
+                or (venue.host_organization_name if venue else None),
+                issns=venue.issn if venue and venue.issn else [],
+                abstract=abstract,
+                doi=cls._identifier_value(
+                    reference, destiny_sdk.identifiers.ExternalIdentifierType.DOI
+                ),
+                accession=str(reference.id),
+                database=cls.DATABASE_NAME,
+                pdf_url=pdf_url,
+                urls=list(set(urls + cls._identifier_urls(reference))),
```
jaybea: I think urls=list(dict.fromkeys(...)) would give a deterministic order here, if we think that would be useful going forward.

jaybea (approved):
This looks good. Sorting out richer bilbio upstream is another effort.

A couple of optional nits inline for you to consider.

app/domain/references/routes.py
Adam-Hammo: Good spot! Correct docs are never too nitpicky :)

app/domain/references/models/projections.py
Adam-Hammo: Ah, python...

PR #774, opened 2026-07-02
Extend README for easier reproducibility

TimRepke:
I tried to get the repository running locally and kept a log of issues I encountered. In hindsight, not too much, but recovering from not setting environment variables for example took a lot of time, hence the README should be more explicit about that.

I also added an option to configure the postgres port because I have multiple instances running.

Unresolved is the "issue" that I need sudo for `uv run fastapi` and `uv run taskiq`, otherwise I get permission errors (apparently the certs are placed in the docker volume directory and some of the volumes are owned by root (not running `docker-compose up` as root)

```
17:10 start
17:18 docker compose done but not starting; >2.5GB storage gone
17:24: needs to `sudo pacman -S docker-buildx`
17:26: error: port 5432 already used
17:34: fixed with added config
17:43: .env template can't be sourced because of tenant ID (replace first)
17:47: need to use DB_conf option with connection string in .env to run alembic migrate
17:49: uv run fastapi seems to work, but not reachable
17:51: taskiq worker fails to run, ES permission denied
17:57: run docker compose down -v --rmi all to clear everything 
17:58: docker compose again with source .env before (eats another 2GB of storage)
17:59: problem persists
18:40: deleted volumes several time to try and flush (possibly) stuck config and wrong password; added override to have volumes in different location
18:44: can now log in via browser, but fastapi now fails with permission denied and taskiq still has permission errors
18:55: some volume directories were now owned by root; changing ownerships, now fastapi runs again (still no opening in browser though)
18:58: changing permissions of volume dir is no good for container apparently and resets are down+up; running fastapi as root works and now also access in browser
18:59: taskiq as root also fixes issues and things work



root causes:
* not enough disk space for volumes causes odd ES behaviour
* not sourcing .env before first run causes stubbornly persistent configs in docker
* file permissions; shouldn't need to run uv as root for fastapi and taskiq

docker system prune -a --volumes helps to clean things up
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,8 @@
 ### DB ###
+POSTGRES_PORT=5432
+
 # Locally (option 1):
-DB_CONFIG='{"DB_FQDN": "localhost", "DB_NAME": "destiny_dev", "DB_USER": "localuser", "DB_PASS": "localpass"}'
+DB_CONFIG='{"DB_FQDN": "localhost", "DB_NAME": "destiny_dev", "DB_USER": "localuser", "DB_PASS": "localpass", "DB_PORT": 5432}'
 
 # Locally (option 2):
 # DB_CONFIG='{"DB_URL": "postgresql+asyncpg://localuser:localpass@localhost:5432/destiny_dev"}'
--- a/README.md
+++ b/README.md
@@ -35,21 +35,46 @@ modification to run the application), copy the example file:
 
 ```shell
 cp .env.example .env
+
+# To run any of the following commands, environment variables need to be set by
+source .env
 ```
+If you want to run locally and have no tenant-id, you at least need to replace the 
+`<` and `>` in the `.env` file.
 
 ### Starting the development server
 
 First you will need to start the auxiliary servers:
 
 ```sh
+source .env  # set environment variables
 docker compose up -d
 ```
+
+Make sure there is enough space where the docker volumes are located (typically 
+`/var/lib/docker/`); elasticsearch will cause issues if the partition is >90%. 
+Plan for 2–3GB for images and volumes.
+
+If you want to locate the volumes in a specific location, you can create 
+`docker-compose.override.yml`:
+```
+volumes:
+  certs:
+    driver: local
+    driver_opts:
+      type: none
+      o: bind
+      device: /path/to/volumes/certs
+  ...
+```
+Remember to create the directory before starting.
 
 #### Database
 
 Once the database server is running, run the migrations to setup the database.
 
 ```sh
+source .env  # set environment variables
 uv run alembic upgrade head
 ```
 
--- a/README.md
+++ b/README.md
@@ -33,13 +33,14 @@ accidental exposure of secrets, this file is ignored using the `.gitignore`
 file. To set up a copy of the configuration (which should not require
 modification to run the application), copy the example file:
 
-```shell
+```sh
 cp .env.example .env
 
 # To run any of the following commands, environment variables need to be set by
 source .env
 ```
-If you want to run locally and have no tenant-id, you at least need to replace the 
+
+If you want to run locally and have no tenant-id, you at least need to replace the
 `<` and `>` in the `.env` file.
 
 ### Starting the development server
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -5,8 +5,9 @@ services:
       POSTGRES_USER: localuser
       POSTGRES_PASSWORD: localpass
       POSTGRES_DB: destiny_dev
+      POSTGRES_PORT: 5432
```
danielbreves: It doesn't look like the [postgres image](https://hub.docker.com/_/postgres/#environment-variables) uses this?

.env.example
```diff
@@ -1,6 +1,8 @@
 ### DB ###
+POSTGRES_PORT=5432
+
 # Locally (option 1):
-DB_CONFIG='{"DB_FQDN": "localhost", "DB_NAME": "destiny_dev", "DB_USER": "localuser", "DB_PASS": "localpass"}'
+DB_CONFIG='{"DB_FQDN": "localhost", "DB_NAME": "destiny_dev", "DB_USER": "localuser", "DB_PASS": "localpass", "DB_PORT": 5432}'
```
danielbreves, line 5: Our DB config doesn't use [DB_PORT](https://github.com/destiny-evidence/destiny-repository/blob/main/app/core/config.py#L29) at the moment.

README.md
```diff
@@ -35,21 +35,46 @@ modification to run the application), copy the example file:
 
 ```shell
 cp .env.example .env
+
+# To run any of the following commands, environment variables need to be set by
+source .env
```
danielbreves, line 40: Both [app/core/config.py:396](https://github.com/destiny-evidence/destiny-repository/blob/main/app/core/config.py#L396) and [app/migrations/config.py:23](https://github.com/destiny-evidence/destiny-repository/blob/main/app/migrations/config.py#L23) load `.env`, so I'm not sure why you would need this.

README.md
```diff
@@ -35,21 +35,46 @@ modification to run the application), copy the example file:
...
 ```
+If you want to run locally and have no tenant-id, you at least need to replace the 
+`<` and `>` in the `.env` file.
```
danielbreves, line 44: I think this is only a problem when you `source .env`, which I'm not sure why it was needed.

docker-compose.yml
danielbreves, line 252: I wonder if this may enable you to load it without sudo?
```suggestion
        # CA is public; make the host copy readable so local (non-docker)
        # uv run fastapi/taskiq can load it without sudo
        chmod 644 /certs-local/ca.crt;
        echo "Waiting for Elasticsearch availability";
```

docker-compose.yml
TimRepke: This is about exposing to localhost and I added this to the ports section (which works)

README.md
TimRepke: This is what I assumed at first and maybe is was a fluke, but after sourcing `.env` some more things started working. It certainly doesn't hurt and is a useful reminder to do for all the things that *do* expect env vars to be set.

README.md
TimRepke: Exactly, only relevant in that case. People not so familiar with how sourcing works might stumble across an error that blocks them for a while. Doesn't hurt to put in a note and be more explicit.

.env.example
TimRepke: Added

docker-compose.yml
danielbreves: I think `ports` works because you have `POSTGRES_PORT` in your .env, no? `environment` is only for the image.

TimRepke: My bad, removed. I thought this defines defaults and provides an accessible overview of available configuration options.

docker-compose.yml
TimRepke: That seems to fix it. Maybe I misinterpreted the error's root cause.

app/core/config.py
```diff
@@ -48,9 +49,19 @@ def connection_string(self) -> str:
...
+                f"{self.db_user}"
+                f"@{self.db_fqdn}:{self.db_port}"
+                f"/{self.db_name}"
+            )
         else:
-            url = f"postgresql+asyncpg://{self.db_user}:{self.db_pass}@{self.db_fqdn}/{self.db_name}"
+            url = (
+                f"postgresql+asyncpg://"
+                f"{self.db_user}:{self.db_pass}"
+                f"@{self.db_fqdn}:{self.db_port}"
```
danielbreves, line 62: This would make `self.db_port` required, so it would break production if we don't also set it in infra. I think it would be safer to keep it optional and only add it if set. Then you shouldn't need to change the validation.

```
host = f"{self.db_fqdn}:{self.db_port}" if self.db_port else self.db_fqdn
```

README.md
danielbreves: I think this may be more misleading than helpful tbh. It only mutates your current shell and may not actually reach the child `uv run` processes.

PR #775, opened 2026-07-03
Deterministic sort in Elasticsearch

### LINKED ISSUE #770: Search pagination is non-deterministic (no stable sort tiebreaker)
## Problem

Search pagination is **not stable across requests**: fetching the same page number twice can return different references / a different order. Confirmed against the real backend — reloading page 1 of an HPV search shows a different set of papers each time.

This makes any workflow that spans pages unreliable — most visibly the new cross-page reference selection in evidence-repo-ui ([#197](https://github.com/destiny-evidence/evidence-repo-ui/issues/197)): after "select all" then deselecting a page, revisiting that page shows it selected again, because the deselected reference ids are no longer on the page. Paginated reads and exports have the same latent exposure.

## Root cause

`ESReferenceRepository.search_with_query_string` (`app/persistence/es/repository.py`) applies only the caller's `sort`:

```python
search = AsyncSearch(...).extra(size=page_size).extra(from_=(page - 1) * page_size).query(...)
if sort:
    search = search.sort(*sort)
```

- The default search (relevance / `Sort: Relevance`) passes **no `sort`**, so results are ordered by `_score` alone. Documents with tied scores have **no tiebreaker**, and Elasticsearch does not guarantee a stable order for ties across requests/shards → page boundaries shift between requests.
- Even an explicit sort (e.g. `publication_year`) is non-unique, so ties within it are equally unstable.
- There is no deterministic final sort key. The reference id is only the ES `_id` (`ReferenceDocument` stores it via `meta={"id": …}`), and `_id` isn't sortable by default (requires fielddata, which ES disallows). `ReferenceSearchFieldsMixin` has no unique sortable field either.

## Proposed fix

Always append a **deterministic tiebreaker** on a unique, indexed field as the final sort key, so total ordering is stable across requests:

1. Add an indexed `Keyword` field to `ReferenceDocument` mirroring the id (e.g. `id`/`doc_id`), populated from `domain_obj.id` in `from_domain` (reindex required).
2. In `search_with_query_string`, append that field (asc) to the sort list for every search — after `_score` for relevance, and after any user-supplied sort keys.

This keeps the existing `from`/`size` pagination. (If deep, cursor-stable pagination is wanted later, `search_after` + PIT on the same tiebreaker is the follow-on.)

## Acceptance criteria

- [ ] Requesting the same page of the same search twice returns the same references in the same order (relevance sort and explicit sorts).
- [ ] The tiebreaker is applied for all search paths (results, id-only search, export) so selection/export over multiple pages is consistent.
- [ ] Integration test: two identical paged searches (with score ties) return identical, non-overlapping pages.
- [ ] Reindex/migration handled for the new sortable id field.

Adam-Hammo:
Closes #770 
- Adds `id` as a first-class document field in `Reference`
- Appends `id` to the end of every sort as a stable tiebreaker
- Adds ability to run a [`painless`](https://www.elastic.co/docs/explore-analyze/scripting/modules-scripting-painless) script in migration, allowing the required migration to run entirely cloud-side

### DIFF
--- a/app/persistence/es/index_manager.py
+++ b/app/persistence/es/index_manager.py
@@ -45,6 +45,7 @@ class IndexManager:
         repair_subset_task: AsyncTaskiqDecoratedTask[..., Coroutine[Any, Any, None]]
         | None = None,
         reindex_status_polling_interval: int = 5 * 60,  # default to 5min
+        reindex_script: dict[str, Any] | None = None,
     ) -> None:
         """
         Initialize the migration manager.
@@ -58,6 +59,7 @@ class IndexManager:
                 in which case subset repair is unsupported for this index.
             version_prefix: Prefix for version numbers in index names
             reindex_status_polling_interval: How often to check status of reindexing (defaults 5s)
+            reindex_script: Painless script applied to each doc during reindex (defaults None)
 
         """  # noqa: E501
         self.document_class = document_class
@@ -65,6 +67,7 @@ class IndexManager:
         self.repair_task = repair_task
         self.repair_subset_task = repair_subset_task
         self.reindex_status_polling_interval = reindex_status_polling_interval
+        self.reindex_script = reindex_script
 
         self.alias_name = document_class.Index.name
         self.otel_enabled = otel_enabled
@@ -362,6 +365,7 @@ class IndexManager:
             conflicts="proceed",
             source={"index": source_index},
             dest={"index": dest_index, "version_type": "external"},
+            script=self.reindex_script,
             wait_for_completion=False,
             refresh=True,
         )
--- a/app/persistence/es/index_manager.py
+++ b/app/persistence/es/index_manager.py
@@ -11,7 +11,7 @@ from elasticsearch.dsl import AsyncDocument, AsyncIndex
 from opentelemetry import trace
 from taskiq import AsyncTaskiqDecoratedTask
 
-from app.core.exceptions import NotFoundError
+from app.core.exceptions import ESError, NotFoundError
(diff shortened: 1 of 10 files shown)


app/persistence/es/index_manager.py
```diff
@@ -58,13 +59,15 @@ def __init__(  # noqa: PLR0913
                 in which case subset repair is unsupported for this index.
             version_prefix: Prefix for version numbers in index names
             reindex_status_polling_interval: How often to check status of reindexing (defaults 5s)
+            reindex_script: Painless script applied to each doc during reindex (defaults None)
```
jack-o-the-green: Add link to Painless here - I feel this reads too like just general English and not like it's a scripting language.

tests/integration/test_es_index_manager.py
```diff
@@ -540,6 +540,40 @@ async def test_we_do_not_roll_back_to_nonexistent_index(index_manager: IndexMana
         await index_manager.rollback()
 
 
+async def test_migrate_applies_reindex_script(
```
jack-o-the-green, line 543: Can we add a test in here for what happens if an invalid painless script is provided. Also if there's a way we can check syntax before triggering the migration that would be great. The parameters get added to the command line on the container app job so I think the potential for human error is slightly higher than usual.

Adam-Hammo:
Good shouts @jack-o-the-green, have updated the docs and added a pre-migration compile step

jack-o-the-green (approved):
LGTM - thanks for adding the script validation step!

PR #792, opened 2026-07-07
Export by reference ids

### LINKED ISSUE #768: Export references by ID list
Backend support for destiny-evidence/evidence-repo-ui#186 ("select references to AI-summarise + export"). No UI; fully independent of the frontend issues and can start immediately.

> As a user exporting my results, I want the export to honour an explicit list of references I've chosen, not just my search criteria.

Today the search-export endpoint takes search criteria (query + filters) and exports everything that matches. To export a hand-picked selection, the API needs to accept the explicit set.

## Acceptance criteria

- [ ] The export API accepts a list of reference IDs and returns an export containing exactly those references.
- [ ] Requests whose reference list exceeds the 10,000 export limit are rejected with a clear, structured error.
- [ ] The existing criteria-based export continues to work unchanged.

Adam-Hammo:
Closes #768 
- Generalises export flow
- Adds reference id export task, models and API endpoint

### DIFF
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -493,6 +493,14 @@ class Settings(BaseSettings):
         ),
     )
 
+    max_reference_export_size: int = Field(
+        default=10000,
+        description=(
+            "Maximum number of reference IDs allowed in a single reference export "
+            "request. Requests exceeding this are rejected."
+        ),
+    )
+
     default_es_indexing_chunk_size: int = Field(
         default=1000,
         description=(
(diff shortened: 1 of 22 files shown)


jaybea (approved):
Wow, nice DRY refactor, roots deep in the repository, service, routing. Looks very good. I'm going to use this for researching presence of references in production maybe (once deployed)!

A few of nits:

- n_references (the count surfaced in the export status) is len(reference_ids) at service.py:228, the raw request-list length, but it's documented as "the number of references included in the produced file". The file is built by _get_deduplicated_references, which writes one line per distinct id, so a literal repeated id in the request makes the reported count overstate the file: [A, A, B] reports n_references: 3 but the file has 2 lines. (Boy that's nit-picky!)

- reference_ids seems to be documented "in the order provided", but the `get_by_pks` doesn't seem to have an ORDER BY, so the output is in DB order. Looks like this is done elsewhere though.

- I couldn't find a test of request count exceeding specified limit.

PR #810, opened 2026-07-09
Scream test

Adam-Hammo:
Non-destructive disabling of Azure auth clients to scream test #645


jaybea (approved):
This ol' checknut is still out there, but I expect you to just modify it if is ever needed again: [EEF ad-hoc deduplication](https://github.com/destiny-evidence/destiny-repository/blob/main/libs/ad_hoc/scripts/deduplicate_eef_references/deduplicate_eef_references.py#L181)

Of course this blocks new tokens, so we need to wait till current ones are expired and people are reaching for access again to REALLY get a scream test (preachin' to the choir here, I'm sure.)

Nice how this is reversible, non-destructive.

PR #817, opened 2026-07-09
Demojibake outgoing enhancements

### LINKED ISSUE #781: Remap mojibake in linked-data display text (presentation-layer)
## Background

#697's latin-1 default mojibake'd multi-byte UTF-8 in the EEF source data. The Vocabulary Mapping Robot consumed the mojibake'd input and carried the bytes into the free text of its linked-data enhancements. #701 remediates the searchable text (title, authorship, abstract); this issue covers the remaining user-visible mojibake in linked-data display text.

This complements #701. Together the two targeted fixes deliver clean search and clean display without a full corpus reimport, which we'd like to avoid. #701 on its own makes affected records findable but leaves their linked-data display text corrupted; this issue on its own cleans the display but does not change search. Implement both and users get findable and cleanly readable records from targeted remaps rather than a re-ingest.

## This is presentation remap only, not persisted repair or a robot re-run

Verified locally:

- Default user search is title and abstract only (`default_search_fields`).
- The only LDE-derived ES field, `linked_data_labels`, is built by `LinkedDataProjectionService` from vocabulary SKOS labels resolved by concept URI, not copied from the LDE payload's free text. So mojibake in LDE free text never reaches the index.
- Local LDE mojibake is confined to human-readable free-text fields: `description` (~4,900 LDEs), `supportingText` (~3,900), `name` (~780), and `@value` literals (~230). None of these are concept labels.

The corruption is therefore display-only. A byte-level persisted repair or a Vocabulary Mapping Robot re-run is not needed to fix search, which is why this is scoped as a presentation-layer fix.

## Approach

- Add a shared display/export normaliser that applies the #697 reversible remap (`s.encode("latin-1").decode("utf-8")`) to detected mojibake. Use the same detector as #701: a lead char in U+00C2 to U+00F4 followed by a continuation in U+0080 to U+00BF, plus C1 controls, confirmed per string by a successful round-trip remap.
- Apply it only to human-readable strings surfaced from the hydrated LDE payload (`supportingText`, `name`, `description`, `@value`), never to URIs, `@id`, codes, or concept identifiers.
- Apply it at the destiny-repository API response-serialization seam that surfaces linked-data display text, at the shared point every LDE-surfacing endpoint passes through so no response route bypasses it. (RIS export is bibliographic projection only and carries no linked-data text, so it is not a seam here.) Consumers render and export whatever the API returns, so normalizing here fixes them all at once, including evidence-repo-ui's display, CSV export, and Excel export; no per-surface or per-repo change is needed. Keep the normaliser out of write paths.

## Caveat

If the published vocabulary itself ever carries mojibake in its SKOS labels, that would surface in `linked_data_labels` and display remap would not fix `linked_data_labels:` fielded searches. That is a separate upstream concern, not this issue; local labels resolve from the clean vocabulary.

## Scope (local dev DB with ESSA and EEF as a sample)

Linked-data enhancements with mojibake in display free text, using the broadened detector (~5,480 of 7,240 LDE enhancements affected):

| LDE field | Affected LDEs (approx) |
|---|---|
| `description` | ~4,870 |
| `supportingText` | ~3,870 |
| `name` | ~780 |
| `@value` | ~230 |

Fields overlap (one LDE can carry mojibake in several), so the column does not sum to the total.

## Dependencies

Independent of #701: different fields, different layer. Can ship with or without #701, and in either order.

## Out of scope

- Persisted linked-data repair or a Vocabulary Mapping Robot re-run: a separate future issue if ever justified (broader risk, including ontology validation and changing projected labels/concepts).
- title, authorship, abstract: #701.
- Repairing the published vocabulary.

## Acceptance criteria

- A shared normaliser remaps detected mojibake (the #701 signature, confirmed by round-trip remap) in LDE human-readable strings at the destiny-repository API response-serialization seam, covering every LDE-surfacing endpoint, never touches URIs, `@id`, or codes, and is not added to write paths.
- Downstream consumers (evidence-repo-ui display, CSV export, Excel export) render clean LDE text with no per-surface change, because they consume the normalized API.
- No change to `linked_data_labels` or default search behavior.

## Related

- #697: original importer bug (root cause, repair recipe)
- #701: searchable-text remediation (title, authorship, abstract)
- #780: venue/pagination enrichment and backfill

Adam-Hammo:
Closes #781 

Adds a step to the anti-corruption layer that amends the mojibake corruption on specified fields.

(Sidebar: this architecture is glorious - one tiny touch point to change all exports/endpoints/etc (except RIS :())

### DIFF
--- a/app/domain/references/models/ris.py
+++ b/app/domain/references/models/ris.py
@@ -15,7 +15,7 @@ from typing import Annotated
 from pydantic import Field
 
 from app.domain.base import ProjectedBaseModel
-from app.utils.strings import flatten_newlines
+from app.utils.strings import demojibake, flatten_newlines
 
 
 class RisType(StrEnum):
@@ -95,4 +95,4 @@ class RisRecord(ProjectedBaseModel):
     @staticmethod
     def _line(tag: str, value: object) -> str:
         """Format a single ``TAG  - value`` line, flattening internal newlines."""
-        return f"{tag}  - {flatten_newlines(str(value))}"
+        return f"{tag}  - {flatten_newlines(demojibake(str(value)))}"
--- a/app/domain/references/services/anti_corruption_service.py
+++ b/app/domain/references/services/anti_corruption_service.py
@@ -39,6 +39,10 @@ from app.persistence.es.persistence import (
     ESSearchResult,
     ESSearchTotal,
 )
+from app.utils.strings import demojibake, demojibake_walk
+
+# JSON-LD literal keys carrying human-readable free text
+_LINKED_DATA_TEXT_KEYS = frozenset({"name", "description", "supportingText", "@value"})
 
 
 class ReferenceAntiCorruptionService(GenericAntiCorruptionService):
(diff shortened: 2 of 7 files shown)


danielbreves (approved):
Looks good! I wondered whether it could be worth using something like https://ftfy.readthedocs.io/en/latest/index.html, in case there are similar issues that we're not yet aware of (e.g. mixed-script strings).

Adam-Hammo:
@danielbreves I really like the idea of using ftfy but my gut feel is that if we implemented something universal we should consider that at the write side and not the read side. 
As this PR is essentially a band-aid, it feels right to me to keep it targeted. What do you think?

danielbreves:
Implementing it at the write side carries more risk, so we'd probably want to do a deeper investigation to understand any potential for undesirable "fixes". But in any case, I'm very comfortable with the targeted approach 👍

Issue #819, opened 2026-07-09
Candidate retrieval policies: compare query variants after baseline API


jaybea:
Part of #791. Follows the Candidate Selection API MVP.

## Summary

Once the baseline Candidate Selection API exists, add a controlled way to compare named retrieval policies without exposing arbitrary Elasticsearch query parameters. This issue owns the experiment surface for deciding whether the current fuzzy title/author/year query should remain the production candidate policy or be replaced by a measured variant.

The baseline API deliberately exposes the current query as `current_fuzzy_v1`. This follow-on adds named, server-side retrieval policies and evaluates them against labelled positives and realistic incoming OpenAlex strata. The goal is to choose, from data rather than code inertia, the first production retrieval policy (`candidate_v1`).

## Motivation

The current candidate query is useful as a control, but it is not validated. Existing measurements already show two tensions:

- It is expensive: fuzzy title expansion dominates ES candidate-retrieval cost.
- It is not obviously the right shape for the incoming OpenAlex population, where identifier coverage, title quality, author count, and work type vary materially.

At the same time, free-form query parameters would make recall results incomparable. A named retrieval policy fixes the full retrieval regime under one label: query shape, searchability gate, identifier-union behaviour, candidate depth defaults, and any route ordering rules.

## Scope

**In scope**

- Add a small retrieval-policy registry behind the Candidate Selection API.
- Keep `current_fuzzy_v1` as the control policy.
- Add the named policies and query shapes to evaluate, for example:
  - `identifier_union_v1`: DOI, PubMed PMID (`pm_id`), and OpenAlex W-ID union plus the current ES query.
  - `loose_title_query`: the primary recall-oriented query shape, chosen from the OpenAlex profiling and baseline results.
  - `no_fuzziness_query`: an optional cost-versus-recall probe if evaluation needs to quantify the throughput trade-off. It should not become the primary candidate policy unless recall supports that choice.
- Keep policies server-side and versioned; do not expose raw fuzziness, boost, `minimum_should_match`, or analyzer knobs as request parameters.
- Return the selected `retrieval_policy` and preserve the MVP's `index_version` in every response.
- Evaluate policies using the recall harness over a depth sweep that includes `k=100`.
- Report recall@K, candidate-set size, truncation, latency/cost, zero-candidate rate, and returned-score distributions by stratum.
[...]

Issue #791, opened 2026-07-07
Deduplication for safer external imports


jaybea:
## Outcome

Match imported scholarly records against what we already hold, without depending on a trusted external identifier. The target path is: candidate selection puts forward possible matches, including exact identifier matches; the Deduper scores each pair and advises whether it clears the threshold; destiny-repository weighs that advice alongside its other decision inputs, such as an existing manual resolution, then records the merge decision and where it came from. The Deduper is released separately, from destiny-evidence/destiny-deduper. First stream to use it: openalex-incremental-updater ingestion.

## Where this stands

In production, a trusted OpenAlex identifier is the only route that can reach a duplicate decision. The trusted-identifier shortcut records `CANONICAL` or `DUPLICATE` from the identifier alone, without candidate retrieval or scoring: a match makes the incoming reference a duplicate, and no match makes it canonical. The shortcut therefore reads a distinct identifier as a distinct publication. Where OpenAlex itself holds two Works for the same publication, both can be recorded as `CANONICAL`, because nothing else compares them. That is the gap this epic closes.

Candidate selection is built, and so is the assessment path with its recorder. Neither has a production caller, and there is no scorer: the only implementation of the pair-scoring contract is a test fake. The decision path is still a placeholder that returns `DUPLICATE` only in the test environment and `UNSEARCHABLE` everywhere else, whether or not candidate selection ran. No scored merge can happen today under any configuration.

#818, #819 and #822 established the retrieval and evaluation approach. The configured policy is deployed in production with candidate search switched off, so no candidate retrieval runs there yet. The Deduper is released: destiny-deduper#14 is closed and 0.2.6 is published. Two things still sit on top of it, #944, the adapter that calls it, and #887, the evaluation run that uses it.

Deduplication terminology, the decision states and their search-index behaviour are documented in `docs/procedures/deduplication.rst`.

## Evaluation first: is deduplication safe, and is it accurate enough?

Two questions, and they get settled at different times.

**Is it safe?** Both parts can be settled before the Deduper arrives: an evaluation run must have no way to write at all (#874 keeps evaluating and writing as separate operations, so an evaluation never reaches the writer), and an automatic decision must not overwrite one a person made (#582).

**Is it accurate enough** to be trusted with automatic merging? That one needs the Deduper, and the production-like evaluation run in #887 is what answers it.

[...]

PR #829, opened 2026-07-15
Candidate Selection API: evaluable baseline retrieval

### LINKED ISSUE #818: Candidate Selection API: evaluable baseline retrieval
Part of #791.

## Summary

Build a read-only, zero-persistence endpoint that takes a reference and returns ranked candidate canonicals from the same candidate-retrieval path destiny-repository intends to evaluate for deduplication.

The immediate goal is not to tune the Elasticsearch query or choose the final production retrieval policy. The goal is to make the current candidate path callable, inspectable, and measurable against production-like data without relying on staging feature flags or writing duplicate decisions.

This issue implements the baseline API agreed from the candidate-selection discussion:

- reference in, ranked candidates out;
- explicit candidate depth `k`, defaulting to `100`;
- the current Elasticsearch candidate query reused through a shared retrieval method;
- optional exact-identifier candidate union from Postgres;
- hydrated, route-labelled candidates and basic diagnostics for evaluation;
- no persistence and no dedup actuation.

## API Contract

`POST /references/deduplication/candidates/`, requiring the existing `reference.deduplicator` auth scope rather than introducing a new scope.

**Request**

- `input`: exactly one of:
  - `{ "reference_id": "<uuid>" }`, where the server hydrates and projects the stored reference into candidate-search fields; or
  - an inline payload with candidate-search fields:
    `{ "title": str, "authors": [str], "publication_year": int, "identifiers": [ { "identifier_type": str, "identifier": str, "other_identifier_name": str? } ] }`
- `k`: integer, optional, bounded; defaults to `100`.
- `include_identifier_matches`: boolean, optional, defaults to `true`.
- `hydrate`: boolean, optional, defaults to `true`.

The Elasticsearch path is not configurable in this issue. It is the current candidate query: fuzzy title match, author `dis_max`, publication-year window, canonical-only filter, and self-exclusion where an input `reference_id` is provided. The implementation should refactor this into a shared method so this endpoint and the existing nomination path cannot drift.

The `include_identifier_matches` flag is an evaluator convenience for the MVP. The follow-on retrieval-policy work may fold this into named policies instead of keeping it as an open request knob; callers should treat the flag as part of the baseline evaluation surface, not a permanent production policy control.

**Response** (shape, exact names can follow local API conventions)

```json
{
  "retrieval_policy": "current_fuzzy_v1",
  "index_version": "reference_v3",
  "k_requested": 100,
  "include_identifier_matches": true,
  "input_searchability": { "searchable": true, "reason": "ok" },
  "diagnostics": {
    "es_took_ms": 272,
    "es_total_hits": 384,
    "es_returned": 100,
    "identifier_returned": 1,
    "candidate_count": 101,
    "truncated": true,
    "kth_es_score": 8.42,
    "lowest_es_score": 8.42
  },
  "candidates": [
    {
      "reference_id": "...",
      "rank": 1,
      "routes": [
        { "type": "elasticsearch", "policy": "current_fuzzy_v1", "rank": 1, "score": 12.3 },
        { "type": "identifier", "matched_identifiers": [ { "identifier_type": "doi", "identifier": "..." } ] }
      ],
      "reference": {
        "title": "...",
        "authors": ["..."],
        "publication_year": 2024,
        "identifiers": []
      }
    }
  ]
}
```

Contract points:

- `retrieval_policy` labels the baseline query being evaluated. It is informational in this MVP, not a request-time selector.
- `index_version` gives the baseline result a minimum comparability stamp. A later policy/fingerprint mechanism can add more detail, but recall numbers should at least identify which reference index version they were measured against.
- ES score and identifier evidence are not blended into one score. A candidate can have one or more `routes`.
- Identifier matches are unioned with ES results by reference id.
- Identifier-only matches should rank ahead of ES-scored matches for evaluator visibility.
- Candidate hydration comes from Postgres/domain projection, not Elasticsearch `_source`.
- When `hydrate=false`, omit the hydrated `reference` object but preserve ids, route provenance, and diagnostics.
- If the input fails the existing searchability/presence gate, return HTTP 200 with an empty candidate list and a reason rather than treating it as an API error.

## Scope

**In scope**

- New read-only endpoint and request/response models.
- Shared retrieval method for the existing Elasticsearch candidate query, with explicit `k`.
- Default `k=100`, with a bounded override for recall@K sweeps.
- Optional Postgres exact-identifier union for DOI, PubMed PMID (`pm_id`), and OpenAlex W-ID candidates, enabled by default.
- Route provenance for ES and identifier matches.
- Hydrated candidates by default, with `hydrate=false` for bulk sweeps.
- Basic diagnostics needed by evaluation: index version, total ES hits where available, returned count, truncation, returned score floor, and identifier-match count.
- Tests proving the endpoint writes no duplicate-decision state.

**Out of scope**

- Changing the ES query shape, fuzziness, boosts, analyzers, mapping, or searchability/title-quality rules.
- Adding multiple named retrieval variants or a policy registry.
- Pipeline fingerprints or mapping-contract CI tests.
- Deduper scoring, threshold calibration, typed proposals, or any automatic actuation.
- Calling the Deduper to score identifier matches, including any higher-threshold "identifier confirmed" path. This API only returns identifier matches as candidate routes; scoring them belongs to the deduper-scoring integration.
- Identifier short-circuiting that skips ES after scoring an exact identifier match.
- Choosing the final production retrieval policy.

## Evaluation Boundary

This endpoint makes candidate retrieval measurable; it does not itself perform the recall evaluation. The recall@K and query-policy work should use this endpoint against labelled positives and realistic hard negatives, then decide whether the current query is adequate or needs a follow-on retrieval policy.

The default `k=100` is intentional: the old implicit ES default of 10 is not a serious target for production deduplication, and starting evaluations at 100 avoids artificially hamstringing the baseline. Evaluation can still sweep other depths through the explicit `k` parameter.

## Acceptance Criteria

- [ ] `POST /references/deduplication/candidates/` accepts either `reference_id` or an inline payload, and rejects requests that provide both or neither.
- [ ] The endpoint requires the existing `reference.deduplicator` auth scope and writes no duplicate-decision or candidate-persistence state.
- [ ] The current Elasticsearch candidate query is exposed through a shared retrieval method and honours explicit `k`, defaulting to `100`.
- [ ] The response identifies the baseline retrieval policy as `current_fuzzy_v1`.
- [ ] Identifier matches for DOI, PubMed PMID (`pm_id`), and OpenAlex W-ID can be unioned from Postgres and are enabled by default.
- [ ] Candidates include route provenance, with ES scores kept separate from identifier evidence.
- [ ] Hydrated candidate bibliographic fields are returned by default from Postgres/domain projection; `hydrate=false` omits only the hydrated reference payload.
- [ ] Diagnostics include `index_version`, ES returned count, total hits or an equivalent truncation signal where available, candidate count, and returned ES score floor.
- [ ] Unsearchable inputs return HTTP 200 with empty candidates and a searchability reason.
- [ ] Tests cover reference-id input, inline-payload input, identifier-union behaviour, `k` override, and no persistence.

danielbreves:
Part of #791. Closes #818.

Adds a read-only, zero-persistence `POST /references/deduplication/candidates/` endpoint so we can measure candidate-retrieval recall@K against production-like data without writing dedup decisions.

## What's here
- **Shared ES retrieval** — the existing candidate query is refactored into one method taking explicit `k` (fixing the silent ES `size=10` cap), used by both this endpoint and the nomination path. The nomination path's result count is unchanged (production K stays the recall@K evaluation's decision).
- **Endpoint** — accepts either a `reference_id` or an inline payload (title/authors/year/identifiers); guarded by the existing `reference.deduplicator` scope; writes nothing.
- **Identifier union** — optional Postgres exact-identifier matches (DOI, `pm_id`, OpenAlex), unioned with ES results and resolved to canonical. ES score and identifier evidence stay separate as `routes`; identifier-only matches rank ahead.
- **Diagnostics** — `index_version`, ES took/total/returned, candidate count, truncation, and ES score floor.
- Request/response are **local domain models** (not the SDK), since the contract is an evolving evaluation surface (#819 may fold `include_identifier_matches`/`k` into named policies).

## Notes / decisions
- Unsearchable inputs (missing title/authors/year) return `200` with empty candidates + a reason, per the issue.
- The canonical-only ES filter stays coupled to publication-year presence (documented, not changed — out of scope); the searchability gate guarantees a year is present.

## Tests
- 13 unit tests (orchestration, identifier-union ranking, duplicate→canonical resolution, `k`, hydration, unsearchable, validation, no-persistence).
- 4 e2e tests against the real stack (ES ∪ identifier union, hydration, `reference_id` input, unsearchable, validation, and a real no-persistence row-count check).
- Existing unit + integration tests updated for the new signature.

### DIFF
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -330,6 +330,17 @@ class GenericExternalIdentifier(DomainBaseModel):
             else None,
         )
 
+    @classmethod
+    def from_generic(cls, identifier: "GenericExternalIdentifier") -> Self:
+        """
+        Validate and canonicalise a loosely-typed identifier via the SDK models.
+
+        Raises ``pydantic.ValidationError`` if the value is not a valid identifier
+        of its declared type.
+        """
+        specific = ExternalIdentifierAdapter.validate_python(identifier.model_dump())
+        return cls.from_specific(specific)
+
 
 class IdentifierLookup(GenericExternalIdentifier):
     """Model to search for an external identifier."""
@@ -766,28 +777,17 @@ class ReferenceSearchFields(ProjectedBaseModel):
 CURRENT_FUZZY_RETRIEVAL_POLICY = "current_fuzzy_v1"
 
 
-class CandidateIdentifier(BaseModel):
-    """An external identifier, used as request input and as match provenance."""
+class CandidateIdentifier(GenericExternalIdentifier):
+    """
+    An external identifier used as candidate-search input and match provenance.
+
+    Narrows :class:`GenericExternalIdentifier` to require an explicit type;
+    candidate search does not accept database-id (untyped) lookups.
+    """
 
     identifier_type: ExternalIdentifierType = Field(
         description="The type of the identifier, e.g. doi, pm_id, open_alex."
     )
-    identifier: str = Field(description="The identifier value.")
-    other_identifier_name: str | None = Field(
-        default=None,
-        description="The name of the identifier when identifier_type is other.",
-    )
-
-    @classmethod
-    def from_specific(cls, external_identifier: ExternalIdentifier) -> Self:
-        """Build from a stored identifier, casting its value to a string."""
-        return cls(
-            identifier_type=external_identifier.identifier_type,
-            identifier=str(external_identifier.identifier),
-            other_identifier_name=getattr(
-                external_identifier, "other_identifier_name", None
-            ),
-        )
 
 
 class CandidateSelectionInput(BaseModel):
@@ -880,8 +880,8 @@ CandidateRoute = Annotated[
 ]
 
 
-class CandidateReferenceProjection(BaseModel):
-    """Hydrated bibliographic projection of a candidate reference."""
+class CandidateReference(ProjectedBaseModel):
+    """Hydrated bibliographic fields of a candidate reference."""
 
     title: str | None = None
     authors: list[str] = Field(default_factory=list)
@@ -897,7 +897,7 @@ class Candidate(BaseModel):
     routes: list[CandidateRoute] = Field(
         description="How this candidate was surfaced (ES and/or identifier)."
     )
-    reference: CandidateReferenceProjection | None = Field(
+    reference: CandidateReference | None = Field(
         default=None,
         description="Hydrated bibliographic fields; omitted when hydrate is false.",
     )
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -778,6 +778,17 @@ class CandidateIdentifier(BaseModel):
         description="The name of the identifier when identifier_type is other.",
     )
 
+    @classmethod
+    def from_specific(cls, external_identifier: ExternalIdentifier) -> Self:
+        """Build from a stored identifier, casting its value to a string."""
+        return cls(
+            identifier_type=external_identifier.identifier_type,
+            identifier=str(external_identifier.identifier),
+            other_identifier_name=getattr(
+                external_identifier, "other_identifier_name", None
+            ),
+        )
+
 
 class CandidateSelectionInput(BaseModel):
     """
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -330,20 +330,20 @@ class GenericExternalIdentifier(DomainBaseModel):
             else None,
         )
 
+
+class IdentifierLookup(GenericExternalIdentifier):
+    """Model to search for an external identifier."""
+
     @classmethod
-    def from_generic(cls, identifier: "GenericExternalIdentifier") -> Self:
+    def from_generic(cls, identifier: GenericExternalIdentifier) -> Self:
         """
-        Validate and canonicalise a loosely-typed identifier via the SDK models.
+        Build a lookup from a loosely-typed identifier, validating and canonicalising.
 
         Raises ``pydantic.ValidationError`` if the value is not a valid identifier
         of its declared type.
         """
         specific = ExternalIdentifierAdapter.validate_python(identifier.model_dump())
         return cls.from_specific(specific)
-
-
-class IdentifierLookup(GenericExternalIdentifier):
-    """Model to search for an external identifier."""
 
 
 class FullTextEnhancement(DomainBaseModel):
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -904,9 +904,15 @@ class Candidate(BaseModel):
 
 
 class InputSearchability(BaseModel):
-    """Whether the input reference is searchable for candidate retrieval."""
+    """Whether the input meets the Elasticsearch bibliographic searchability gate."""
 
-    searchable: bool = Field(description="Whether candidate retrieval was attempted.")
+    searchable: bool = Field(
+        description=(
+            "Whether the input meets the Elasticsearch searchability gate (title, "
+            "authors, publication year). Exact identifier matching runs regardless, "
+            "so candidates may be present even when this is false."
+        )
+    )
     reason: str = Field(description="Explanation of the searchability decision.")
 
 
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -763,6 +763,170 @@ class ReferenceSearchFields(ProjectedBaseModel):
         return [cls._normalise_string(v) for v in value if v]
 
 
+CURRENT_FUZZY_RETRIEVAL_POLICY = "current_fuzzy_v1"
+
+
+class CandidateIdentifier(BaseModel):
+    """An external identifier, used as request input and as match provenance."""
+
+    identifier_type: ExternalIdentifierType = Field(
+        description="The type of the identifier, e.g. doi, pm_id, open_alex."
+    )
+    identifier: str = Field(description="The identifier value.")
+    other_identifier_name: str | None = Field(
+        default=None,
+        description="The name of the identifier when identifier_type is other.",
+    )
+
+
+class CandidateSelectionInput(BaseModel):
+    """
+    The reference to find candidates for.
+
+    Provide exactly one of a stored ``reference_id`` (projected into search
+    fields server-side) or an inline set of candidate-search fields.
+    """
+
+    reference_id: UUID | None = Field(
+        default=None,
+        description="A stored reference to hydrate and project into search fields.",
+    )
+    title: str | None = Field(default=None, description="The title to search on.")
+    authors: list[str] = Field(
+        default_factory=list, description="The authors to search on."
+    )
+    publication_year: int | None = Field(
+        default=None, description="The publication year to search on."
+    )
+    identifiers: list[CandidateIdentifier] = Field(
+        default_factory=list,
+        description="Identifiers to union with the Elasticsearch candidates.",
+    )
+
+    @model_validator(mode="after")
+    def validate_exactly_one_input_mode(self) -> Self:
+        """Require exactly one of reference_id or inline candidate-search fields."""
+        has_reference_id = self.reference_id is not None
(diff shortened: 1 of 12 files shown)


jaybea (approved):
This is really solid, I put it through its paces (at least on a small scale). Put the ASySD SRSR 36k refs into my local repository (along with EEF and ESSA already there) and hit with 16k candidate retrievals. It found over 99.6%+.

I didn't expect it to be well over 1000 lines though! But I couldn't find anything excess, lots of picky modifications to existing code, lots of unit and some e2e tests, and a big refactor, which is great.

It also provided all the correct candidates for the EEF arm-links, except for one that used ProQuest-only identifiers where title/author/year did not line up.

A little early indicators that we should widen our year range, but that was expected. 

Can't wait to see how it does on against 470M!

A few things to consider:

1. `track_total_hits=True` is now imposed on the production nomination path, where it really wasn't before. Previously with this off, it didn't worry about doing a full count if it exceeds 10k. But this makes ES find the actual total number answering the query. And it doesn't look like the value is every used, we only return `search_result.hits`. Maybe make it a parameter with default off.

2. I couldn't find a test for when an invalid identifier (like  {"identifier_type": "doi", "identifier": "not-a-doi"} ) fails validation in ExternalIdentifierAdapter, which raises and returns a 422. The unit `TestCandidateSelectionInputValidation` covers only neither/both/k-bounds, but not an invalid-identifier.

app/domain/references/routes.py
```diff
@@ -1445,3 +1452,41 @@ async def make_duplicate_decisions(
 
 
 reference_router.include_router(deduplication_router)
+
+
+@candidate_selection_router.post(
```
Adam-Hammo, line 1457: Is this a `POST` in order to use a query body? Feels itchy but I get it

Relevant but not useful to us yet: https://dev.to/tykok/the-new-http-method-query-2bec

app/domain/references/models/models.py
```diff
@@ -763,6 +763,187 @@ def normalise_string_list_validator(cls, value: list[str]) -> list[str]:
         return [cls._normalise_string(v) for v in value if v]
 
 
+CURRENT_FUZZY_RETRIEVAL_POLICY = "current_fuzzy_v1"
+
+
+class CandidateIdentifier(BaseModel):
```
Adam-Hammo: I think this is identical to `GenericExternalIdentifier` which already exists in this file - why do we redefine it here?

app/domain/references/service.py
```diff
@@ -530,6 +572,230 @@ async def get_references_from_identifiers(
...
+        self, request: CandidateSelectionRequest
+    ) -> CandidateSelectionResult:
+        """
+        Retrieve ranked candidate canonicals for a reference, without persisting.
+
+        Read-only evaluation surface for deduplication candidate retrieval: it runs
+        the shared Elasticsearch candidate query and, optionally, unions exact
+        identifier matches from Postgres, returning route provenance and retrieval
+        diagnostics. It writes no duplicate-decision or candidate state.
+        """
```
Adam-Hammo, line 547: It feels a little to me that some of this could possibly live in `deduplication_service`?

Adam-Hammo: And similar for other methods here - just makes this file seem like less of a behemoth IMO

app/domain/references/service.py
```diff
@@ -530,6 +572,230 @@ async def get_references_from_identifiers(
...
+        fields = ReferenceSearchFieldsProjection.get_from_reference(reference)
+        return CandidateReferenceProjection(
+            title=fields.title,
+            authors=fields.authors,
+            publication_year=fields.publication_year,
+            identifiers=[
+                CandidateIdentifier.from_specific(linked.identifier)
+                for linked in (reference.identifiers or [])
+            ],
+        )
```
Adam-Hammo: A nitpick - this feels like a slight departure from the existing projection pattern that might proliferate if the projection becomes more complex

app/domain/references/service.py
```diff
@@ -530,6 +572,230 @@ async def get_references_from_identifiers(
...
+    ) -> IdentifierLookup:
+        """Normalise an input identifier into a lookup, canonicalising its value."""
+        try:
+            specific = ExternalIdentifierAdapter.validate_python(
+                identifier.model_dump()
+            )
+        except ValueError as exc:
+            msg = f"Invalid identifier for candidate selection: {identifier.identifier}"
+            raise DeduplicationValueError(msg) from exc
+        return IdentifierLookup.from_specific(specific)
```
Adam-Hammo: Would this be nicer as an `IdentifierLookup.from_generic(identifier: GenericExternalIdentifier)`?

app/domain/references/routes.py
danielbreves: Yea 😕

Adam-Hammo: Elasticsearch shamelessly uses `GET` with bodies. I wonder where they get the nerve

app/domain/references/models/models.py
danielbreves: Good catch. One tiny difference is that the `identifier_type` is required, but it can probably subclass.

app/domain/references/routes.py
danielbreves: Looks like ES [_search](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-search) supports both GET and POST, similar with GraphQL queries, but it says supporting methods other than POST is at the discretion of the GraphQL server. And when supporting GET the query [should be in the query string](https://graphql.org/learn/serving-over-http/#get-request-and-parameters).

Adam-Hammo (approved):

jaybea:
I need to reverse myself on the track_total_hits=true. I reran queries in ES staging with it turned on, and there's effectively no difference on latency (I guess it's just bit-counting) and it gives real insights about the actual candidate set size. This will tell us whether the query is worthless (returns count of millions of hits, like I'm seeing against staging), or is closer to our k of 100, or 150...

danielbreves:
@jaybea that's good to know! The candidate selection API will return this, but I don't think we need to change the nomination path, given it's not being used at the moment.

jaybea:
Hey @danielbreves, I can't do actual retrieval-recall against staging because it only has about 1% of the eval dataset. But I wanted to get a feel the mechamisms for how I might have to throttle this in production, so I stress tested staging. Big surprise for me was (on staging) postgres hydration appears more significant for tail latency than ES search!

Looking forward to this being deployed to production, hope this gives some confidance.

I hit the API end to end with inline OpenAlex eval dataset v1 candidate inputs, with `k=100` and hydration enabled. This does exercise the full query and hydration path, but is not a recall measurement against staging.

- Sustained `concurrency=4` for 210s: **1,101/1,101 HTTP 200s**, **5.23 QPS**, wall-clock p95 **1.19s**, ES p95 **471ms**.
- Short ramps remained error-free through `c=20`, but the practical knee was about **13 QPS**: at `c=20`, throughput was 13.17 QPS while wall-clock p95 rose to 2.38s and ES p95 to 1.46s.

On staging, hydration appears to be the larger tail-latency concern than the ES candidate lookup: one representative trace spent 652ms in ES search and 2.63s in `Get Reference.hydrated`. A small `k=100` A/B also had p95 3.69s with hydration versus 1.02s without it. This is staging-only evidence; production ES will search a much larger corpus and needs its own research.

I thought this was good evdience that the API works under sustained staging traffic and degrades with latency rather than errors. In fact there wasn't a single error returned.

If I COULD run the recall eval against staging, I'd stick to concurrency = 4 (getting about 5 QPS), and only push up to c=10-12 for monitored short batch type work.

danielbreves:
@jaybea great stress testing. It looks like `get_hydrated` eager-loads enhancements and identifiers with two joinedloads. Perhaps switching those to `selectinload` could help?

PR #835, opened 2026-07-20
feat(dedup): named candidate-retrieval policies for candidate selection

jaybea:
## Summary

Adds a named, server-side candidate-retrieval policy registry to the candidate-selection API (`POST /v1/references/deduplication/candidates/`, introduced in #818). A request selects a policy by name, and the policy fixes the policy-controlled retrieval axes (year strategy, identifier union, and input searchability) under one stable, versioned name, so recall@K stays comparable across evaluation runs. Shared query shape and scoring config are unchanged, and k stays request-scoped. New policies are added only through a focused experiment cycle; an existing name's semantics never change.

## What ships

Three policies, registered in a read-only mapping:

- `current_fuzzy_v1` (default, control): reproduces the existing baseline exactly. Hard ±1 year window, identifier union on, requires title, authors, and publication year.
- `no_year_filter_v1`: drops the year filter on the query side but still requires a publication year as input.
- `no_year_filter_year_optional_v1`: drops the year filter and makes the publication year optional as input, so references with no year become searchable.

The `include_identifier_matches` request toggle is folded into the policy as `union_identifiers` and removed from the request and result. This breaks the request contract, but the field is not in the SDK or consumed outside internal evaluation, and the endpoint already carried the pre-existing `@experimental` marker (documented as subject to change) from #818; a test pins that a client still sending the field now gets a 422. The nomination path is untouched.

## Implementation notes

- Policies are immutable (`frozen=True`) and exposed through a `MappingProxyType`, so a name's semantics cannot be mutated at runtime. An unknown policy name raises `DeduplicationValueError`, which the route surfaces as HTTP 422 listing the valid names.
- Input searchability is policy-relative and split from the query-side year strategy. The control policy's year-presence check uses `bool(publication_year)`, so a `0` (or JSON `false`, which pydantic coerces to `0`) counts as absent, keeping the control exactly equivalent to the pre-existing baseline gate `all((publication_year, authors, title))` rather than running an ES query the baseline never ran.
- The two axes are independent: `year_strategy` governs the query (hard window vs no filter), while `requires_publication_year` governs the input gate. That is why `no_year_filter_v1` can drop the filter yet still require a year as input, while `no_year_filter_year_optional_v1` relaxes both.

## Validation performed

- Unit, integration, and end-to-end suites green for the candidate-selection area (52 tests across the changed files); mypy and pre-commit clean.
- Live-validated against a local DR stack (host-native app, Elasticsearch `reference_v1` at 28,299 docs, Postgres at 30,195 references). Eight cases over the live endpoint:
  - Control policy with `publication_year` of `0` and of JSON `false` both returned `searchable: false` and ran no Elasticsearch query, matching the baseline gate.
  - `no_year_filter_year_optional_v1` admitted a no-year input (searchable, query ran); `no_year_filter_v1` left the same input unsearchable.
  - An unknown policy returned 422 listing the valid policies.
  - A control pair with identical title and authors and a deliberately far year (1990 against a 2006 reference) flipped the target from excluded under the ±1 window to included once the year filter was dropped, confirming the year strategy actually drives the query rather than being ignored.

Part of #819. This delivers the policy mechanism and the API surface; the issue's downstream evaluation deliverables (depth sweep, stratified recall@K report, production-policy recommendation) land separately, so this does not close it.

### DIFF
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -774,7 +774,23 @@ class ReferenceSearchFields(ProjectedBaseModel):
         return [cls._normalise_string(v) for v in value if v]
 
 
-CURRENT_FUZZY_RETRIEVAL_POLICY = "current_fuzzy_v1"
+class RetrievalPolicyName(StrEnum):
+    """
+    Named, versioned candidate-retrieval policies for the Candidate Selection API.
+
+    Each value fixes a full retrieval regime (query shape, year strategy,
+    identifier union) under a stable name so recall@K stays comparable across
+    runs. A name's semantics never change; a new regime gets a new name.
+    """
+
+    CURRENT_FUZZY_V1 = "current_fuzzy_v1"
+    """Control: fuzzy title/author match with a hard +/-1 publication-year
+    window; requires publication year as input. Mirrors the baseline gate."""
+    NO_YEAR_FILTER_V1 = "no_year_filter_v1"
+    """Drops the year filter from the query, but still requires publication year
+    as input."""
+    NO_YEAR_FILTER_YEAR_OPTIONAL_V1 = "no_year_filter_year_optional_v1"
+    """Drops the year filter and makes publication year optional as input."""
 
 
 class CandidateIdentifier(GenericExternalIdentifier):
@@ -838,8 +854,8 @@ class CandidateSelectionRequest(BaseModel):
     input: CandidateSelectionInput = Field(
         description="The reference to find candidates for."
     )
-    retrieval_policy: str = Field(
-        default=CURRENT_FUZZY_RETRIEVAL_POLICY,
+    retrieval_policy: RetrievalPolicyName = Field(
+        default=RetrievalPolicyName.CURRENT_FUZZY_V1,
         description=(
             "Named, server-side retrieval policy fixing the full candidate regime "
             "(query shape, year strategy, identifier union). Defaults to the "
@@ -937,7 +953,7 @@ class CandidateSelectionDiagnostics(BaseModel):
 class CandidateSelectionResult(BaseModel):
     """Ranked candidates and diagnostics for a candidate-selection request."""
 
-    retrieval_policy: str = CURRENT_FUZZY_RETRIEVAL_POLICY
+    retrieval_policy: RetrievalPolicyName = RetrievalPolicyName.CURRENT_FUZZY_V1
     index_version: str | None = Field(
         default=None,
         description="The reference index version the candidates were drawn from.",
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -913,8 +913,9 @@ class InputSearchability(BaseModel):
     searchable: bool = Field(
         description=(
             "Whether the input meets the Elasticsearch searchability gate (title, "
-            "authors, publication year). Exact identifier matching runs regardless, "
-            "so candidates may be present even when this is false."
+            "authors, and publication year unless the selected retrieval policy "
+            "makes it optional). Exact identifier matching runs regardless, so "
+            "candidates may be present even when this is false."
         )
     )
     reason: str = Field(description="Explanation of the searchability decision.")
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -652,8 +652,8 @@ class CandidateCanonicalSearchFields(ProjectedBaseModel):
     """
     Fields used for candidate canonical selection.
 
-    The search implementation lives at
-    :attr:`app.domain.references.repository.ReferenceESRepository.search_for_candidate_canonicals`.
+    The retrieval policy is interpreted by
+    :func:`app.domain.references.services.deduplication_service.build_candidate_canonical_search_query`.
     """
 
     publication_year: int | None = Field(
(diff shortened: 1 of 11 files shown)


jaybea:
Following up on @danielbreves's  suggestion to try `selectinload` for the `get_hydrated` eager loads, I pushed an update `perf(dedup)` commit on this branch.

Measured on staging (100 real references, through the Postgres ORM path, warm connection):

- `joinedload`: 58.6 ms median (p95 81 ms)
- `selectinload`: 27.6 ms median (p95 50 ms)

About 2x faster, and the lever is payload rather than row count. `joinedload` cross-joins the two sibling collections, so each enhancement's `content` JSONB is sent once per identifier; on that sample it duplicated enhancement content 3.2x. `selectinload` sends each row once, at the cost of one extra query per collection.

Two caveats for the record:
- The particular `TABLESAMPLE` batch was heavier than corpus average (byte duplication ~3.2x vs the corpus-wide ~1.7x identifier multiplier), so the median win across the full corpus is likely a bit under 2x. Still material: ~66% of references have >=2 identifiers.
- Measured against quiescent staging on a warm connection, so it excludes client-to-service network latency and production concurrency.

Net: this seems to confirm hydration payload, not ES or row count, was the tail-latency source  from #829 stress test.

app/domain/references/models/retrieval_policy.py
```diff
@@ -0,0 +1,93 @@
...
+    union_identifiers: bool
+    year_strategy: YearStrategy
+    requires_publication_year: bool = True
+
+    def is_input_searchable(
+        self, search_fields: CandidateCanonicalSearchFields
+    ) -> bool:
+        """Whether the input has the fields this policy needs for ES retrieval."""
+        has_core = bool(search_fields.title and search_fields.authors)
+        # Falsy year (0/None) counts as absent, matching the baseline all(...) gate.
```
Adam-Hammo, line 49: Let's hope we don't import the Gospel of John

app/domain/references/models/retrieval_policy.py
```diff
@@ -0,0 +1,93 @@
...
+        has_year = bool(search_fields.publication_year)
+        return has_core and (has_year or not self.requires_publication_year)
+
+
+NO_YEAR_FILTER_POLICY = "no_year_filter_v1"
+YEAR_OPTIONAL_POLICY = "no_year_filter_year_optional_v1"
```
Adam-Hammo: Could `CURRENT_FUZZY_RETRIEVAL_POLICY` move here too, to keep the registry all in the same place?

app/domain/references/models/models.py
```diff
@@ -833,9 +833,19 @@ def validate_exactly_one_input_mode(self) -> Self:
 class CandidateSelectionRequest(BaseModel):
     """A request for candidate canonical references, for dedup evaluation."""
 
+    model_config = ConfigDict(extra="forbid")
+
     input: CandidateSelectionInput = Field(
         description="The reference to find candidates for."
     )
+    retrieval_policy: str = Field(
```
Adam-Hammo: It would be nice if this could be an enum so we can autogenerate the available options in docs. Is that possible?
Maybe something like:
```
class RetrievalPolicy(StrEnum):
    CURRENT_FUZZY_RETRIEVAL_POLICY="current_fuzzy_v1"
    ...
```

Adam-Hammo: We can also then autogenerate docs about what each retrieval policy means via docstrings - see our SDK enums for an example

app/domain/references/repository.py
```diff
@@ -679,15 +680,81 @@ def _build_author_dis_max_query(
         # Enough to prefer multi-author overlap without summing to inflation.
         return Q("dis_max", queries=queries, tie_breaker=0.1)
 
+    @staticmethod
+    def _year_filter_clauses(
```
Adam-Hammo: Musing, not for now but to keep in mind as we build this out - I wonder if we should be constructing the query in `deduplication_service` and passing it to the repository. This feels like a bit of domain information and business logic is leaking into the repository.

app/domain/references/repository.py
```diff
@@ -679,15 +680,81 @@ def _build_author_dis_max_query(
...
+        author_query = self._build_author_dis_max_query(
+            search_fields.authors,
+            max_clauses=scoring_config.max_author_clauses,
+            min_token_length=scoring_config.min_author_token_length,
+        )
+        should_clauses = [author_query] if author_query else []
+        # Canonical filter is unconditional: duplicates are never candidates,
+        # whatever the year strategy or year presence. Only the year range is
+        # strategy-dependent. (This also avoids race conditions where two references
+        # being deduplicated concurrently could create conflicting relationships.)
```
Adam-Hammo: Mega nitpick: to me the first half of this comment just says exactly what the code below it makes obvious, and makes it harder to see the actual "why" about the race etc. I think it could be shorter or even removed.

app/domain/references/models/retrieval_policy.py
Adam-Hammo: (Might tie in with my enum comment elsewhere)

Adam-Hammo (commented):
Looking good, just musings about annoying structures and architecture. Sorry 🙈

app/domain/references/models/retrieval_policy.py
jaybea: Ahh, had to look that one up. "And ye shall know the truth, and the truth shall set your year to falsy." (Said in Monty Python ~~falsy~~ falsetto.)

jaybea:
You say "sorry", I say "cool!".  Enum-type autogenerating schema values, and with the SDK-style docstrings! Cool! And being collected together feels much more cohesive than global string consts.

On the refactor, I don't think there will be a better time than now do this, it actually was pretty close to neutral line-wise. You're right it makes more sense to have the domain specific stuff in `deduplication_service.py` rather than creeping into `repository.py`. It's not perfect, but I think it's in the right direction, let me know how it could be better.

Fixed the mega-comment too.

Adam-Hammo (approved):
Looks good!

Issue #752, opened 2026-06-17
Ingest and map previously excluded EEF jacobs domain references with eppi_id as identifier


jack-o-the-green:
## Context 
Previously ~25% of references in the EEF dataset were not imported into destiny-repository because they lacked a form of external identifier. We are now relaxing previous constraints to allow import with an eppi item id as long as those references are ingested into a specific domain (in this case with the annotation `domain-inclusion/jacobs-education`). 

The [eppi import script](https://github.com/destiny-evidence/destiny-repository/blob/main/libs/eppi_importer/import_from_eppi.py) was updated in #745 to add the new command flag `--include-eppi-id`, and the `--tags` flag was updated to allow adding domain inclusion annotations. 

References that failed the first ingest were written out in the same import format to `eef-eppi-export-2025-12-03/parse-failures/` in the `destinyadhocstoragesa` storage account, so these files should be able be re-parsed and imported without the risk of introducing duplicates of the previous import. 
  * Recommend doing a comparison on eppi item id with the jsonl files previous imported (eef-eppi-export-2025-12-03/parsed-for-ingestion` to confirm there is no eppi item id overlap between the two files. 

## Import process
Importing and mapping EEF references has slightly more complexity than other ingests, as the duplicates indicate multiple study arms. So we will need to do the following 

* Parse the references for ingestion with the `--include-eppi-id` and `--tags domain-inclusion/jacobs-education` flags (as well as `--include-raw` and the appropriate sources etc).

> [!NOTE]
> We don't want to use the same import/enhancement `--source`  for this batch as the previous EEF references even though they are part of the previous export. This is because we need a way to filter references down before requesting enhancements against them. `--source-export-date` can stay the same. 

* Ingest the references
* Deduplicate the references in destiny-repository
   *  We can't follow the previously implemented deduplication method https://github.com/destiny-evidence/destiny-repository/issues/570 because we previously deduplicated on external identifier matches and there's no external identifiers in this batch. 
   * Check filtering conditions as this is a second eef import - there might be duplicates we ingest now that are arms of a previously ingested reference. 

> [!NOTE]
> automatic index refreshing is turned off in production currently, you will need to manually trigger it from the elasticloud console with a `POST /reference/_refresh` for the references to show up in search. You'll probably need to do this after ingest, after deduplicating, and after mapping. 

* Add the new source mapping to the vocabulary mapping robot and trigger an enhancement request on the canonical reference id of all references created by the ingest
[...]

PR #839, opened 2026-07-21
Expose import record note on import cli args

Adam-Hammo:
Small update to accommodate #752


jaybea (approved):
Nice small change. Ran it, watched the notes come through, verified them in `import_record.notes`. 

Looks good!

Issue #842, opened 2026-07-21
Design deep pagination interface


Adam-Hammo:
#661 adds the backend implementation required to paginate beyond 10,000 references.

**If** we want to expose that more widely via API (open question!), then:
- How do we represent search state to the caller? We need to preserve state of PIT ID and sort values per request. There's likely a trade-off between convenient abstraction and implementation complexity.
- What endpoints should support it?
- How deep do we want to allow it to go?
- Who should be allowed to go so deep?

Related: #665

PR #843, opened 2026-07-22
Unbounded search

### LINKED ISSUE #661: Support unbounded searching in the persistence layer

Adam-Hammo:
Closes #661 

Adds `scan` generators to the generic ES repository and reference search service.

These take a PIT snapshot and iterate over it using `search_after`, yielding pages of `page_size` until the optional `limit` is reached. There's no upper bound to the number of documents that can be retrieved.

### DIFF
--- a/app/domain/references/repository.py
+++ b/app/domain/references/repository.py
@@ -410,7 +410,12 @@ class ReferenceESRepository(
         limit: int | None = None,
         page_size: int = 500,
     ) -> AsyncGenerator[ESSearchResult, None]:
-        """Scan references matching ``query`` in pages."""
+        """
+        Scan references matching ``query`` in pages.
+
+        ``scan_with_query_string`` appends the ``id`` tiebreaker, so unlike
+        :meth:`search` this method doesn't add one itself.
+        """
         sort_keys: list[str | dict[str, Any]] = list(sort) if sort else ["_score"]
         async for page in self.scan_with_query_string(
             query.query_string,
--- a/app/domain/references/repository.py
+++ b/app/domain/references/repository.py
@@ -2,7 +2,7 @@
 
 import datetime
 from abc import ABC
-from collections.abc import Mapping, Sequence
+from collections.abc import AsyncGenerator, Mapping, Sequence
 from typing import Any, ClassVar, Literal
 from uuid import UUID
 
@@ -26,7 +26,10 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import joinedload
 
 from app.core.config import DedupCandidateScoringConfig, get_settings
-from app.core.telemetry.repository import trace_repository_method
+from app.core.telemetry.repository import (
+    trace_repository_generator,
+    trace_repository_method,
+)
 from app.domain.references.models.es import (
     ReferenceDocument,
     RobotAutomationPercolationDocument,
@@ -399,6 +402,27 @@ class ReferenceESRepository(
             filter_clauses=self._build_filter_clauses(query),
             parse_document=False,
         )
+
+    @trace_repository_generator(tracer)
+    async def scan(
+        self,
+        query: SearchQuery,
+        sort: list[str] | None = None,
+        limit: int | None = None,
+        page_size: int = 500,
+    ) -> AsyncGenerator[ESSearchResult, None]:
+        """Scan references matching ``query`` in pages."""
+        sort_keys: list[str | dict[str, Any]] = list(sort) if sort else ["_score"]
+        async for page in self.scan_with_query_string(
+            query.query_string,
+            fields=self.default_search_fields,
+            limit=limit,
+            page_size=page_size,
+            sort=sort_keys,
+            filter_clauses=self._build_filter_clauses(query),
+            parse_document=False,
+        ):
+            yield page
 
     @trace_repository_method(tracer)
     async def aggregate_facets(
--- a/app/persistence/es/repository.py
+++ b/app/persistence/es/repository.py
@@ -2,7 +2,8 @@
 
 import json
 from abc import ABC
-from collections.abc import AsyncGenerator, Sequence
+from collections.abc import AsyncGenerator, AsyncIterator, Sequence
+from contextlib import asynccontextmanager
 from typing import Any, Generic, Never
 from uuid import UUID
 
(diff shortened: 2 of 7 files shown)


app/domain/references/repository.py
```diff
@@ -399,6 +402,27 @@ async def search(
             parse_document=False,
         )
 
+    @trace_repository_generator(tracer)
+    async def scan(
+        self,
+        query: SearchQuery,
+        sort: list[str] | None = None,
+        limit: int | None = None,
+        page_size: int = 500,
```
jack-o-the-green, line 411: Do we care that all the other default page_sizes are `20`? Assume we're doing big searches with scan so I don't have a problem with it.

app/domain/references/repository.py
```diff
@@ -399,6 +402,27 @@ async def search(
...
+    ) -> AsyncGenerator[ESSearchResult, None]:
+        """Scan references matching ``query`` in pages."""
+        sort_keys: list[str | dict[str, Any]] = list(sort) if sort else ["_score"]
```
jack-o-the-green, line 419: Do we want to add the sort tiebreaker to scanning by default as well for order guarantee?

app/persistence/es/repository.py
```diff
@@ -259,6 +275,140 @@ async def search_with_query_string(  # noqa: PLR0913
         response = await self._execute_search(search)
         return self._parse_search_result(response, page, parse_document=parse_document)
 
+    @asynccontextmanager
```
jack-o-the-green, line 278: `nitpick` I'm getting a warning that is decorator is deprecated

```
The function "asynccontextmanager" is deprecated
  Annotating the return type as `-> AsyncIterator[Foo]` with `@asynccontextmanager` is deprecated. Use `-> AsyncGenerator[Foo]` instead.
```

app/domain/references/repository.py
jack-o-the-green: I've seen the tiebreaker added in down in the persistence/es/repository.py. Are we able to add that in here instead or change the search to add in later. Right now it's harder to see that both `search` and `scan` both tiebreak.

jack-o-the-green (approved):
LGTM some minor structuring comments.

Also wanted to confirm my understanding of what this gives us - this would allow us to do something like export more than 10,000 records, as we'd use the generator to build the full result set and return the jsonl. However, this _wouldn't_ give us the ability do the paged searching across >10,000 records that the repository ui does, because we're not passing in a starting page to the scan/the generator wouldn't live between subsequent api calls. Just want to make sure I haven't missed an implementation quirk.

Adam-Hammo:
Correct on your understanding - this gives us the ability to get as many records as we want in a single task. A follow-up to implement this on exports would be relatively simple.

#842 discusses what it might take across API calls.

app/persistence/es/repository.py
Adam-Hammo: Good catch!

app/domain/references/repository.py
Adam-Hammo: Yeah this is a deliberate nudge that this repository method is generally used to fry bigger fish.

app/domain/references/repository.py
Adam-Hammo: I've added a note to the docstring about this. To me there's a subtle difference between the two - tiebreaking is an invariant property of PIT scanning, always required, whereas it isn't necessarily for arbitrary searching.

Adam-Hammo: Moving the `search` tiebreak down into the generic layer wouldn't unify the logic either - `scan` has a tiebreak fallback to `_shard_doc` that `search` doesn't have (it instead falls back to no tiebreak)

jack-o-the-green: Docstring note works for me 👌

jack-o-the-green (approved):
LGTM

PR #854, opened 2026-07-23
test(e2e): allow dedup collision retries to settle

### LINKED ISSUE #547: Fix E2E test flake on tests/e2e/imports/test_deduplication.py::test_deduplication_shortcut_from_scratch
We're seeing occasional test flake on our deduplication shortcut from scratch test in our e2e tests, this has only shown up on github. @Adam-Hammo has called out that we've only seen this on merges to `main` when we have a lot of actions running, so it may be a timing and/or resources issue. 

[Failing Action](https://github.com/destiny-evidence/destiny-repository/actions/runs/21848416322/attempts/2)

[honeycomb logs for github actions](https://ui.honeycomb.io/destiny-evidence/environments/github-actions/result/dq75TWtQoSU?tab=traces).

jaybea:
## Summary

Fixes the intermittent e2e failure in `test_deduplication_shortcut_from_scratch`
(tracked in #547). It surfaced only on post-merge `main` runs under heavy Actions
load, e.g. the [failed e2e job](https://github.com/destiny-evidence/destiny-repository/actions/runs/29887841465/job/88822026167)
([run, attempt 1](https://github.com/destiny-evidence/destiny-repository/actions/runs/29887841465/attempts/1)),
while PR #835's [pre-merge e2e passed](https://github.com/destiny-evidence/destiny-repository/actions/runs/29829623842).

## Root cause

The fixture imports two references sharing a trusted identifier in one batch, so
their deduplication tasks run concurrently. Both take the identifier shortcut and
write side-effect active decisions, racing on the
`uix_reference_one_active_decision_constraint` partial unique index. The task
recovers by retrying (`process_reference_duplicate_decision`), and during that
recovery the reference transiently holds no active decision.

`poll_duplicate_process` stopped after a fixed five attempts (~4s). When the
collision-retry recovery took longer than that, the poll saw zero active
decisions and raised `TestPollingExhaustedError` (`tests/e2e/utils.py`), even
though the decision settled shortly after.

This is a test-timing issue rather than a production problem: the underlying
race (lineage in #534) is handled and self-healing. The worker log for the
failed attempt shows the retry completing its side effects, and the re-run
passed, so `main` is green. We're leaving the production path as-is, in keeping
with that self-healing behaviour and its production track record.

## Change

- `poll_duplicate_process` polls on an elapsed-time budget (default 20s) rather
  than a fixed attempt count.
- The active-decision invariant is unchanged: a decision is only active once
  terminal (`check_active_decision_is_terminal`), so `required_state=None` still
  waits for a settled decision.
- The separate `sleep(10)` workaround (delete-then-rededup transition) is
  untouched.

## Validation performed

- Added deterministic unit tests (`tests/unit/test_e2e_poll_duplicate_process.py`)
  with mocked clock and sleep, pinning both the "returns a decision that settles
  past the old attempt budget" and "raises on the elapsed-time budget" paths.
  Confirmed they fail against the former five-attempt behaviour and pass against
  the elapsed-time budget.
- `uv run pytest tests/unit/test_e2e_poll_duplicate_process.py` green (2 passed).
- `pre-commit run --files ...` clean (ruff, ruff-format, mypy, vulture).

Closes #547
Refs #534

### DIFF
--- a/tests/e2e/utils.py
+++ b/tests/e2e/utils.py
@@ -1,6 +1,8 @@
 """Utility functions for import end-to-end tests."""
 
+import asyncio
 import datetime
+import time
 from collections.abc import Callable, Mapping
 from contextlib import _AsyncGeneratorContextManager
 from uuid import UUID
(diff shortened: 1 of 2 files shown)


Adam-Hammo (approved):
LGTM, good find and thanks for fixing!

PR #856, opened 2026-07-24
feat(dedup): add soft_year_decay candidate-retrieval policy

jaybea:
## Summary

Adds one new versioned candidate-retrieval policy and the single Elasticsearch
translation feature it needs: a bounded publication-year proximity bonus. This lets
soft publication-year decay be evaluated through the existing Candidate Selection API
without touching the production nomination path.

New policies, exposed by name through `POST /v1/references/deduplication/candidates/`
(evaluator-facing only):

- `soft_year_decay_v1` — drops the hard ±1 publication-year filter and instead
  multiplies in a bounded native Elasticsearch exponential year-proximity bonus (at
  most +10%, halving around ten years). Year becomes corroboration, not a gate:
  nothing is excluded, near-year candidates just rank above far-year ones.
- `soft_year_decay_nonfuzzy_probe_v1` — evaluation-only probe: the same decay with a
  zero-edit-distance (non-fuzzy) title match, to measure the recall and latency impact
  of disabling fuzzy title matching for this query shape.

How it is layered:

- Immutable, versioned decay parameters (`YearDecayConfig`) live on the policy. The
  service query builder resolves them into a query-carried `YearDecay` (adding the
  per-query origin); the repository translates a decay-bearing query into an ES
  `function_score` wrapping the otherwise-unchanged `bool`. The repository only
  translates a resolved query object; the policy behaviour is chosen in the service.
- Title fuzziness moves from a hardcoded builder literal to a policy attribute so the
  probe can set it. Behaviour-preserving for every existing policy (all default
  `AUTO`).

## Benchmark support: track_total_hits (default-preserving)

Also adds `track_total_hits` as a per-request option on the candidate-selection
endpoint. The default stays `true`, so existing responses keep exact totals, which
reveal the true candidate-pool size past the Elasticsearch 10k default cap (useful
for evaluation, and measured latency-neutral on staging: p50 ~268ms vs ~260ms). A
benchmark sets it to `false` for the Elasticsearch default lower-bound total,
matching the production nomination path. This is configurable for benchmark
comparability, default-preserving, and changes no production behaviour.

## Not changed

- Default candidate-selection policy remains `current_fuzzy_v1`.
- The production nomination path is unchanged: `nominate_candidate_canonicals` still
  uses `current_fuzzy_v1` and leaves `track_total_hits` off.
- Scoring config and duplicate-decision actuation are unchanged; the endpoint stays
  read-only and writes no decision or candidate state.
- No persisted-model or migration changes. Domain models and the ES query only; the
  retrieval policy is not part of the SDK.

## Validation performed

- Typecheck clean (`pyright`) on the changed modules; full references unit suite green.
- ES integration arm proves `soft_year_decay_v1` accepts the `function_score` DSL,
  preserves the no-filter matching set (the ten-year-drifted duplicate is still
  returned), and ranks the near-year duplicate above the far-year one.
- Live-validated the endpoint against local DR by capturing the emitted query per
  policy via search slowlog: `soft_year_decay_v1` emits the `function_score` with the
  `exists: publication_year` guard and no range filter; `current_fuzzy_v1` (control)
  keeps the `range publication_year` filter and emits no `function_score`; the
  non-fuzzy probe emits `fuzziness:"0"`. The multipliers observed against real
  Elasticsearch matched the spec (missing-year 1.00, near-year 1.10, ten-year 1.05).
- Confirmed the searchability gate blocks a yearless `soft_year_decay_v1` input from
  the ES route, and an unknown policy returns HTTP 422 at request validation.
- Live-validated `track_total_hits`: `false` omits the count from the emitted query
  (Elasticsearch default lower bound); default `true` emits the exact full count.

## Follow-ups

- Weight/curve calibration, the recall/precision sweep, and the fuzzy-vs-non-fuzzy
  comparison run in the evaluation harness. These policies are exposed by name only
  and are not wired into the nomination default.

### DIFF
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -836,9 +836,9 @@ class RetrievalPolicyName(StrEnum):
     exponential proximity bonus adds at most 10% to the title/author score.
     Requires publication year as input."""
     SOFT_YEAR_DECAY_NONFUZZY_PROBE_V1 = "soft_year_decay_nonfuzzy_probe_v1"
-    """Evaluation-only probe: soft_year_decay_v1 with a non-fuzzy (zero
-    edit-distance) title match, to measure the recall cost of dropping fuzzy
-    title expansion. Not for production nomination."""
+    """Evaluation-only probe: soft_year_decay_v1 with zero edit-distance title
+    matching. Used to measure the recall and latency impact of disabling fuzzy
+    title matching for this query shape. Not for production nomination."""
 
 
 class CandidateIdentifier(GenericExternalIdentifier):
@@ -926,9 +926,9 @@ class CandidateSelectionRequest(BaseModel):
         default=True,
         description=(
             "Compute the exact Elasticsearch total-hit count for reporting. Set "
-            "false for the production-throughput proxy: the total becomes a lower "
-            "bound and the query skips the full count, which is expensive on broad "
-            "(e.g. no-year-filter) queries."
+            "false for the production-throughput proxy: Elasticsearch may return a "
+            "lower-bound total and can avoid exact full-count work, which is "
+            "especially relevant on broad no-year-filter queries."
         ),
     )
 
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -925,10 +925,11 @@ class CandidateSelectionRequest(BaseModel):
     track_total_hits: bool = Field(
         default=True,
         description=(
-            "Compute the exact Elasticsearch total-hit count for reporting. Set "
-            "false for the production-throughput proxy: Elasticsearch may return a "
-            "lower-bound total and can avoid exact full-count work, which is "
-            "especially relevant on broad no-year-filter queries."
+            "Compute the exact Elasticsearch total-hit count. Defaults to true: "
+            "exact totals reveal the true candidate-pool size that the "
+            "Elasticsearch 10k default caps, which is useful for evaluation. Set "
+            "false for the Elasticsearch default (a lower-bound total), matching "
+            "the production nomination path."
         ),
     )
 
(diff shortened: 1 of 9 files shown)


app/domain/references/models/models.py
```diff
@@ -891,6 +922,16 @@ class CandidateSelectionRequest(BaseModel):
...
         description="Include hydrated bibliographic fields on each candidate.",
     )
+    track_total_hits: bool = Field(
+        default=True,
+        description=(
+            "Compute the exact Elasticsearch total-hit count. Defaults to true: "
+            "exact totals reveal the true candidate-pool size that the "
+            "Elasticsearch 10k default caps, which is useful for evaluation. Set "
+            "false for the Elasticsearch default (a lower-bound total), matching "
+            "the production nomination path."
```
Adam-Hammo: Nitpick: this is pretty wordy

app/domain/references/models/retrieval_policy.py
```diff
@@ -9,26 +9,28 @@
...
 
 class YearStrategy(Enum):
     """
     How a policy treats the publication-year signal.
 
-    Extension seam: HARD_WINDOW keeps the hard ±1 filter, NO_FILTER drops the
-    year clause. A soft year-distance decay strategy (SOFT_DECAY) is added here
-    once we have grounding on its weights/policy.
+    HARD_WINDOW keeps the hard ±1 filter, NO_FILTER drops the year clause, and
+    SOFT_DECAY drops the hard filter while applying a bounded year-proximity bonus.
```
Adam-Hammo: Nit: these couple lines might be better placed as docstrings for each member individually

Adam-Hammo (approved):
Just a couple of very nitty docs things, feature looks good

app/domain/references/models/models.py
jaybea: Yeah, got it.

app/domain/references/models/retrieval_policy.py
jaybea: Makes more sense. Gotta use those doc-strings.

PR #857, opened 2026-07-24
Enhancement request by search

### LINKED ISSUE #838: Request enhancements from a search
Add a new endpoint or extend the existing one to allow a search to trigger an enhancement request. The work will need to occur in a prioritised task.

The search must be allowed to be arbitrarily large - eg https://github.com/destiny-evidence/impact-case-1/blob/main/ic1/query/revisions/query_20260408.py.

This must use scanning implemented in #661.

Consider:
- disabling scoring on the query (it just iterates over all with no need for ordering by default)
- implementing a way to check the status of the task (how far through it is out of total amount, when is it complete, did it error)
- what search fields are necessary (minimum: query string)
- a dry-run parameter to synchronously report back the number of references it would match

Adam-Hammo:
Closes #838:
- adds search columns to `EnhancementRequest` in the DB
- adds `POST /enhancement-requests/search/` endpoint 
  - when invoked with `dry-run=True`, synchronously returns the number of references that would have enhancements requested, and checks the robot exists
    - adds `count()` to the ES repository to support this
  - when invoked otherwise, asynchronously queues a high-priority long-running task which:
    - uses `scan` from #661 to iterate over all matching documents
      - N.B. added the option to use unscored queries here - this is much faster in scenarios where we are treating conditions as filters and retrieving all results
    - requests enhancements page-by-page for each reference
- adds `GET /enhancement-requests/search/id/` endpoint
  - returns the status of the search + request
  - can also be used after the search is complete to monitor enhancement fulfilment
- adds SDK models to support the above

Not covered: resilience. This is a long-running task and resilience needs _some_ consideration, however this PR implements the MVP. 
Our base case in a task failure now would be re-running - you could add an exclusion for the requested enhancement to the query too to minimise duplicate enhancements. More on this next week.

API docs:
image
image


tests/unit/domain/references/test_service.py
```diff
@@ -1283,3 +1287,144 @@ async def expire_pending_enhancements_past_expiry(self, now, statuses):
...
+    search_service = _search_service_counting(ESSearchTotal(value=7, relation="eq"))
+
+    with patch.object(service, "_search_service", search_service):
+        total = await service.count_search_matches(SearchQuery(query_string="climate"))
+
+    assert total.value == 7
+
+
+@pytest.mark.asyncio
+async def test_collect_pending_enhancements_from_search(fake_repository, fake_uow):
```
danielbreves, line 1340: Worth adding a test for the `n_matched=0` branch?

app/domain/references/service.py
```diff
@@ -735,6 +736,129 @@ async def create_pending_enhancements(
...
+    @sql_unit_of_work
+    async def register_search_enhancement_request(
+        self, enhancement_request: EnhancementRequest
+    ) -> EnhancementRequest:
+        """Persist a pending search enhancement request."""
+        await self.sql_uow.enhancement_requests.add(enhancement_request)
+        return enhancement_request
+
+    @es_unit_of_work
+    async def collect_pending_enhancements_from_search(
```
danielbreves, line 752: Plagiarising your previous feedback, would it make sense to move these to another service? I couldn't see one where this would obviously belong, so I wondered whether we're missing an enhancement_request_service.py? Though it would likely be a significant refactor to properly move this, so feel free to ignore.

danielbreves (approved):
Very solid MVP! It seems fair to defer the resilience considerations, similar to how we deferred leasing/expiring pending enhancements perhaps.

app/domain/references/service.py
Adam-Hammo: This is a good challenge - I think you're right to consider a sub-service now this logic is growing. Will have a gander

Adam-Hammo: Moving _all_ of the methods is relatively involved due to cross-service dependencies, but I'm able to move the low-level methods to `EnhancementService` cleanly.

danielbreves (approved):
LGTM!

PR #858, opened 2026-07-24
Support relevance as a search parameter

jack-o-the-green:
## Context 
Resolves errors seen here https://covidence.slack.com/archives/C099F6PR8R2/p1784816418339419 with a user attempting to sort by relevance (which was previously implied to be available as a parameter by our documentation https://destiny-evidence.github.io/destiny-repository/procedures/search.html#sort).

## What's Changed

I've introduced a normalizing function to convert `relevance` to a `_score` search dict in the ReferenceESRepository. 

## Design Decision
Bare relevance defaults to a **descending order search**, as a user ordering by relevance would expect best matches to be returned first. This is opposite behaviour to document fields like publication year, which sort ascending by default. Have added some documentation within codebase to this.

### DIFF
--- a/app/domain/references/repository.py
+++ b/app/domain/references/repository.py
@@ -284,6 +284,9 @@ class ReferenceESRepository(
     }
     """Mapping from a facet type to the ES field its counts are aggregated on."""
 
+    _RELEVANCE_SORT_KEY: ClassVar[str] = "relevance"
+    """Public sort token for ES's built-in ``_score``; the default sort."""
+
     def __init__(self, client: AsyncElasticsearch) -> None:
         """Initialize the repository with the Elasticsearch client."""
         super().__init__(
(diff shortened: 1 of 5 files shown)


jack-o-the-green:
@acdonald would love your eyes over the design decision. It's an intentional inconsistency currently, but good to decide if we think it's worth it.

acdonald:
thanks @jack-o-the-green - feels like a good one to fix the inconsistency with and allow users to do

app/domain/references/repository.py
```diff
@@ -392,12 +408,13 @@ async def search(
         tiebreaker: dict[str, Any] = {
             "id": {"order": "desc", "unmapped_type": "keyword"}
         }
+        sort_keys = [self._normalize_sort_key(key) for key in (sort or ["relevance"])]
```
Adam-Hammo: Should this use `_RELEVANCE_SORT_KEY` instead of the magic string? Same below

jack-o-the-green:
Ann sorry @acdonald badly worded comment on my part - this PR is intentionally introducing an inconsistency to search behaviour because I believe it would better align with user expectations, not fixing one.

Adam-Hammo (approved):
One nit, code looks good. Should we update the docs at https://destiny-evidence.github.io/destiny-repository/procedures/search.html#sort also?

app/domain/references/repository.py
jack-o-the-green: Yes it should 🤦

Adam-Hammo (approved):

PR #861, opened 2026-07-27
Add db-level idempotency guard for pending enhancements

Adam-Hammo:
Fast follow of #838 to ensure every search enhancement request reaches a terminal state.

Main case this handles is a worker going down mid-task, which would leave the task in a `searching` state, which would then be ignored when the task is redelivered. Now, redelivered tasks in that state are picked up again and idempotently (if naively) finish requesting the enhancements.

It does this by adding a unique index (`enhancement_request_id`, `reference_id`) where not a retry, and ignoring conflicts. I've checked prod and staging, no existing rows violate this constraint, and the migration contains a fast-fail if one materialises before we run it.

This also covers the existing id-list endpoint. Rather than letting duplicate ids there silently dedupe at the DB, that endpoint now returns a 422.

### DIFF
--- a/app/domain/references/models/sql.py
+++ b/app/domain/references/models/sql.py
@@ -801,14 +801,14 @@ class PendingEnhancement(GenericSQLPersistence[DomainPendingEnhancement]):
         "EnhancementRequest", back_populates="pending_enhancements"
     )
 
-    original_uniqueness_index = Index(
-        "uq_pending_enhancement_request_reference_original",
+    non_retry_uniqueness_index = Index(
+        "uq_pending_enhancement_request_reference_non_retry",
         "enhancement_request_id",
         "reference_id",
         unique=True,
         postgresql_where=text("retry_of IS NULL"),
     )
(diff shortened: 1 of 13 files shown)


app/domain/references/models/sql.py
```diff
@@ -800,6 +801,17 @@ class PendingEnhancement(GenericSQLPersistence[DomainPendingEnhancement]):
         "EnhancementRequest", back_populates="pending_enhancements"
     )
 
+    original_uniqueness_index = Index(
```
jack-o-the-green: `nitpick` original feels like an onverloaded term - it's easy to conceptualize `original_uniqueness_index` as "the first uniqueness index we had" which could send people looking for newer ones etc. I don't know if there's a better option though.

jack-o-the-green (approved):
LGTM - this is a neat solution 👌

app/domain/references/models/sql.py
Adam-Hammo: Fair call, renaming to `non_retry_uniqueness_index`

Issue #862, opened 2026-07-27
Improve resiliency of the stale pending-enhancement reaper


jaybea:
## Summary

The scheduled task `expire_and_replace_stale_pending_enhancements` works the whole set of stale `PROCESSING` pending enhancements in one unbounded pass inside a single transaction, with a per-row recursive query in the loop. At small volumes that finishes and commits fine. Once the set is large enough that a run can't complete within its window, the run is cut short, nothing commits, and the next run starts over on the same rows, so the backlog never comes down.

This happened on staging with a backlog of ~103k rows. The backlog has since been manually cleared, so this issue is about making the task hold up under loads that would be encountered in production.

The task is our recovery mechanism for enhancements that were claimed and never returned. The failure is silent: no exception is raised.

Dataset `destiny-repository-scheduled-job-staging`:

See the task execution history on honeycomb:
- Onset zoom: https://ui.honeycomb.io/destiny-evidence/environments/staging/datasets/destiny-repository-scheduled-job-staging/result/7emhvcmJZKN
- 24h per-run breakdown: https://ui.honeycomb.io/destiny-evidence/environments/staging/datasets/destiny-repository-scheduled-job-staging/result/954SBFdY3iW
- Robot poll status by robot_id: https://ui.honeycomb.io/destiny-evidence/environments/staging/datasets/destiny-repository-staging/result/oUWveKtVfTj

I think this is caused by:

1. **The expire is unbounded.** `PendingEnhancementSQLRepository.expire_pending_enhancements_past_expiry` (`app/domain/references/repository.py`) issues a single `UPDATE ... SET status=EXPIRED WHERE expires_at < now AND status IN (PROCESSING) RETURNING *` with no batch limit, so it takes on the entire backlog at once and returns every row.

2. **Retry depth is recomputed per row.** `ReferenceEnhancementService.create_retry_pending_enhancements` (`app/domain/references/services/enhancement_service.py`) loops over the returned rows and calls `PendingEnhancementSQLRepository.count_retry_depth` for each. That's a recursive CTE walking the `retry_of` chain, so the query count scales one-to-one with backlog size.

[...]

PR #863, opened 2026-07-27
Batch expired enhancement retry depths into one query

Adam-Hammo:
Speculative mitigator for #862 

Turns `n+1` queries into `2` queries.


app/domain/references/repository.py
```diff
@@ -1362,48 +1362,44 @@ async def find_available_for_robot(
...
+        base = (
             select(
-                SQLPendingEnhancement.id,
+                SQLPendingEnhancement.id.label("seed_id"),
                 SQLPendingEnhancement.retry_of,
                 literal(0).label("depth"),
             )
-            .where(SQLPendingEnhancement.id == pending_enhancement_id)
-            .cte(name="retry_chain", recursive=True)
+            .where(SQLPendingEnhancement.id.in_(ids))
```
jaybea: Looks like this hits the asyncpg query argument limit, the ol' 32,767 variety. A workaround could be passing the ids as a single array instead.

```suggestion
            .where(
                SQLPendingEnhancement.id
                == any_(literal(ids, postgresql.ARRAY(postgresql.UUID(as_uuid=True))))
            )
```

Needs `any_` adding to the `from sqlalchemy import (...)` block and `from sqlalchemy.dialects import postgresql`, so it won't apply clean on its own.

jaybea (changes requested):
The batching looks right, and in testing on my machine at least, it is fast. Postgres 17 2,000 id call is 7.9 ms against roughly 1.1s for the same per-row loop.

But I ran into a gotcha in replicating the original issue with the 100k+ backlog. Past 32k, Postgres runs into a limitation on query arguments.

I proposed a fix using an ARRAY in repository.py, take a look. I tested this locally with the 100k backlog.

app/domain/references/repository.py
Adam-Hammo: Oooh this is a good catch. I reckon we have this problem elsewhere too. Let me sus

Adam-Hammo: See what you think!

jaybea (approved):
Looks good! Works great with 40,000 ids in a single query. 

Maybe just consider a test at `get_retry_depths` above the limit, since that's the site that started this and the one with the unbounded id list.

 Amazing how fast these run, 40k ids barely registers.

PR #865, opened 2026-07-28
Add container app env cpu guardrail

Adam-Hammo:
Mitigating change for the incident seen in https://covidence.slack.com/archives/C0810ARPCRG/p1785207268941949 by adding a guardrail to ensure that terraform variables can't be configured in a way that exceeds our container app env CPU quota.

Speculative TF plan deliberately failing to show the current misconfiguration that led to the above. I will drop the number of tasks replicas to 40 in production (update: done).

image


danielbreves (approved):
Nice, LGTM!

PR #871, opened 2026-07-29
Add sdk, cli, and docs for search enhancement request

### LINKED ISSUE #868: Add documentation and SDK helpers

Adam-Hammo:
Closes #868

Adds simple SDK helpers, pivots the CLI to use them instead of the ghetto chaining, and adds docs.

Core docs:

[Requesting Enhancements from a Search — DESTINY Repository v1.0 documentation.pdf](https://github.com/user-attachments/files/30492520/Requesting.Enhancements.from.a.Search.DESTINY.Repository.v1.0.documentation.pdf)


cli/client.py
```diff
@@ -57,7 +60,14 @@ def parse_args(  # type: ignore[override]
...
+        """
+        Parse arguments and attach configured clients to the namespace.
+
+        Both ``oauth_client`` (the SDK :class:`OAuthClient`) and ``client`` (its
+        underlying ``httpx.Client``) are attached; they share one session, so
+        scripts can use whichever fits.
+        """
         parsed = super().parse_args(args, namespace)
-        parsed.client = get_client(parsed.env, parsed.url)
+        parsed.oauth_client = get_oauth_client(parsed.env, parsed.url)
```
jack-o-the-green, line 71: Is the reason we're bothering to support both types because we don't want to update every cli script in this PR? i.e. would there be a reason to deprecate the use of the get httpx.Client function now for later tidyup?

cli/request_enhancements_for_search.py
```diff
@@ -1,102 +1,94 @@
...
-def _preview(reference_ids: list[UUID], limit: int = 10) -> None:
-    """Print the first `limit` reference IDs."""
-    for reference_id in reference_ids[:limit]:
-        print(reference_id)
-    if len(reference_ids) > limit:
-        print("...")
+def poll_search_enhancement_request(
+    client: OAuthClient,
+    request_id: UUID,
+    poll_interval: float = 5,
```
jack-o-the-green: Reckon we're expecting these to be a bit longer running than usual, update default poll_interval to 30s?

cli/request_enhancements_for_search.py
```diff
@@ -1,102 +1,94 @@
...
+        robot_id=robot_id,
+        search_query=query,
+        source=source,
+    )
+    print(
+        f"Created search enhancement request {request.id} "
+        f"(search_status={request.search_status})."
+    )
+
+    status = poll_search_enhancement_request(client, request.id, poll_interval)
```
jack-o-the-green: Maybe change name of `poll_search_enhancement_request` to `poll_until_search_complete`?

cli/request_enhancements_for_search.py
```diff
@@ -1,102 +1,94 @@
...
-    reference_ids = [rid for rid in all_reference_ids if rid not in exclude]
-    if exclude:
-        print(f"{len(reference_ids)} references after exclusions.")
-
-    if not reference_ids:
-        print("No references to enhance. Exiting.")
-        return
-
     if dry_run:
+        total = client.request_search_enhancement(
```
jack-o-the-green, line 61: I would now expect the dry_run to take quite a long time if the search is big - should call this out in the procedure doc?

docs/procedures/requesting-search-enhancements.rst
```diff
@@ -0,0 +1,106 @@
...
+    # Dry run
+    total = client.request_search_enhancement(
+        robot_id="<robot-uuid>",
+        search_query='title:"climate change" AND abstract:health',
+        dry_run=True,
+    )
+    print(f"{total.count} references would have enhancements requested")
+
+    # Send request
+    request = client.request_search_enhancement(
```
jack-o-the-green, line 75: I don't think we should provide a code block that executes both a dry run and an actual run in the same script. I can see a world where someone copies this directly without thinking too hard about it and then just fills out the necessary fields and executes the whole thing.

libs/sdk/src/destiny_sdk/client.py
```diff
@@ -932,6 +934,106 @@ def lookup(
         self._raise_for_status(response)
         return TypeAdapter(list[Reference]).validate_python(response.json())
 
+    @overload
```
jack-o-the-green, line 937: Oh hell yeah this is the first I've seen `@overload`

Adam-Hammo: Usually it's a hint that your function is trying to do too much, but it feels okay here!

cli/client.py
Adam-Hammo: Yeah, it's pretty much a "use the SDK if you can, otherwise call the API yourself via the underlying client". Deprecating use of the httpx client requires adding a bunch of client methods for importing etc (possibly the long-term goal but not required presently)

jack-o-the-green (approved):
LGTM!

PR #872, opened 2026-07-29
Fix tracing code broken by starlette version bump

mootpointer:
When we bumped the starlette version with FastAPI, it broke the tracing middleware we were using to get the path params. The reason we had to jump through hoops is because starlette does routing *after* middleware which meant we had to essentially reach in and do some routing ourselves to get the path params.

What we now do instead is do the routing as part of a FastAPI dependency which occurs after routing is done (hooray!).

The relevant tests have also been fixed so we'll catch this kind of issue in the future.

### DIFF
--- a/app/api/root.py
+++ b/app/api/root.py
@@ -4,7 +4,7 @@ import pathlib
 from collections.abc import Callable
 from contextlib import AbstractAsyncContextManager
 
-from fastapi import APIRouter, FastAPI, Request
+from fastapi import APIRouter, Depends, FastAPI, Request
 from fastapi.middleware import Middleware
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.openapi.docs import get_redoc_html
@@ -32,7 +32,7 @@ from app.core.exceptions import (
     SDKToDomainError,
     VocabularyFetchError,
 )
-from app.core.telemetry.fastapi import FastAPITracingMiddleware
+from app.core.telemetry.fastapi import FastAPITracingMiddleware, trace_path_params
 from app.core.telemetry.logger import get_logger
 from app.domain.imports.routes import router as import_router_v1
 from app.domain.references.routes import (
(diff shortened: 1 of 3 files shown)


app/core/telemetry/fastapi.py
```diff
@@ -89,6 +77,32 @@ async def dispatch(
...
+        None, after binding path parameters to the span and log context.
+
+    """
+    current_span = trace.get_current_span()
+    if current_span.is_recording():
+        for key, value in request.path_params.items():
+            current_span.set_attribute(
+                f"{Attributes.HTTP_REQUEST_PATH_PARAMS}.{key}", str(value)
+            )
+    with bound_contextvars(**request.path_params):
```
mootpointer, line 97: I'll be honest, I don't quite know why we need to bind the context vars but it was done in the previous implementation so I've retained it.

app/api/root.py
```diff
@@ -51,7 +51,9 @@
 
 def create_v1_router() -> APIRouter:
     """Create the v1 API router with all domain-specific routers."""
-    api_v1 = APIRouter(prefix="/v1", tags=["v1"])
+    api_v1 = APIRouter(
+        prefix="/v1", tags=["v1"], dependencies=[Depends(trace_path_params)]
```
Adam-Hammo, line 55: ~Is it worth doing this on the `app` level?~ Decided I like this, app-level creates other concerns

app/core/telemetry/fastapi.py
```diff
@@ -89,6 +77,32 @@ async def dispatch(
             return result
 
 
+async def trace_path_params(request: Request) -> AsyncGenerator[None, None]:
+    """
+    Attach request path parameters to the current span and log context.
+
+    This is a FastAPI dependency rather than middleware logic because Starlette
```
mootpointer: Probably don't need this in the docstring.

Adam-Hammo (approved):
Looks good to me, thank you!

Worth noting a minor regression that path parameters won't trace on routing 404s as the dependency doesn't fire, but that seems pretty low-value so a fair trade-off.

mootpointer:
> Worth noting a minor regression that path parameters won't trace on routing 404s as the dependency doesn't fire, but that seems pretty low-value so a fair trade-off.

I don't know that we need path params for a routing 404, as that's a path problem as much as it is a params problem, right?

app/core/telemetry/fastapi.py
Adam-Hammo: It propagates down to our structured logging by memory

Adam-Hammo:
> I don't know that we need path params for a routing 404, as that's a path problem as much as it is a params problem, right?

Yeah agreed, the only possible use case I can think of would be to help triangulate a request but that's clutching at straws - just included the regression note for completeness

PR #879, opened 2026-07-30
Sort by `_doc` when scanning with an arbitrary order, always tiebreak by `_shard_doc`

Adam-Hammo:
Fixes performance issues when scanning all documents for a search in an arbitrary order.

After investigation in Elasticsearch:
- `_shard_doc` does nothing special performance wise despite documentation: https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/search/sort/ShardDocSortField.java#L40C81-L40C93.
- `_doc` however does support pruning.
- However, `_doc` is unreliable as it's a per-shard ID (also per-replica but that doesn't matter in a PIT). So it's not unique, and this is bad for `search_after` as you can skip/duplicate documents. However however, we can find use for `_shard_doc`  as a tiebreaker, as it's globally unique across the PIT.
- So, sorting and search_aftering on `[_doc, _shard_doc]` does the trick.

### DIFF
--- a/app/persistence/es/repository.py
+++ b/app/persistence/es/repository.py
@@ -439,6 +439,7 @@ class GenericAsyncESRepository(
         - if you're investigating reduced performance after an ES upgrade perchance,
         try going back to just ``_shard_doc``. If that works, proceed to deliver me a
         crisp high-five.
+        See also: https://github.com/elastic/elasticsearch/issues/155559.
 
         We always append ``_shard_doc`` as a tiebreaker to guarantee a total order.
         This is required even when using ``_doc`` as it is not unique.
--- a/app/persistence/es/repository.py
+++ b/app/persistence/es/repository.py
@@ -431,16 +431,22 @@ class GenericAsyncESRepository(
         self, sort: list[str | dict[str, Any]] | None
     ) -> list[str | dict[str, Any]]:
         """
-        Append a total-order tiebreaker to the caller's sort keys.
+        Order the scan for efficient deep pagination under a PIT.
 
-        Prefer the document ``id`` where it exists, otherwise fall back to
-        ``_shard_doc``, ES's PIT-only built-in tiebreaker.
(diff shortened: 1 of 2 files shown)


app/persistence/es/repository.py
```diff
@@ -431,16 +431,22 @@ def _scan_sort_keys(
...
-        Append a total-order tiebreaker to the caller's sort keys.
+        Order the scan for efficient deep pagination under a PIT.
 
-        Prefer the document ``id`` where it exists, otherwise fall back to
-        ``_shard_doc``, ES's PIT-only built-in tiebreaker.
+        If no sort is supplied, we lead with ``_doc``: as the primary sort it lets ES
+        prune when querying each page, providing the fastest behaviour.
+        This is in contradiction to the documentation and seems to mimic old behaviour
+        - if you're investigating reduced performance after an ES upgrade perchance,
+        try going back to just ``_shard_doc``. If that works, proceed to deliver me a
```
jack-o-the-green, line 440: `nitpick` should we biff a ticket in the backlog for doing this after the next upgrade and reference this on it?

jack-o-the-green (approved):
LGTM

app/persistence/es/repository.py
Adam-Hammo: I'd lean no - it's completely speculative, nothing actually says it will behave this way after a future upgrade, so it feels like confusing ticket bloat. I just wanted to give a hint because _if_ it happened it'd be awful to debug.

Adam-Hammo: It's also nigh on impossible to pin this behaviour to a test without said test taking ages to run - the seed data required to make the performance noticeable is huge.

Adam-Hammo: Adding a link in the comment to https://github.com/elastic/elasticsearch/issues/155559

PR #880, opened 2026-07-31
Search enhancement request niggles

Adam-Hammo:
A couple of niggles noticed when running in staging:
- CLI dry runs should have a longer timeout to allow for counting
- Plumb in elasticsearch timeout/retry config so we can adjust it in production
- Performance improvements:
  - Remove overthought code for the bulk conflict-do-nothing execute - let SQLAlchemy handle it. `execute()` with `rows=` proved _much_ faster than `.values()`. Each page went from 6s to <1s to insert in staging.
  - Overlap ES reads with SQL inserts. 10-50% performance increase, depending on the search query.

Staging run post-fixes: https://ui.honeycomb.io/destiny-evidence/environments/staging/datasets/destiny-repository-task-staging/result/ncNMEyyxxRL/trace/dDoFgRiR7zh?fields[]=s_name&fields[]=s_serviceName&span=0ac752c69d7c7f8b
Main redeployed to staging.

### DIFF
--- a/app/utils/aio.py
+++ b/app/utils/aio.py
@@ -29,7 +29,9 @@ async def prefetch(source: AsyncGenerator[T, None]) -> AsyncGenerator[T, None]:
             async with contextlib.aclosing(source) as stream:
                 async for item in stream:
                     await queue.put(item)
-        except Exception as exc:  # noqa: BLE001 - re-raised in the consumer
+        except asyncio.CancelledError:
+            raise
+        except BaseException as exc:  # noqa: BLE001 - re-raised in the consumer
             await queue.put(exc)
         else:
             await queue.put(_DONE)
@@ -45,5 +47,9 @@ async def prefetch(source: AsyncGenerator[T, None]) -> AsyncGenerator[T, None]:
             yield item
     finally:
         producer.cancel()
-        with contextlib.suppress(asyncio.CancelledError):
+        try:
             await producer
+        except asyncio.CancelledError:
+            consumer = asyncio.current_task()
+            if consumer is not None and consumer.cancelling():
+                raise
--- a/app/utils/aio.py
+++ b/app/utils/aio.py
@@ -3,9 +3,11 @@
 import asyncio
 import contextlib
 from collections.abc import AsyncGenerator
-from typing import TypeVar
+from typing import Any, Final, TypeVar
 
 T = TypeVar("T")
+
+_DONE: Final = object()
 
 
 async def prefetch(source: AsyncGenerator[T, None]) -> AsyncGenerator[T, None]:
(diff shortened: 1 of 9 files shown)


tests/unit/domain/references/test_repository.py
```diff
@@ -342,3 +349,33 @@ async def test_bulk_update_by_filter_allows_valid_transitions(
...
+
+        await repo.add_bulk_ignore_conflicts(records)
+
+        statements = [call.args[0] for call in mock_session.execute.await_args_list]
+        assert len(statements) > 1, "expected the insert to be chunked"
+
+        bound = 0
+        for statement in statements:
+            params = len(statement.compile(dialect=postgresql.dialect()).params)
+            assert params <= 2**15 - 1
```
mootpointer: This feels magic-numbery...

Adam-Hammo: It definitely is, but it doesn't exist anymore - the implementation moved on from the magic numbers in a later commit

app/utils/aio.py
```diff
@@ -0,0 +1,34 @@
...
+
+import asyncio
+import contextlib
+from collections.abc import AsyncGenerator
+from typing import TypeVar
+
+T = TypeVar("T")
+
+
+async def prefetch(source: AsyncGenerator[T, None]) -> AsyncGenerator[T, None]:
```
mootpointer, line 13: This is really neat. I wonder how much lookahead would be optimal (starting with 1 is fine!)

mootpointer (commented):
Nice improvements (but someone with developer in their title should approve the PR)

tests/unit/test_aio.py
```diff
@@ -0,0 +1,144 @@
...
+    assert elapsed < serial * 0.8
+
+
+@pytest.mark.asyncio
+async def test_prefetch_reads_at_most_one_ahead():
+    produced: list[int] = []
+    seen: list[int] = []
+
+    async for item in prefetch(_slow_source(6, 0, produced)):
+        await asyncio.sleep(0)
```
danielbreves: if the consumer was slower than the producer, the producer can put an item in the queue and fetch another from the source, so it would be 2 ahead at that point?

Adam-Hammo: Good point - adjusted the test and docstring instead of making it strictly 1-ahead (which is possible but more complex). This is potentially a nice way to absorb spiky latencies too

app/utils/aio.py
```diff
@@ -0,0 +1,49 @@
...
+        while True:
+            item = await queue.get()
+            if item is _DONE:
+                return
+            if isinstance(item, BaseException):
+                raise item
+            yield item
+    finally:
+        producer.cancel()
+        with contextlib.suppress(asyncio.CancelledError):
```
danielbreves: Could this suppress a cancellation of the worker task as well as the producer's?

app/utils/aio.py
```diff
@@ -0,0 +1,49 @@
...
+    the loop body - must wrap this in :func:`contextlib.aclosing`.
+    """
+    queue: asyncio.Queue[Any] = asyncio.Queue(maxsize=1)
+
+    async def _produce() -> None:
+        try:
+            async with contextlib.aclosing(source) as stream:
+                async for item in stream:
+                    await queue.put(item)
+        except Exception as exc:  # noqa: BLE001 - re-raised in the consumer
```
danielbreves: Should this be BaseException? Looks like [CancelledError](https://docs.python.org/3/library/asyncio-exceptions.html) is a subclass of BaseException.

app/utils/aio.py
Adam-Hammo: Good point - raising in that case.

app/utils/aio.py
Adam-Hammo: We _don't_ want to catch `CancelledError` here. If we've cancelled the stream, the consumer has possibly stopped reading too, so nothing empties the queue and we're stuck here trying to put the exception in it.

Updated the code to make that more explicit

danielbreves (approved):
This hurt my brain 😅 Brilliant work though!

PR #886, opened 2026-08-03
Add restricted ability for robots to write raw enhancements

### LINKED ISSUE #881: Allow entitled robots to write RawEnhancements to existing references
## Context

`RawEnhancement` writes are currently only allowed at import stage. Robots attempting to add one to an existing reference are rejected with `Robot returned illegal raw enhancement type`, and there's no path to attach a raw enhancement to an existing canonical outside imports.

This blocks two concrete use cases:

- **Geolocation robot** (external collaborator, PIK): runs Mordecai over reference abstracts and produces intermediate resolved-place data. The target LinkedData vocabulary isn't settled, and forcing the robot to commit to a final shape now is premature. Storing the raw output as a `RawEnhancement` and mapping later is the natural pattern.
- **GALENOS ingest** (#837): #836 established that 140 of 160 GALENOS studies already exist in the corpus. #837 currently takes Option A (import as new references and link duplicates) specifically because raw-on-existing isn't available; that route creates ~140 duplicate references that then need linking, versus the cleaner Option B of attaching the raw data directly to the existing canonicals. #837 deferred Option B pending a decision.

More broadly, a `RawEnhancement`-first pattern lets robots and importers contribute intermediate data now, deferring vocab and LinkedData composition to a later step. Raw fragments become a provenance trail; aggregation into LinkedData happens as separate work.

## Scope

- Allow `RawEnhancement` writes to existing references (new capability alongside the existing import-time path).
- Add a per-robot entitlement flag; unentitled callers continue to be rejected as they are now.

## Related

- Multi-source LDE coexistence in projection: #637
- GALENOS coverage analysis: #836
- GALENOS ingest (currently on Option A workaround): #837

Adam-Hammo:
Closes #881 

Loosens the existing block on robot-produced raw enhancements to be an entitlement check.

### DIFF
--- a/libs/sdk/src/destiny_sdk/robots.py
+++ b/libs/sdk/src/destiny_sdk/robots.py
@@ -401,6 +401,9 @@ class RobotEntitlement(StrEnum):
     FULL_TEXT = auto()
     """The robot is entitled to read full texts in order to derive enhancements."""
 
+    RAW_ENHANCEMENT_WRITER = auto()
+    """The robot is entitled to return raw enhancements."""
+
 
 class _RobotBase(BaseModel):
     """
--- a/.github/workflows/sdk-version-bump.yml
+++ b/.github/workflows/sdk-version-bump.yml
@@ -1,0 +1,59 @@
+name: SDK Version Bump
+
+on:
+  pull_request:
+
+jobs:
+  check-sdk-version-bumped:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          version: "0.9.22"
+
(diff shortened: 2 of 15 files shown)


libs/sdk/src/destiny_sdk/robots.py
```diff
@@ -401,6 +401,9 @@ class RobotEntitlement(StrEnum):
     FULL_TEXT = auto()
     """The robot is entitled to read full texts in order to derive enhancements."""
 
+    RAW_ENHANCEMENT_WRITER = auto()
```
danielbreves, line 404: Should we bump the SDK version to expose this?

Adam-Hammo: Yeah, probably. We should check this in linting tbh

Adam-Hammo: Ta-da!

danielbreves (approved):
Nice! LGTM!

PR #892, opened 2026-08-04
Remove uv.lock from the sdk

mootpointer:
We had added the uv.lock to the SDK folder and it has been maintained, but it is not used anywhere downstream. This is causing any security issues to be duplicated, even when dependencies are transient rather than direct.

We no longer track the lockfile, which should simplify things for us.


jack-o-the-green (approved):
LGTM

Issue #837, opened 2026-07-20
GALENOS: attach raw study data to existing DESTINY references


danielbreves:
## Context

GALENOS living systematic reviews extract structured data from each included study, currently stored as CSVs in GitHub. We're bringing that data into DESTINY as `RawEnhancement`s — essentially the JSON-ified CSV rows — so it can be served by the new data browser (#848).

#836 matched **~140 of the 160 studies to references that already exist** in DESTINY (mostly from the OpenAlex import), and #881 adds the ability to write a `RawEnhancement` onto an existing reference. For those ~140 we attach the study data straight onto the matching reference, instead of importing a duplicate and linking it — which is what this issue does. The remaining ~20 aren't in DESTINY yet and have open data-quality questions, handled in #882 (waiting on researcher feedback).

## Scope

#836 produced a map of each GALENOS study to the existing DESTINY reference (its canonical) it matches. This issue processes the matched studies:

- **Build a raw enhancement per study** — wrap the study's raw row(s) from its LSR CSV/xlsx, unchanged, as a `RawEnhancement`. Preserve the original column names (one entry per study arm) and enough provenance to trace each row back to its study and its LSR — the LSR number especially, since the GALENOS mapping is joined back on it later. Don't fold in the GALENOS ontology/mapping; that comes later.
- **Attach it to the matched reference** — write the `RawEnhancement` onto that reference and tag the reference with a `domain-inclusion/galenos` annotation.

`parsers/eppi_parser.py` and `libs/eppi_importer/` are a close template for building the enhancement and driving the write.

## Acceptance criteria

- Each matched study's reference carries its GALENOS `RawEnhancement` and a `domain-inclusion/galenos` annotation, with **no duplicate reference created**.
- The full set is returned by `GET /v1/references/search/?annotation=domain-inclusion/galenos`.
- Each `RawEnhancement`'s `data` preserves the original column names and the distinction between empty, `"NA"`, and not-collected values — the browser relies on it to reshape the data later.

> [!NOTE]
> Use a GALENOS-specific enhancement `source` (e.g. `galenos-lsr-export@<date>`), distinct from other batches, so the set can be filtered later — same reasoning as the EEF import (#752).

Adam-Hammo:
This is actioned in production. 148 studies mapped onto 146 destiny references (2x arms represented as individual studies in the LSRs). Attached are the remaining 12 to be clarified and/or imported.

[...]

Issue #836, opened 2026-07-20
GALENOS: match included studies to DESTINY references (coverage analysis)


danielbreves:
Establish which GALENOS LSR studies already exist in the DESTINY corpus and produce the verified `study → reference` mapping the raw import depends on. Run the analysis **read-only, before any import**, so the duplicate rate can inform the ingest issue's approach and sizing.

**Why not just trust DOIs:** studies mostly carry a DOI, but in an untrusted, messy form — LSR1's `DOI_or_URL` mixes bare DOIs (`10.1097/…`) and DOI URLs (`https://doi.org/…`), DOIs can be wrong, and the other available fields vary per LSR. So match on **DOI as the primary key, corroborated against per-LSR data**, validating every match on **first-author surname + year (+ title)**.

### Suggested Scope
1. **Normalise + inventory per LSR** — extract/normalise the DOI (DESTINY's `DOIIdentifier` strips URL prefixes); catalogue the corroboration fields each LSR actually provides (e.g. LSR1 `studlab` "Agosti (1991)" → surname+year, `year`; LSR3 `.bib` author/year/title), supplementing from the repo bibliographies (EPPI export CSV / `.bib`). E.g.
    - https://github.com/galenos-project/LSR3_taar1_H/blob/main/lsr3_references.bib
    - https://github.com/galenos-project/LSR1_anhedonia_H/blob/main/data/human/LSR1%20-%20EPPI%20export%20-%2020240131.csv
    - https://github.com/galenos-project/LSR6_CBM/blob/main/data/eppi_export_070525.xlsx
3. **DOI lookup + corroboration (read-only)** — `GET /v1/references/?identifier=doi:<doi>`; accept only if surname + year (ideally title) agree, else flag a *suspect DOI* and fall through to step 3.
4. **Bibliographic search for the rest (read-only)** — title + year (+ author) search (`GET /v1/references/search/`), writing top-N candidates to a human-review file.
5. **Coverage report** — classify every study: *already-in-corpus* / *suspect-DOI* / *not-found* (→ would become a new canonical). **This duplicate rate feeds the ingest issue.**
This issue is **read-only** — it produces the map and does not write to the repository. #837 uses the map to attach the GALENOS raw data to the matched references; the ~20 studies that didn't match are handled in #882.

[...]

PR #895, opened 2026-08-05
Add static enhancement CLI utility

Adam-Hammo:
CLI utility to support #837 

Allows the direct upload of static enhancements to specified references in special circumstances.

Will be used with the references found in #836 to fulfil #837.

The script:
- takes in either a list of enhancements with reference ids on them, or a list of reference ids with a single enhancement to add to all references.
- either accepts an existing robot, or creates one for the run. in both cases the owner must be `cli/add_static_enhancements` as a safety guard
  - cycles the secret - the owner guard makes this safe and means we don't need to be copy-pasting secrets
- requests the enhancements
- fulfills the enhancements

### DIFF
--- a/cli/add_static_enhancements.py
+++ b/cli/add_static_enhancements.py
@@ -1,0 +1,418 @@
+r"""
+A utility to add a static list of enhancements to references.
+
+This utility:
+- Accepts either:
+    - a .jsonl file of `Enhancement`s
+    - a text file of reference ids and a single
+      `destiny_sdk.enhancements.EnhancementFileInput` to apply to each of them
+- Requests enhancements from the repository for each unique reference_id
+- Acts as a robot to add the enhancements to the references in the repository
+- Prints the results of the original enhancement request
+
+Add an enhancement per reference, with a robot to write them::
+
+    uv run python -m cli.add_static_enhancements --env staging \
+        --enhancement-file enhancements.jsonl \
+        --create-robot 'your-project-static-enhancement-robot'
+
+Tag those same references, reusing a robot::
+
+    uv run python -m cli.add_static_enhancements --env staging \
+        --reference-id-file references.txt \
+        --enhancement '{"source": "your-project-initial-hydration",
+            "visibility": "public", "content": {"enhancement_type": "annotation",
+            "annotations": [{"annotation_type": "boolean",
+            "scheme": "domain-inclusion", "label": "your-project", "value": true}]}}' \
+        --robot-id 019ec8b0-9173-76f0-9d09-b7bdf4e31047
+
+The caller must have robot.writer and enhancement_request.writer permissions.
+"""
+
+# ruff: noqa: T201
+import argparse
+import json
+import sys
+import time
+from collections.abc import Iterable
+from pathlib import Path
+from uuid import UUID
+
+import httpx
+from destiny_sdk.client import RobotClient
+from destiny_sdk.enhancements import Enhancement, EnhancementFileInput, EnhancementType
+from destiny_sdk.robots import (
+    EnhancementRequestIn,
+    EnhancementRequestRead,
+    EnhancementRequestStatus,
+    ProvisionedRobot,
+    Robot,
+    RobotEnhancementBatch,
+    RobotEnhancementBatchResult,
+    RobotEntitlement,
+    RobotIn,
+)
+from pydantic import HttpUrl, ValidationError
+
+from cli.client import ApiArgumentParser
+from cli.register_robot import register_robot
+
+# Marks robots provisioned by this utility. Only robots with this owner are allowed to
+# be represented by this utility.
+ROBOT_OWNER_MARKER = "cli/add_static_enhancements"
+
+ROBOT_DESCRIPTION = (
+    "Writes static enhancements supplied by cli.add_static_enhancements. Its secret "
+    "is cycled on each run and it does not poll for work of its own."
+)
+
+BATCH_SIZE = 10_000
+
+
+def enhancement_input(value: str) -> EnhancementFileInput:
+    """Parse an `EnhancementFileInput` passed as a json string."""
+    try:
+        return EnhancementFileInput.model_validate_json(value)
+    except ValidationError as exc:
+        raise argparse.ArgumentTypeError(str(exc)) from exc
+
+
+def read_reference_ids(file: Path) -> set[UUID]:
+    """Read reference ids from a text file, one per line."""
+    return {
+        UUID(line.strip()) for line in file.read_text().splitlines() if line.strip()
+    }
+
+
+def load_enhancements(
+    enhancement_file: Path | None,
+    reference_id_file: Path | None,
+    enhancement: EnhancementFileInput | None,
+) -> dict[UUID, Enhancement]:
+    """Build the enhancement to add for each reference, keyed by reference id."""
+    if enhancement_file:
+        enhancements = [
+            Enhancement.from_jsonl(line)
+            for line in enhancement_file.read_text().splitlines()
+            if line.strip()
+        ]
+    else:
+        enhancements = [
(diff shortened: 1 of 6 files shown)


libs/sdk/src/destiny_sdk/client.py
```diff
@@ -202,7 +202,7 @@ def renew_robot_enhancement_batch_lease(
             use a default lease duration.
         :type lease_duration: str | None
         """
-        response = self.session.post(
+        response = self.session.patch(
```
Adam-Hammo, line 205: Wee bug I found

cli/add_static_enhancements.py
```diff
@@ -0,0 +1,418 @@
...
+
+    uv run python -m cli.add_static_enhancements --env staging \
+        --reference-id-file references.txt \
+        --enhancement '{"source": "your-project-initial-hydration",
+            "visibility": "public", "content": {"enhancement_type": "annotation",
+            "annotations": [{"annotation_type": "boolean",
+            "scheme": "domain-inclusion", "label": "your-project", "value": true}]}}' \
+        --robot-id 019ec8b0-9173-76f0-9d09-b7bdf4e31047
+
+The caller must have robot.writer and enhancement_request.writer permissions.
```
jack-o-the-green: I believe the caller must also have `robot.entitlement.writer` in the case that they want to add raw enhancements using this script to a robot they create. 

I think this script would also allow anyone to add raw enhancements if they reused a robot that already had that entitlement that was created by this utility (even if they don't have robot entitlement writing permissions), which we might consider a security gap (though requires quite a knowledge of internals). 

Furthermore, changing a CLI script locally is easy, so anyone could remove the ownership check when reusing robots to allow the use of robots not created by this utility, which would be more likely to have entitlements. Feels more like a guardrail on the script than actual scrutiny. I believe we could get scenarios where someone looks like they're acting as the vocabulary-mapping-robot (for example), or even rotate the vocabulary-mapping-robot secret.

I am wondering if we want to limit access to this process further as it seems quite powerful?

Some future thinking as well, we might want to be able to remove robots from the pool of available robots eventually, unsure if we want the ones created by this process to be long lived.

cli/add_static_enhancements.py
```diff
@@ -0,0 +1,418 @@
...
+
+def resolve_robot(
+    client: httpx.Client,
+    robot_id: UUID | None,
+    name: str | None,
+    entitlements: set[RobotEntitlement],
+) -> ProvisionedRobot:
+    """Provision a bespoke robot, or reuse one this utility provisioned earlier."""
+    if name:
+        robot = register_robot(
```
jack-o-the-green: Are we happy with the error thrown here if the user is not entitled to create the robot entitlements? https://github.com/destiny-evidence/destiny-repository/blob/main/app/domain/robots/services/access_control_service.py#L10.

tests/cli/test_add_static_enhancements.py
```diff
@@ -0,0 +1,219 @@
...
+        EnhancementFileInput(
+            source=SOURCE, visibility=Visibility.PUBLIC, content=ANNOTATION
+        ),
+    )
+
+    assert enhancements.keys() == {first, second}
+    assert {enhancement.source for enhancement in enhancements.values()} == {SOURCE}
+
+
+def test_two_enhancements_for_one_reference_are_rejected(tmp_path: Path) -> None:
```
jack-o-the-green, line 100: Confirming we do not consider a canonical and a duplicate the same reference for this utility (I don't think we should).

cli/add_static_enhancements.py
```diff
@@ -0,0 +1,418 @@
...
+    try:
+        enhancements = load_enhancements(
+            enhancement_file=args.enhancement_file,
+            reference_id_file=args.reference_id_file,
+            enhancement=args.enhancement,
+        )
+        entitlements = required_entitlements(enhancements.values())
+        print(f"{len(enhancements)} enhancements to add.")
+
+        if args.dry_run:
```
jack-o-the-green: Ideally a dry run would also check robot resolution if an id is passed, or indicate intention to create a robot with any entitlements.

cli/add_static_enhancements.py
```diff
@@ -0,0 +1,418 @@
...
+
+
+def poll_until_complete(
+    client: httpx.Client,
+    request_id: UUID,
+    poll_interval: float = 5,
+) -> EnhancementRequestRead:
+    """Poll the enhancement request until it stops progressing."""
+    print(f"Polling enhancement request {request_id}...")
+    while True:
```
jack-o-the-green: I think this will poll forever if there is somehow a glitch in the enhancement request moving to a terminal state, add a timeout at some point?

cli/add_static_enhancements.py
```diff
@@ -0,0 +1,418 @@
...
+# Marks robots provisioned by this utility. Only robots with this owner are allowed to
+# be represented by this utility.
+ROBOT_OWNER_MARKER = "cli/add_static_enhancements"
+
+ROBOT_DESCRIPTION = (
+    "Writes static enhancements supplied by cli.add_static_enhancements. Its secret "
+    "is cycled on each run and it does not poll for work of its own."
+)
+
+BATCH_SIZE = 10_000
```
jack-o-the-green, line 73: Do we also want to limit the number of references that can go into a single enhancement request with this process? I think probably not.

cli/add_static_enhancements.py
```diff
@@ -0,0 +1,418 @@
...
+        )
+        raise ValueError(msg)
+    if missing := entitlements - existing.entitlements:
+        msg = (
+            f"Robot {existing.id} is not entitled to {', '.join(sorted(missing))}, so "
+            "it cannot write these enhancements."
+        )
+        raise ValueError(msg)
+
+    # No secret is passed in, so take a fresh one. The owner marker makes this safe.
```
jack-o-the-green, line 176: Not safe if I (as the script runner) just remove the owner marker, though I would need robot.writer perms to do this.

cli/add_static_enhancements.py
```diff
@@ -0,0 +1,418 @@
...
+            UUID(json.loads(line)["id"])
+            for line in response.iter_lines()
+            if line.strip()
+        }
+
+
+def poll_until_complete(
+    client: httpx.Client,
+    request_id: UUID,
+    poll_interval: float = 5,
```
jack-o-the-green: Maybe 10s poll interval? I always found 5s quite small

jack-o-the-green (commented):
Great approach, basically a robot mocker and I love it. Lots of access comments from me.

cli/add_static_enhancements.py
Adam-Hammo: Yeah my thinking was that would suffice, and if it doesn't, the API should probably be better

cli/add_static_enhancements.py
Adam-Hammo: As discussed, in general we'll seek to close this gap with better permissioning - the gap exists to those with `robot.writer` in production, which should be a very select group of people.

I'll update the docstring to mention entitlements however

tests/cli/test_add_static_enhancements.py
Adam-Hammo: Correct

cli/add_static_enhancements.py
Adam-Hammo: Good points, will do

cli/add_static_enhancements.py
Adam-Hammo: Same as above - in the first instance, we'll limit the number of permissioned folks to those that are very trusted (or whose jobs are on the line)

cli/add_static_enhancements.py
Adam-Hammo: I agree no limit, but maybe we should batch the requests also. Massive requests would just take a while at the API level and risk timing out on the httpx client.

jack-o-the-green (approved):
LGTM

PR #897, opened 2026-08-06
Add count of mapped references to api facet count response

### LINKED ISSUE #896: Show unique reference count in top left of evidence map
## Outcome
In the top right corner of an evidence map, users see the 'Total unique investigations x' that are displayed on the map

## Context
Feedback from Ruth:
> "I have visualised a map with 'Thematic focus - primary' as the columns, and 'Fragility and Humanitarian contexts' as the rows. I have generated the map (screenshot below) and it highlights that 1961 studies are included (top left box of the map), but I know that is how many studies are in the repo, so I'm confused about how many papers are in the map. Can this info be indicated somewhere - at the moment it is misleading as it appears to a new user that there are 1961 studies in the map."

After speaking with Britta she confirmed that a user expected to see the total count of unique records that are displayed on the evidence map, not the total number of records that were included with the filters.

## Scope
- update the figure to be the total number of unique records included in the map
- update the text to read 'Total unique investigations x' where x is the number of unique investigations

## Acceptance criteria
- When the count and the text are updated as per the scope above

## Related
n/a

Adam-Hammo:
Closes #896 

Returns the total number of references on the map alongside the total number of references matching the search.

Somewhat back-compatible - the `total` field stays, but becomes `total.mapped` instead of `total.searched`. The only "breaking" consumer is evidence repository and this break will bring it in line with the goal of the ticket on API release.


jaybea (commented):
The evidence-repo-ui control flow pays attention to `total`, which is now mapped :) to `mapped`. So you can have 0 refs intersecting on the selected axis, so the user is told they've overfiltered and to reset their search, which is actually showing 1000's of results.

Is there some UI change I'm missing? Or are we just going to accept this case?

Image

Adam-Hammo:
@jaybea the corresponding UI change: https://github.com/destiny-evidence/evidence-repo-ui/pull/238

Adam-Hammo:
> Since the new SDK requires totals and cannot parse a response from the older server, should this be a minor version bump—perhaps 0.15.0—rather than a patch bump?

Yes, this is fair. We "know" the UI is our only consumer but that's no reason to stray from proper versioning! Thank you.

> It looks like the repository layer now decides what “included in the evidence map” means. Would it be better for the search service to resolve that domain behaviour, leaving the repository to translate the resolved constraints into Elasticsearch operations?

Good question! I think that could potentially come at the risk of leaking ES implementation details up into the domain, whereas to me the repository exists to translate domain requests into persistence operations. The domain decision kind of already sits in `SearchService._resolve_cross_facet_axis` - that's what determines each axis's `facet_type` and `include` set, and the resolved `CrossFacetAxis` is the constraint. There was definitely some proliferation of terms though, I've renamed `in_map` to `both_axes_present`.

Do you have a specific example of what you were thinking? I might have misunderstood.

> Does the bucket guard need to count the new in_map filter bucket? The current calculation is a + a*b, but the aggregation now creates 1 + a + a*b. At a long-shot (!), that could request 65,537 buckets, which I think is one more than the Elasticsearch default limit.

I just tested this - it's quite funny behaviour. I successfully created 65,537 buckets! I think Elasticsearch counts the buckets in a multi-bucket agg and ignores the single bucket aggregations, so the current implementation is safe.

jaybea:
> Good question! I think that could potentially come at the risk of leaking ES implementation details up into the domain, whereas to me the repository exists to translate domain requests into persistence operations. The domain decision kind of already sits in `SearchService._resolve_cross_facet_axis` - that's what determines each axis's `facet_type` and `include` set, and the resolved `CrossFacetAxis` is the constraint. There was definitely some proliferation of terms though, I've renamed `in_map` to `both_axes_present`.

I like the rename. You're right, I guess I was getting a bit grandiose, imagining the search service resolving each axis into some sort of persistence independent constraint, but an intermediary class doesn't make sense here. I'm especially going to take on your  characterisation of the repository as translating domain request into persistence ops, make total sense. 
 
> I just tested this - it's quite funny behaviour. I successfully created 65,537 buckets! I think Elasticsearch counts the buckets in a multi-bucket agg and ignores the single bucket aggregations, so the current implementation is safe.

Cool, not a problem then!

jaybea (approved):
Looks good!

Issue #887, opened 2026-08-03
Run the production-like evaluation with the released Deduper


jaybea:
Part of #791. Needs the adapter from #944, and a local corpus that holds the answers.

## Outcome

Publish one production-like evaluation run using the released Deduper, so #877 can set the first automatic policy on measured evidence.

## Scope

Run the evaluation offline, against a rebuilt local corpus holding only the held side of the labelled OpenAlex set, then import the incoming side through the normal path with shadow retrieval and assessment enabled. Record every assessment under purpose `performance_shadow`. The evidence is the `deduplication_assessment` record table, queryable and covering 100% of assessments, not a blob artifact bundle.

This runs the same code that later runs in production, so the measurement transfers. It also avoids the input-shape ceiling: a locally held reference is hydrated through `SCORED_ENHANCEMENT_TYPES` exactly as production hydrates it, so venue and pagination reach the scorer rather than being rebuilt from a thin row.

State the adjudication method next to every figure. The labels are **silver, deterministic**, from landing-page identifier and ORCID evidence, and are **not human-adjudicated**. Whether that is strong enough to authorise automatic merging is a judgement #877 has to make explicitly, not inherit.

Include records with no publication year, which production never retrieves today, so #877 can judge a year-optional policy on measured results. This is a reason to run offline rather than on production traffic: production applies one configured retrieval policy, and locally the policy can be varied.

Self-matching does not arise on this route and needs no exclusion logic. The rebuilt baseline deliberately holds only the held side, and the incoming side is imported afterwards, so each incoming record carries its own corpus UUID and the existing UUID-keyed self-exclusion removes it from its own candidates. Report the corpus composition so a reader can confirm the two sides were disjoint.

The run writes no duplicate decision, changes nothing about which reference is canonical in Elasticsearch, and starts no background work. #922 and #947 carry the public-contract tests that prove the shadow path cannot reach an applier; this issue does not re-prove that, but it does confirm on its own run that no duplicate decision rows were written for the incoming set.

## Complete when

A production-like evaluation run is published, linked from this issue, and #877 has the evidence it needs, with the adjudication method stated alongside every figure.

The published run reports the corpus composition and confirms no duplicate decisions were written for the incoming set.

Issue #874, opened 2026-07-30
Build the deduplication assessment path


jaybea:
Part of #791. Ready now: everything here can be built and tested without the Deduper. The runner that drives it over a supplied record set is #915; swapping in the released Deduper and publishing the evaluation run is #887.

## Outcome

Replace the deduplication placeholder with the real assessment path: find candidates, fetch their stored identifiers and bibliographic enhancements needed by the scorer, score every pair, and produce one assessment for the incoming reference. The same code serves the evaluation run and, later, the live path.

## The boundary that makes this safe

Two operations, not one mode flag.

`evaluate` finds candidates, fetches their stored identifiers and bibliographic enhancements needed by the scorer, scores each pair and returns an assessment. It reads Postgres and Elasticsearch, and it has no access to the decision mapper, so it cannot write a duplicate decision, change which reference is canonical in Elasticsearch, or start any background work.

`apply` is the only writer, and it belongs to #875.

A flag around today's `process_reference_duplicate_decision` would not achieve this. Candidate selection updates the decision row four times and determination once more, all before the mapper runs, so that intermediate state has to become an in-memory assessment first. #876 makes candidate selection stop writing interim results.

Whether a run evaluates or applies is a property of the run, fixed when it starts. It is not ambient configuration that could turn a reporting run into a writing one part-way through.

Consume the complete candidate union from #876. Hydrate and score every candidate under one assessment, carrying its identifier and retrieval routes as provenance. Retrieval route gives a candidate no scoring or decision priority. Do not filter or reorder candidates by community membership. Record whether the Elasticsearch route was available so #877 can distinguish an evaluable no-match from an input that could not be searched.

Summarise the complete scored candidate set without making a persisted determination. An unscorable pair prevents a proposal even when another candidate clears the threshold. An input without enough bibliographic fields for Elasticsearch search also prevents a no-match result from being treated as evidence that the incoming reference is canonical.

| Assessment evidence | Assessment result |
|---|---|
| Every pair scored, none cleared, and the input was searchable | `PROPOSE_CANONICAL` |
| Every pair scored and exactly one candidate cleared | `PROPOSE_DUPLICATE`, with that candidate in `proposed_duplicate_of_id` |
| More than one candidate cleared, any pair was unscorable, or a no-match input was unsearchable | `NO_PROPOSAL`, retaining the relevant candidate IDs |

These are read-only proposals. #875 decides whether an assessment is sufficient to write a determination or warrants review.

[...]

PR #905, opened 2026-08-06
Select candidates by configured policy, without interim writes

### LINKED ISSUE #876: Select candidates by configured policy, without writing
Part of #791. Everything in this issue can land now. It configures candidate selection but does not enable automatic decisions; that happens in #885 after #887's evaluation and #877's policy decision.

## Outcome

Switch automatic candidate selection from its hard-coded fuzzy title, author and year query to the evaluated candidate retrieval policy selected by #819, chosen through configuration.

The selected first automatic shape has identifier union enabled and uses soft publication-year decay, non-fuzzy title matching, a publication-year requirement for the Elasticsearch route, and `k=10`, as decided in [#819's closeout](https://github.com/destiny-evidence/destiny-repository/issues/819#issuecomment-5099312229). It is parameter-identical to `soft_year_decay_nonfuzzy_probe_v1`, but that name is reserved for evaluation, so the production policy needs a stable name.

This is the query-policy switch only. Switching the trusted-identifier shortcut off and enabling automatic merging both happen at the turn-on, not here.

## Scope

Build the candidate set as the union of the top `k=10` Elasticsearch candidate canonicals and any held references found by exact identifier lookup, resolved to their current canonicals. Wait for both routes, deduplicate the complete union by canonical ID and retain every route that surfaced each candidate. For example, if an incoming OpenAlex dataset sourced from DataCite lacks authors but its DOI matches a held reference, add that reference's current canonical to the union even though the Elasticsearch route cannot run. A canonical found by both DOI and Elasticsearch appears once with both routes recorded. An identifier match nominates a candidate; it does not select the winner or create a duplicate decision.

Keep `k=10` for the first policy: evaluation found the useful retrieval knee there. Do not exclude community-bound canonicals or rank them after other candidates. They are valid targets for a new non-community reference, so filtering them would reduce recall and could hide the best match. Community protection is applied to the incoming reference after scoring, under #877 and #875.

- Give the selected shape a stable production policy name.
- Make candidate selection resolve that named policy, and the number of candidates it returns, through configuration.
- Make the configured policy reversible to `current_fuzzy_v1` without a code change.
- Fix the eligibility gate. Which incoming references we search is hard-coded to title, authors and year, ignoring the policy's own `requires_publication_year`, so a year-optional policy still cannot see year-less records. `RetrievalPolicy.is_input_searchable` already implements this and is unused on this path.
- Add the identifier candidate route, so the automatic path honours `union_identifiers` as the Candidate Selection API already does. It has no such route today, so a record that fails the searchability gate gets no candidates at all, even where its exact identifiers match. This is not the trusted-identifier shortcut, which makes a terminal decision rather than offering a candidate to score, and it is what preserves same-identifier matching when that shortcut is switched off at the turn-on.
- Make candidate selection not write interim results. Return the candidate identifiers, every route that surfaced each one and the policy that produced them; persist none of it here. Today it updates the decision row four times and sets `NOMINATED`, a state nothing reads and the same transaction overwrites. Pass the candidate set to #874 as part of the assessment. Candidate selection returns its policy provenance but does not persist it. #875 records the policy generation, retrieval policy name, configured `k` and actual candidate count on every terminal decision; assessments retained for review keep the same policy provenance. Until then, #874 cannot use the same path without also writing decisions.

## Notes

Candidate selection is disabled in production, which is how the #640 Elasticsearch timeouts were mitigated, so re-enabling it needs both this configured policy and the feature flag. Non-fuzzy title matching measures cheaper on retrieval, which is most of the per-record cost.

Publication year is required only by the Elasticsearch route in the first automatic policy. Keep `soft_year_decay_year_optional_v1` available to #874's write-free runner so yearless references continue through retrieval and Deduper scoring, with the results retained to inform sparse-record recalibration. This path writes no duplicate decisions. On the live first-policy path, a yearless reference still reaches scoring when exact identifier lookup nominates a candidate.

## Complete when

A test shows automatic candidate selection and the Candidate Selection API resolve the same query and eligibility rules for a configured policy name, including that policy's own year requirement, and reverting to `current_fuzzy_v1` needs no code change.

Tests show that identifier and Elasticsearch results are united and deduplicated by canonical ID, the union can exceed ten candidates, and community-bound canonicals remain eligible.

## Out of scope

- Choosing a year-optional policy. Honouring the configured policy is in scope; selecting one is not.
- Mapping or analyzer changes requiring a reindex.
- Deduper calibration and automatic merging.
- Switching the shortcut off, which happens at the turn-on, and deleting its code (#650).

jaybea:
Closes #876

## Outcome

Candidate selection now uses the configured production retrieval policy and returns the complete, non-persisted candidate set required by later evaluation and scoring work.

The Candidate Selection API and automatic path share the same policy resolution, eligibility rules, identifier union and Elasticsearch retrieval behaviour.

## What changed

- Added the stable `candidate_selection_v1` production policy, matching [#819's selected retrieval shape](https://github.com/destiny-evidence/destiny-repository/issues/819#issuecomment-5099312229):
  - soft publication-year decay
  - non-fuzzy title matching
  - publication year required for the Elasticsearch route
  - exact identifier union enabled
- Made the default retrieval policy and candidate count configurable through `DEDUP_SCORING`.
- Set the defaults to `candidate_selection_v1` and `k=10`.
- Kept rollback to `current_fuzzy_v1` available through configuration alone.
- Made automatic candidate selection return `CandidateSelectionResult`, including candidate routes, retrieval policy, index version, input searchability and retrieval diagnostics.
- Removed the interim `NOMINATED` decision and candidate-ID writes.
- Unified Elasticsearch and exact-identifier candidates by canonical ID while retaining every matching route. The resulting union may exceed `k`.
- Kept community-bound canonicals eligible for retrieval.
- Kept exact identifier matching available when the input cannot use the Elasticsearch route.
- Kept exact Elasticsearch hit counts enabled for retrieval diagnostics.
- Passed the complete candidate-selection result into the temporary determination step.
- Confined placeholder determination behaviour to tests. Non-test environments remain `UNSEARCHABLE` until the deep-deduplication assessment path is enabled.
- When candidate search is disabled, the existing decision is marked `UNSEARCHABLE` without running retrieval.

The trusted OpenAlex identifier shortcut remains enabled and unchanged. This PR does not enable automatic merging or switch off that shortcut.

The placeholder determinator is deliberately temporary. #874 replaces it with the write-free deep-deduplication assessment path using a stand-in scorer; #887 later integrates the released Deduper.

## Verification

- Full default suite: 1,041 passed, 1 intentionally skipped
- Candidate Selection end-to-end tests: 9 passed
- Changed-file linting, formatting and type checks passed

The skipped test is a long-running 100,000-reference SQL performance test that is deliberately excluded from regular test runs.

### DIFF
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -400,7 +400,7 @@ class DedupCandidateScoringConfig(BaseModel):
         le=1000,
         description="Default number of candidate references to retrieve.",
     )
-    retrieval_policy: RetrievalPolicyName = Field(
+    default_retrieval_policy: RetrievalPolicyName = Field(
         default=RetrievalPolicyName.CANDIDATE_SELECTION_V1,
         description="Candidate retrieval policy used when no override is supplied.",
     )
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -19,7 +19,10 @@ from pydantic import (
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 from app.core.telemetry.logger import get_logger
-from app.domain.references.models.models import ExternalIdentifierType
+from app.domain.references.models.models import (
+    ExternalIdentifierType,
+    RetrievalPolicyName,
+)
 from app.persistence.blob.models import BlobContainer, BlobStorageLocation
 from app.utils.time_and_date import iso8601_duration_adapter
 
@@ -392,10 +395,14 @@ class DedupCandidateScoringConfig(BaseModel):
         description="Minimum token length for author name matching.",
     )
     candidate_k: int = Field(
-        default=100,
+        default=10,
         ge=1,
         le=1000,
         description="Default number of candidate references to retrieve.",
+    )
+    retrieval_policy: RetrievalPolicyName = Field(
+        default=RetrievalPolicyName.CANDIDATE_SELECTION_V1,
+        description="Candidate retrieval policy used when no override is supplied.",
     )
 
 
--- a/app/domain/references/service.py
+++ b/app/domain/references/service.py
@@ -1289,18 +1289,25 @@ class ReferenceService(GenericService[ReferenceAntiCorruptionService]):
                     )
                 return
 
-        # Carry on with normal processing
-        reference_duplicate_decision = (
-            await self._deduplication_service.nominate_candidate_canonicals(
-                reference_duplicate_decision
+        if settings.feature_flags.enable_canonical_candidate_search:
+            candidate_selection = (
+                await self._deduplication_service.select_candidate_canonicals(
+                    reference_duplicate_decision.reference_id
+                )
             )
-        )
-
-        reference_duplicate_decision = (
-            await self._deduplication_service.determine_canonical_from_candidates(
-                reference_duplicate_decision
+            reference_duplicate_decision = (
+                await self._deduplication_service.determine_canonical_from_candidates(
+                    reference_duplicate_decision,
+                    candidate_selection,
+                )
             )
-        )
+        else:
+            reference_duplicate_decision = reference_duplicate_decision.model_copy(
+                update={
+                    "duplicate_determination": DuplicateDetermination.UNSEARCHABLE,
+                    "detail": "Canonical candidate search is disabled.",
+                }
+            )
 
         (
             reference_duplicate_decision,
--- a/app/domain/references/service.py
+++ b/app/domain/references/service.py
@@ -1302,11 +1302,11 @@ class ReferenceService(GenericService[ReferenceAntiCorruptionService]):
                 )
             )
         else:
-            reference_duplicate_decision = reference_duplicate_decision.model_copy(
-                update={
-                    "duplicate_determination": DuplicateDetermination.UNSEARCHABLE,
-                    "detail": "Canonical candidate search is disabled.",
-                }
+            reference_duplicate_decision.duplicate_determination = (
+                DuplicateDetermination.UNSEARCHABLE
+            )
+            reference_duplicate_decision.detail = (
+                "Canonical candidate search is disabled."
             )
 
         (
--- a/app/domain/references/services/deduplication_service.py
+++ b/app/domain/references/services/deduplication_service.py
@@ -28,7 +28,6 @@ from app.domain.references.models.models import (
     Reference,
     ReferenceDuplicateDecision,
     ReferenceDuplicateDeterminationResult,
-    RetrievalPolicyName,
     YearDecay,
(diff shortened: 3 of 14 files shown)


app/core/config.py
```diff
@@ -392,11 +395,15 @@ class DedupCandidateScoringConfig(BaseModel):
         description="Minimum token length for author name matching.",
     )
     candidate_k: int = Field(
-        default=100,
+        default=10,
         ge=1,
         le=1000,
         description="Default number of candidate references to retrieve.",
     )
+    retrieval_policy: RetrievalPolicyName = Field(
```
Adam-Hammo: Nit: can we maybe call `default_retrieval_policy` to make it clear it's only used when not otherwise specified?

app/domain/references/service.py
```diff
@@ -1289,18 +1289,25 @@ async def process_reference_duplicate_decision(
...
+                    reference_duplicate_decision,
+                    candidate_selection,
+                )
+            )
+        else:
+            reference_duplicate_decision = reference_duplicate_decision.model_copy(
+                update={
+                    "duplicate_determination": DuplicateDetermination.UNSEARCHABLE,
+                    "detail": "Canonical candidate search is disabled.",
+                }
```
Adam-Hammo: Why can we not just update this instance in-place if we overwrite it anyway?

app/domain/references/services/deduplication_service.py
```diff
@@ -521,119 +511,61 @@ async def register_duplicate_decision_for_reference(
...
-            )
-        else:
-            # Is there a search result score that would be enough for us to mark as
-            # duplicate without proceeding to the next step?
-            reference_duplicate_decision = (
-                await self.sql_uow.reference_duplicate_decisions.update_by_pk(
-                    reference_duplicate_decision.id,
-                    candidate_canonical_ids=[
-                        result.id for result in search_result.hits
-                    ],
```
Adam-Hammo, line 592: Is it worth us storing a distilled version (or all of) the `CandidateSelectionResult`? I don't really know what we'd use it for but it just feels like such rich data to be throwing away in the GC cycle

Adam-Hammo (approved):
Looking good! Two questions and a mega-nit-pick, pre-approving.

app/core/config.py
jaybea: I like it.

app/domain/references/service.py
jaybea: You're right.

PR #907, opened 2026-08-06
Add reference migrator CLI utility

Adam-Hammo:
Adds CLI utility to migrate a set of references between environments (but never to production).

See docstring for more.

Required for #837 in order to hydrate staging with references to attach enhancements too (they all exist solely in production). Likely has many other uses!

A future implementer could/should expand this to a search interface. I'd be interested in looking at an easy way to flip _all_ CLI scripts between id-list and search-query inputs.

### DIFF
--- a/cli/migrate_references.py
+++ b/cli/migrate_references.py
@@ -1,0 +1,214 @@
+"""
+A utility to copy references from one environment into another.
+
+Actions:
+1. Export the references from the source, in chunks
+2. Register each export file as a batch of one import record in the destination
+3. Finalise the record and summarise each batch
+
+Export files are passed on by URL, never rewritten.
+
+*NOTE* This is lossy:
+- ``derived_from``, ``robot_version``, all ``id``s and all timestamps are dropped
+- duplicate clusters collapse into one reference holding the union of their content
+- the destination re-runs deduplication over its own corpus
+
+*BONUS*: It's totally valid to have a local `dest` if you want some remote references!
+"""
+
+# ruff: noqa: T201
+import sys
+import time
+from typing import TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


cli/migrate_references.py
```diff
@@ -0,0 +1,214 @@
...
+import time
+from typing import TYPE_CHECKING
+
+import httpx
+from destiny_sdk.references import ExportStatus, ReferenceExportRead
+
+from app.core.config import Environment
+from app.utils.lists import list_chunker
+from cli.client import ApiArgumentParser, get_client
+from cli.post_import_file import (
```
jack-o-the-green, line 30: Conscious that we're starting to get dependencies between cli scripts now, feels like some of these functions should be migrated to the sdk. Not scope for this PR but might be good to raise a tech debt ticket for?

cli/migrate_references.py
```diff
@@ -0,0 +1,214 @@
...
+    )
+    parser.add_argument(
+        "--notes",
+        default=None,
+        help="Notes for the destination import record. Defaults to a summary.",
+    )
+    parser.add_argument(
+        "--poll-interval",
+        type=float,
+        default=5,
```
jack-o-the-green: `nitpick` increase to 10s?

jack-o-the-green (approved):
LGTM and love that we can do this locally!

cli/migrate_references.py
Adam-Hammo: Yeah really good point, would be nice to have the core functionality in the SDK client instead. 
Ticketed here: #917. I haven't called it debt (yet) as it's a genuine user improvement too.

Issue #771, opened 2026-07-02
Fix truncation in HPV abstracts and titles by assigning canonicals with better bibliographic information


jack-o-the-green:
### Context 
There are 108 references in the HPV that have a truncated title, a truncated abstract, or both. Truncation is most frequently present in abstracts.

Example: [019ec892-1de7-7717-9b7d-159c4b454cbe](https://data.evidence-repository.org/hpv/references/019ec892-1de7-7717-9b7d-159c4b454cbe).

We want to remove the surfacing of this truncated bibliographic information to the user by assigning these references as duplicates of references that have better bibliographic information, as the canonical bibliographic information is what is surfaced to the user. 

### Identifying all truncated HPV references 
The following elasticsearch query will get you all 108 references with truncated titles/abstracts. It's important the term filter runs as a `post_filter` because it's doing a lot of lifting, this query won't work across our fill reference set which is why its domain bounded. 

```
GET /reference/_search
{
  "size": 2000,
  "_source": ["title", "abstract"],
  "runtime_mappings": {
    "title_lead": {
      "type": "boolean",
      "script": {
        "source": "String t = params._source.title; if (t == null) { emit(false); } else { t = t.trim(); emit(t.startsWith('…') || t.startsWith('...')); }"
      }
    },
    "abstract_lead": {
      "type": "boolean",
      "script": {
        "source": "String a = params._source.abstract; if (a == null) { emit(false); } else { a = a.trim(); emit(a.startsWith('…') || a.startsWith('...')); }"
      }
    }
  },
  "query": {
    "term": { "annotations": "domain-inclusion/hpv" }
  },
  "post_filter": {
    "bool": {
      "should": [
        { "term": { "title_lead": true } },
        { "term": { "abstract_lead": true } }
      ],
      "minimum_should_match": 1
    }
  }
}
```
### Part 1: Canonical reference selection
We want to create a preliminary list of canonical references for our truncated and abstract references, which will be confirmed by our HPV community folks. Selection criteria for canonicals for our truncated references are as follows.

* First, look for bibliographic duplicates within references to the HPV community (matches on title (fuzzy), authors (or near), publication year)
   * Select canonical references from references with the `domain-inclusion/hpv` annotation
   *  De-duplicating within the community also reduces duplicate reference rates
* Then, for remaining references, find duplicates that are in our broader reference cohort.

Create a list of references and proposed canonicals, and work with @acdonald to get them checked over by the HPV community folks for confirmation. 

[...]

PR #912, opened 2026-08-07
Record duplicate decision provenance and never replace a person's decision automatically

### LINKED ISSUE #582: Skip automatic deduplication if the most recent decision is manual
Per #569, we can now manually resolve duplicate decisions.

If the deduplication process is re-run, it's possible that it will come to a different conclusion than the manual deduplication.

We should consider skipping automatic deduplication if the active decision has manual provenance (we also then need to track provenance).

This is a low-priority issue - the only paths this can be hit are when doing `POST /references/duplicate-decisions/invoke/` or after #262 is fulfilled. Even then, it can only create `decoupled` decisions, not any decision with an immediate impact.

jaybea:
Closes #582.

Scope and Complete-when were edited today to protect `person` only, matching the Outcome they
had contradicted; see the comment on the issue if you read it before that. The backfill
therefore carries the protection alone, so a manual decision it fails to identify is
replaceable, which raises the stakes on the #752 population and #771's identifier capture.

### Beyond the issue

- Splits `register_duplicate_decision_for_reference` into `register_pending_import_decision` and
  `register_exact_duplicate_import_decision`. Behaviour preserving; the two callers always wrote
  different shapes through one parameterised method.
- SDK 0.14.2 adds both fields to `MakeDuplicateDecisionResult`, defaulted so a newer client still
  parses an older server.
- Server defaults on the new NOT NULL columns are permanent rather than dropped after the
  backfill: deploy and migrate are unordered jobs, so an old revision keeps inserting without
  these columns while the migration runs.

### Known limits

- `enhancement` is defined as a trigger but nothing writes it, since no path creates
  enhancement-triggered decisions today.
- The guard covers `map_duplicate_decision`, and the import registrars insert directly. Safe
  because those rows are either inactive or belong to a reference never stored, but nothing
  enforces the invariant that anything replacing an active decision goes through the mapper.

### Validation performed

- Ran the migration against a local database rolled back one revision. The backfill reclassified
  exactly 1,664 rows, matching a production audit of the manual population.
- Live run over two references, one `person` and one `unclassified`, both holding an active
  canonical decision. The proposal against the person decision landed inactive as `decoupled`
  with the original still active; against the unclassified one, the same path deactivated the
  old row and became active. One path, opposite outcomes.
- Removed the server defaults temporarily and confirmed an insert omitting the new columns
  fails, then restored them.
- Unit, integration and e2e suites green, pre-commit clean, docs build clean.

Three e2e deduplication tests set their prior decision through the manual API, so the guard now
intercepts ahead of the graph-shape branches they were written against. One of them asserted a
manual canonical being replaced by an automatic duplicate, which is the behaviour this change
removes; it is renamed and rewritten to assert the protection. The graph-shape branches keep
their unit coverage, where the active decision carries no authority.

### DIFF
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -180,6 +180,34 @@ class DuplicateDetermination(StrEnum):
         }
 
 
+class DuplicateDecisionAuthority(StrEnum):
+    """Who had authority for a duplicate decision."""
+
+    UNCLASSIFIED = auto()
+    """The authority of a historical decision has not been established."""
+    PERSON = auto()
+    """A person made the decision."""
+    SYSTEM = auto()
+    """The repository made the decision automatically."""
+
+
+class DuplicateDecisionTrigger(StrEnum):
+    """The event that caused a duplicate decision to be created."""
+
+    UNCLASSIFIED = auto()
+    """The trigger for a historical decision has not been established."""
+    MANUAL_API = auto()
+    """The manual duplicate-decision API created the decision."""
+    IMPORT = auto()
+    """Reference import processing created the decision."""
+    ENHANCEMENT = auto()
+    """Reserved. No path creates enhancement-triggered decisions yet."""
+    EXPLICIT_RERUN = auto()
+    """An explicit deduplication rerun caused the decision."""
+    MIGRATION = auto()
+    """A data migration classified or created the decision."""
+
+
 class Reference(
     DomainBaseModel,
     ProjectedBaseModel,  # References can self-project to the same structure
@@ -1142,6 +1170,14 @@ class ReferenceDuplicateDecision(DomainBaseModel, SQLAttributeMixin):
     """Model representing a decision on whether a reference is a duplicate."""
 
     reference_id: UUID = Field(description="The ID of the reference being evaluated.")
+    decision_authority: DuplicateDecisionAuthority = Field(
+        default=DuplicateDecisionAuthority.UNCLASSIFIED,
+        description="Whether a person or the system made the decision.",
+    )
+    decision_trigger: DuplicateDecisionTrigger = Field(
+        default=DuplicateDecisionTrigger.UNCLASSIFIED,
+        description="The event that caused the decision.",
+    )
     enhancement_id: UUID | None = Field(
         default=None,
         description=(
@@ -1168,6 +1204,20 @@ class ReferenceDuplicateDecision(DomainBaseModel, SQLAttributeMixin):
         default=None,
         description="Optional additional detail about the decision.",
     )
+
+    @model_validator(mode="after")
+    def check_provenance_is_classified_together(self) -> Self:
+        """Require authority and trigger to be classified together."""
+        authority_unclassified = (
+            self.decision_authority == DuplicateDecisionAuthority.UNCLASSIFIED
+        )
+        trigger_unclassified = (
+            self.decision_trigger == DuplicateDecisionTrigger.UNCLASSIFIED
+        )
+        if authority_unclassified != trigger_unclassified:
+            msg = "decision_authority and decision_trigger must be classified together"
+            raise ValueError(msg)
+        return self
 
     @model_validator(mode="after")
     def check_canonical_reference_id_populated_iff_duplicate(self) -> Self:
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -202,10 +202,8 @@ class DuplicateDecisionTrigger(StrEnum):
     """Reference import processing created the decision."""
     ENHANCEMENT = auto()
     """Reserved. No path creates enhancement-triggered decisions yet."""
-    EXPLICIT_RERUN = auto()
-    """An explicit deduplication rerun caused the decision."""
-    MIGRATION = auto()
-    """A data migration classified or created the decision."""
+    INVOKE_API = auto()
+    """The duplicate-decisions invoke API created the decision."""
 
 
 class Reference(
--- a/app/domain/references/service.py
+++ b/app/domain/references/service.py
@@ -27,6 +27,8 @@ from app.domain.references.models.models import (
     CandidateSelectionRequest,
     CandidateSelectionResult,
     CrossFacetCell,
+    DuplicateDecisionAuthority,
+    DuplicateDecisionTrigger,
     DuplicateDetermination,
     Enhancement,
     EnhancementRequest,
@@ -659,15 +661,14 @@ class ReferenceService(GenericService[ReferenceAntiCorruptionService]):
                 reference_id=str(reference.id),
                 canonical_reference_id=str(canonical_reference.id),
             )
-            await self._deduplication_service.register_duplicate_decision_for_reference(
+            await self._deduplication_service.register_exact_duplicate_import_decision(
                 reference_id=reference.id,
-                duplicate_determination=DuplicateDetermination.EXACT_DUPLICATE,
                 canonical_reference_id=canonical_reference.id,
             )
             return reference_create_result
 
         duplicate_decision = (
-            await self._deduplication_service.register_duplicate_decision_for_reference(
+            await self._deduplication_service.register_pending_import_decision(
                 reference_id=reference.id
             )
         )
@@ -1420,6 +1421,8 @@ class ReferenceService(GenericService[ReferenceAntiCorruptionService]):
             [
                 ReferenceDuplicateDecision(
                     reference_id=reference_id,
+                    decision_authority=DuplicateDecisionAuthority.SYSTEM,
+                    decision_trigger=DuplicateDecisionTrigger.EXPLICIT_RERUN,
                     duplicate_determination=DuplicateDetermination.PENDING,
                 )
                 for reference_id in reference_ids.reference_ids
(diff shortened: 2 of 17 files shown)


app/domain/references/models/models.py
```diff
@@ -180,6 +180,34 @@ def get_terminal_states(cls) -> set["DuplicateDetermination"]:
         }
 
 
+class DuplicateDecisionAuthority(StrEnum):
+    """Who had authority for a duplicate decision."""
+
+    UNCLASSIFIED = auto()
```
Adam-Hammo, line 186: Is there an advantage to `unclassified` over just a nullable column/field?

app/domain/references/services/anti_corruption_service.py
```diff
@@ -682,6 +684,8 @@ def duplicate_decision_from_sdk_make(
         try:
             reference_duplicate_decision = ReferenceDuplicateDecision(
                 reference_id=make_duplicate_decision.reference_id,
+                decision_authority=DuplicateDecisionAuthority.PERSON,
+                decision_trigger=DuplicateDecisionTrigger.MANUAL_API,
```
Adam-Hammo, line 688: This is one of those times where the anti-corruption layer really earns its keep!

app/domain/references/models/models.py
```diff
@@ -180,6 +180,34 @@ def get_terminal_states(cls) -> set["DuplicateDetermination"]:
...
+    """The trigger for a historical decision has not been established."""
+    MANUAL_API = auto()
+    """The manual duplicate-decision API created the decision."""
+    IMPORT = auto()
+    """Reference import processing created the decision."""
+    ENHANCEMENT = auto()
+    """Reserved. No path creates enhancement-triggered decisions yet."""
+    EXPLICIT_RERUN = auto()
+    """An explicit deduplication rerun caused the decision."""
```
Adam-Hammo: Is this a triggering from the `/references/duplicate-decisions/invoke/` endpoint? If so the doc could maybe be clearer, if not I think I'm misunderstanding

app/domain/references/service.py
```diff
@@ -1420,6 +1421,8 @@ async def add_pending_duplicate_decisions_for_reference_ids(
             [
                 ReferenceDuplicateDecision(
                     reference_id=reference_id,
+                    decision_authority=DuplicateDecisionAuthority.SYSTEM,
+                    decision_trigger=DuplicateDecisionTrigger.EXPLICIT_RERUN,
```
Adam-Hammo: Nitpick: these should maybe be passed in as function arguments. I could see a case where future code calls this function without realising these values are encoded within it, leading to messy provenance.

docs/procedures/deduplication.rst
```diff
@@ -202,39 +235,45 @@ The bold lines in the flowchart indicate what we expect to be nominal flow.
     flowchart LR
 
         N["New Decision (N)"]
+        C0{"N names a non-canonical reference?"}
```
Adam-Hammo, line 225: Just a sense-check, have you sighted the built docs? Sometimes the mermaid silently omits cell captions when they're too long

tests/e2e/imports/test_deduplication.py
Adam-Hammo, line 1: Does this file's diff represent a reduction in test surface?

I'd maybe suggest we don't need an e2e test for this guard - it doesn't test anything more than integration/unit testing would, while the old test did assert through to Elasticsearch that decisions took effect.

I see how it happened - the e2e tests use the API, which assume they're people, so can't be overwritten. Open to thoughts on how to structure it, and if the tests were doing much in the first place, otherwise I guess we need to do DB writes to set the test up as desired.

app/migrations/versions/6df1b2b092ed_add_duplicate_decision_provenance.py
```diff
@@ -0,0 +1,81 @@
...
+    expected = connection.execute(
+        sa.text(
+            "SELECT count(*) FROM reference_duplicate_decision WHERE "
+            + _MANUAL_EEF_DECISIONS
+        )
+    ).scalar_one()
+    result = connection.execute(
+        sa.text(
+            "UPDATE reference_duplicate_decision "
+            "SET decision_authority = 'person', decision_trigger = 'migration' "
```
Adam-Hammo: I'm torn as to whether `migration` is correct or whether we should represent the reality of `manual_api`

docs/procedures/deduplication.rst
```diff
@@ -187,6 +187,39 @@ The ``POST /references/duplicate-decisions/`` endpoint provides an interface to
 This may be useful if the automated process raises a reference for manual review, or if a user simply wants to manually deduplicate references.
 
 
+Decision authority and precedence
+---------------------------------
```
Adam-Hammo, line 191: Nitpick on these docs (this is very opinionated so feel free to push back / discuss!)

This feels like it will go over most folks' heads, and I'm not sure how much they need to know it anyway. I almost think your point under special cases (_"An automatic proposal would replace a person-made active decision. This applies even when the proposed result is identical."_) is all that's need to be said on user-facing documentation, maybe along with a sentence or two about triggers and authorities and their possible values (we could just drop the enum in?).

Anyway I could be wrong, I'm bad at writing docs, just my 2 cents

Adam-Hammo (commented):
Looking good, leaving a collection of questions and nitpicks

app/domain/references/models/models.py
jaybea: Well, it seemed like "We looked and could not establish it", was clearer than "no value here". And I think this means consumer of the SDK get a named value with documented meaning rather than a null they need to lookup or interpret.

app/domain/references/services/anti_corruption_service.py
jaybea: Sweet!

docs/procedures/deduplication.rst
jaybea: I'd looked at the diagrams. FYI, they are unreadable in dark mode for someone with my colour blindness, the lines just aren't there. Looks great in light mode!

I rechecked the title lengths and they seem to render fine.

app/domain/references/models/models.py
jaybea: Yes, it was meant to be the invoke api. A bare INVOKE seemed to generic. But I've changed it to INVOKE_API, to match MANUAL_API, that's better. Happy to change if you've got a better one!

app/domain/references/service.py
jaybea: Well, I've kept authority and trigger in rather than exposing them as arguments, since this operation has one origin, and renamed the method to register_pending_invoke_api_decisions so that origin is visible at every call site. If it gets more involved we could refactor.

tests/e2e/imports/test_deduplication.py
jaybea: You're right! The earlier rewrite reduced the transition and Elasticsearch coverage. CI tests failed and I just did a bulldoze on them.

I've restored the three transition cases. Setup goes through the manual API now so the normal SQL-to-Elasticsearch side effects run, then flips the decision's provenance to unclassified so automation is allowed to replace it, which is what the old tests were relying on.

The canonical-to-duplicate case asserts the reference is in Elasticsearch before the rerun and gone afterwards (that was kinda fun). And the duplicate-change case shows the duplicate-only annotation turning up on the canonical once it's attached.

The person guard is covered in the unit tests as you suggested.

app/migrations/versions/6df1b2b092ed_add_duplicate_decision_provenance.py
jaybea: Don't be torn, I'll just rename it to `manual_api`, that make more sense. The trigger name should be what caused the event, not the later mechanism that classified it. Thanks for the reality-check.

docs/procedures/deduplication.rst
jaybea: Backed it off, but did leave in awareness that other refs can attach as a duplicate to manually decided canonicals.

app/domain/references/service.py
Adam-Hammo: I more meant like

```python
    async def add_pending_duplicate_decisions_for_reference_ids(
        self,
        reference_ids: ReferenceIds,
        decision_authority: DuplicateDecisionAuthority = DuplicateDecisionAuthority.SYSTEM,
        decision_trigger: DuplicateDecisionTrigger = DuplicateDecisionTrigger.EXPLICIT_RERUN,
    ) -> list[ReferenceDuplicateDecision]:
```

Adam-Hammo: (I don't see the `register_pending_invoke_api_decisions` change I think?)

Adam-Hammo: Never mind, github did me dirty. Sorry!

tests/e2e/imports/test_deduplication.py
```diff
@@ -53,6 +55,38 @@
...
+        json=[payload],
+    )
+    assert response.status_code == 201
+    decision_id = UUID(response.json()[0]["id"])
+
+    pg_session.expire_all()
+    decision = await pg_session.get(SQLReferenceDuplicateDecision, decision_id)
+    assert decision
+    decision.decision_authority = DuplicateDecisionAuthority.UNCLASSIFIED
+    decision.decision_trigger = DuplicateDecisionTrigger.UNCLASSIFIED
```
Adam-Hammo, line 85: Nice workaround!

Adam-Hammo (approved):
Looks good to me!

PR #913, opened 2026-08-07
Container app job alerts

### LINKED ISSUE #866: Alert on container app job failures
Add alerting if a container app job fails, or reach their cutoff time (see also #862).

This should cover pending enhancement expiry, DB migration and ES migration. This should also ideally cover future jobs in the same container app environment.

Existing alerting configuration for other purposes exists in `alerts.tf` that may be fit for usage.

It's most important that this covers infrastructural issues (eg failed to start, crash loop), though ideal if it also covers unhealthy application exits. Failure code paths are already monitored in Honeycomb but two alerts is better than none!

Example incident of silent container app job failures: https://covidence.slack.com/archives/C0810ARPCRG/p1785207268941949

jack-o-the-green:
## Context

Closes #866 

Introduce alerting on container app job failures, one on execution failures and one on container app console logs. 

## What's Changed?
* Have added an alert for failing container app job executions that provides the job execution name in slack 
    * This replaced an earlier attempt to do an alert with a direct link to failing systems logs, which was too noisy (see [slack thread](https://covidence.slack.com/archives/C0998FP5743/p1786314471377389) if curious)
    * Alert example https://covidence.slack.com/archives/C0998FP5743/p1786317696596409
* Have added an alert that returns a container app console log failure
    * This includes a link to all log lines 10s before and 10s after the failure happens.
    * Alert example https://covidence.slack.com/archives/C0998FP5743/p1786072361930959
* The console alert will capture any new container app jobs manually, while the metric alert will catch any new jobs added using our scheduling terraform, but will need manual jobs added to it).
* We now have a separate slack flow for posting log search alerts to our alerting channels as well as a custom slack flow for posting container app job metrics alerts. 
    * This appropriately formats log search alerts and includes links to the relevant logs when they fire, as well as including execution name in the body of the metrics-based alert
    * Have accepted some duplication in terraform here to avoid long long strings of conditional formatting in the body of the slack message in terraform. It was getting out of hand and wholly unreadable. I also prioritized having the most useful alerts in slack possible instead of trying to limit the amount of terraform. 

I will do a follow up PR to rename the metrics alert resources to match the log metrics resources naming convention, e.g. `slack_alerts` -> `slack_metric_alerts`. I didn't want to muddy the diff and terraform plan/applies here. 

I will also revert the turning on of alerts in dev before merging.

## Some explicit decisions to improve diagnostics

* Log searchs alerts set the Resource they alert on to the _log analytics workspace_ the search is run against. 
   * This prevents us linking from slack alert -> affected job without manual brittle construction of resource urls that we shove into the alert body. I've elected to leave direct job resource links off the slack alerts in favour of logging links that actually work
   * Overriding the resource id breaks logging urls 
* Alert Events on the console logs alert will be high - it includes all of the related log lines, whereas with the system logs alert it should group slightly better. This is due to the fact that we are returning enough log lines from the alert query to immediately provide diagnostic information, otherwise if the alert is triggered off a "Traceback" match, you will get the single Traceback line with no additional info and this isn't particular actionable. 
    * There's potentially more conditional faff we could do in the slack alert body, but this was starting to feel like a rabbithole so I've left it alone.    
* We **do not** alert if a scheduled job fails to schedule in a way that does not trigger an execution. This felt like a huge rabbithole for minimal value, we are not broadly impacted if an expiry of pending enhancements does not run. 

## If you want to click something and/or test

Feel free to trigger failures in the development environment to test things out, for example you can run a start command with a non-existent image to get a console log alert. 

```
az containerapp job start \
  -g rg-destiny-repository-development \
  -n expire-pending-enhancements-deve \
  --container-name expire-pending-enhancements-deve \
  --image destinyevidenceregistry.azurecr.io/destiny-repository:91ebf3c \
  --command "/bin/echo" \
  --args "level=error manual alert test"
```

It creates an alert like here https://covidence.slack.com/archives/C0998FP5743/p1786072361930959

### DIFF
--- a/infra/app/alerts.tf
+++ b/infra/app/alerts.tf
@@ -13,6 +13,8 @@ resource "azurerm_logic_app_workflow" "slack_alerts" {
   tags = local.minimum_resource_tags
 }
 
+# --- Metrics Alerts ---
+
 resource "azurerm_logic_app_trigger_http_request" "slack_alerts" {
   count = local.env.alerts_enabled && var.alert_slack_webhook_url != "" ? 1 : 0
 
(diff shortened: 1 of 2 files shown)


infra/app/locals.tf
```diff
@@ -16,7 +16,13 @@ locals {
       db_ha_enabled   = false
       db_sku_name     = "GP_Standard_D2ds_v4"
 
-      alerts_enabled = false
+      alerts_enabled          = true
```
jack-o-the-green: I'll revert this before merging

infra/app/alerts.tf
```diff
@@ -381,3 +388,231 @@ resource "azurerm_monitor_metric_alert" "ui_restart_count" {
...
+  count = local.env.alerts_enabled ? 1 : 0
+
+  name                = "${local.name}-job-failure-events"
+  resource_group_name = azurerm_resource_group.this.name
+  location            = azurerm_resource_group.this.location
+  scopes              = [data.azurerm_log_analytics_workspace.container_apps.id]
+  description         = "A container app job in ${var.environment} has failed."
+  severity            = 1 # error
+
+  evaluation_frequency    = "PT1M"
```
jack-o-the-green, line 522: This is quite frequent, but the next tier up is 5min and that seems too long for catching container app job failures, especially as they will highlight failing database migrations

infra/app/alerts.tf
```diff
@@ -381,3 +388,402 @@ resource "azurerm_monitor_metric_alert" "ui_restart_count" {
...
+  # body: https://api.slack.com/block-kit
+  # expressions: https://learn.microsoft.com/en-us/azure/logic-apps/workflow-definition-language-functions-reference
+  body = <<-BODY
+    {
+      "blocks": [
+        {
+          "type": "header",
+          "text": {
+            "type": "plain_text",
+            "text": "@{if(equals(triggerBody()?['data']?['essentials']?['monitorCondition'], 'Resolved'), '✅ RESOLVED', '❌ FAILURE')} @{first(triggerBody()?['data']?['alertContext']?['condition']?['allOf'][0]?['dimensions'])?['value']}",
```
Adam-Hammo: Nit: we're very defensive, which is good, except on `allOf`

```suggestion
            "text": "@{if(equals(triggerBody()?['data']?['essentials']?['monitorCondition'], 'Resolved'), '✅ RESOLVED', '❌ FAILURE')} @{first(triggerBody()?['data']?['alertContext']?['condition']?['allOf']?[0]?['dimensions'])?['value']}",
```

infra/app/alerts.tf
```diff
@@ -381,3 +388,402 @@ resource "azurerm_monitor_metric_alert" "ui_restart_count" {
...
+            "emoji": true
+          }
+        },
+        {
+          "type": "section",
+          "fields": [
+            { "type": "mrkdwn", "text": "*Status:*\n@{triggerBody()?['data']?['essentials']?['monitorCondition']}" },
+            { "type": "mrkdwn", "text": "*Alert Events:*\n@{if(equals(triggerBody()?['data']?['essentials']?['monitorCondition'], 'Resolved'), '0', string(triggerBody()?['data']?['alertContext']?['condition']?['allOf'][0]?['metricValue']))}" }
```
Adam-Hammo:
```suggestion
            { "type": "mrkdwn", "text": "*Alert Events:*\n@{if(equals(triggerBody()?['data']?['essentials']?['monitorCondition'], 'Resolved'), '0', string(triggerBody()?['data']?['alertContext']?['condition']?['allOf']?[0]?['metricValue']))}" }
```

Adam-Hammo: etc

infra/app/alerts.tf
Adam-Hammo: We could also consider factoring out some of this shared parsing

```tf
locals {
  alert_condition = "triggerBody()?['data']?['alertContext']?['condition']?['allOf']?[0]"
}
```

infra/app/alerts.tf
```diff
@@ -532,12 +529,18 @@ resource "azurerm_monitor_scheduled_query_rules_alert_v2" "job_failure_events" {
   tags = local.minimum_resource_tags
 
   criteria {
+    # Returning the lines within 10s of a failure
```
Adam-Hammo, line 529: I think this might make the count in the alert just the total number of logs around the first and last failure irrespective of log level - probably inflating the count.
I don't know if that _matters_, I don't think the size of the number really impacts the urgency of the alert in any case

Adam-Hammo (commented):
Looking good, just defensive nits!

infra/app/alerts.tf
jack-o-the-green: Yeah it does do that, I felt it was worth it because otherwise we only link to the actual failing line which isn't as diagnostically useful out of context. I think I'll just remove it from the alert body, it's not useful as each alert represents a single container app job failure.

infra/app/alerts.tf
jack-o-the-green: You can't see if from this PR as it isn't a change but this condition parsing method isn't used for the other metrics alerts, so it feels inappropriate to me to pull out into a local generic. 

When I do the follow up PR to rename some existing resources, I'll have a look at options for reducing duplication if that works for you?

Adam-Hammo (approved):
Looks good to me!

Issue #915, opened 2026-08-08
Build the evaluation runner


jaybea:
Part of #791. Ready now: the runner can be built against a stand-in for the Deduper. Running it with the released Deduper is #887.

## Outcome

Assess a supplied set of records against the corpus and publish reviewable artifacts, without mutating repository-domain state. This issue is the tool; #887 is the run that uses it.

The runner is expected to write evaluation artifacts. It must not persist references, enhancements, import results or duplicate decisions; change Elasticsearch canonical state; or queue background work.

## Context

#874 builds `evaluate`, which finds candidates, fetches their full records and scores each pair. Nothing drives it over a supplied set of records, which #887 needs to publish a production-like evaluation.

An assessment is dated evidence against the corpus as it stood, and nothing promotes a stored one into a decision. The live path evaluates again under its own corpus and guards.

End-to-end coverage for the deduplication decision path stays with #874 / PR #916. This runner has focused tests for its artifacts, per-record failures and manifest. Non-mutation is not tested here. It follows from what the module can reach, which is blob storage and the assessor it is given, not references, Elasticsearch or background work. #887 is where a read-only proof becomes possible.

## Scope

- Read a supplied line-delimited blob, reusing import's parsing and conversion code, construct each incoming reference without persisting it, call `evaluate_supplied()` from #874 and write the assessment out.
  - Not through the import endpoint: it writes references, import results and a decision row per line, seeding the corpus with the records being evaluated.
  - Preserve records that are absent from the corpus and retrieval-negative cases. Build a frozen supplied reference and explicit candidate-selection input for each valid row, retaining the dataset `query_id` as its correlation key. A held-reference ID is optional and is used for self-exclusion; it is not a prerequisite for assessment.
- Treat the supplied dataset blob as immutable input. Write a new run-scoped bundle through `BlobRepository`; do not append results to, replace or annotate the input object. Give the run a UUIDv7 ID and write the four artifacts beneath `deduplication_evaluation/{run_id}/`. #915 produces the bundle using the configured blob backend; #887 publishes the selected production-like bundle to the shared `dedup-evals` container.
- Bypass the trusted-identifier shortcut, which otherwise returns before candidate selection and means no OpenAlex record reaches scoring. It stays configured and stays the only writer until the turn-on.
- Keep `soft_year_decay_year_optional_v1` available so yearless references continue through retrieval and scoring, which #877 needs to judge a year-optional policy on evidence.
[...]

Issue #875, opened 2026-07-30
Write duplicate decisions and record how each was reached


jaybea:
Part of #791. The writer, its guards and the provenance schema can be built now against the assessment contract from #874. Final mapping of Deduper fields is blocked on destiny-evidence/deduplication-toolkit#14 for the pair result and configuration interface.

## Outcome

`apply` is the only path that turns an assessment into a duplicate decision. When policy prevents a terminal decision, it retains the assessment for review with enough information for a person to understand why no decision was written. A terminal decision explains itself without re-running retrieval.

`ReferenceDuplicateDecision` remains where a terminal `CANONICAL`, `DUPLICATE` or `UNSEARCHABLE` outcome is recorded. #582 adds authority and trigger for every decision and delivers the guard protecting manual decisions; this issue calls it. #876 supplies the retrieval policy name and configured `k`; #874 supplies the candidate set and scores.

## Scope

Apply an assessment in one transaction: refuse automatic action unless the automatic-decision flag is on; re-check that any proposed canonical is still canonical; load the incoming reference's current persisted annotations; invoke the #582 authority guard; retain the assessment for review or write the permitted terminal decision. Start retryable Elasticsearch and background work only after a terminal decision commits; an assessment retained for review starts no decision side effects. The flag ships off; turning it on is where deep deduplication goes live, not here.

Apply the versioned acting policy from #877 to the assessment supplied by #874. The policy determines whether this writer records a terminal `CANONICAL`, `DUPLICATE` or `UNSEARCHABLE` decision or retains the assessment for review; do not reinterpret its threshold counts or evidence eligibility here. Neither retrieval route nor a Deduper probability is permission to act. Retain an assessment that #877 marks evidence-ineligible for review rather than treating it as an ordinary below-threshold rejection. Write `UNSEARCHABLE` only when #874 reports that the Elasticsearch route was unavailable and identifier lookup found no held candidate under #876.

The proposed canonical remains eligible when it is community-bound because its decision is not being changed.

Assessment happens before this transaction and may wait to be applied, so reload the incoming reference's annotations when applying the result. Base the decision on that state using the versioned annotation scheme/label list from the policy generation. The proposed canonical and checked annotation state must remain valid through commit; if either changes after its check, write no decision and retry. Check only the incoming reference for community membership, not every member of the candidate cluster. If no configured community annotation is present, treat the incoming OpenAlex reference as non-community. If its current annotations cannot be loaded, write no decision and retry or report the failure.

[...]

PR #916, opened 2026-08-10
Build the deduplication assessment path

### LINKED ISSUE #874: Build the deduplication assessment path

jaybea:
## Summary

Build the read-only deep-deduplication assessment path described in #874.

The new service:

- retrieves the complete candidate union from exact-identifier and Elasticsearch routes;
- hydrates stored inputs and candidates with identifiers and bibliographic enhancements only;
- scores every candidate through a `PairScorer` protocol;
- projects both sides of every pair into `DeduplicationPaper`, a repo-owned model built to the shape the deduplication toolkit scores, so no `Reference` crosses the scorer boundary;
- records scorer metadata, retrieval provenance, field-level evidence and threshold outcomes;
- supports supplied domain references for evaluation datasets without importing them first; and
- returns `PROPOSE_CANONICAL`, `PROPOSE_DUPLICATE` or `NO_PROPOSAL` without persisting a duplicate decision or starting background work.

## Assessment rules

- `PROPOSE_CANONICAL`: every pair was scored, none cleared the threshold, and the input was searchable.
- `PROPOSE_DUPLICATE`: every pair was scored and exactly one candidate cleared the threshold.
- `NO_PROPOSAL`: multiple candidates cleared, any pair was unscorable, or a no-match input was not searchable.

Threshold and scorer configuration are captured in the assessment so results remain explainable and reproducible.

## Boundaries

This PR does not decide whether a proposal may be applied. It has no decision-writer dependency and does not update PostgreSQL, Elasticsearch canonical state or background tasks.

The production Deduper adapter is intentionally deferred to #887. Tests and local verification use a `FakePairScorer` implementing the same protocol.

## Verification

- `uv run pytest`
  - 1,108 passed
  - 1 skipped
- `uv run pytest tests/e2e --log-cli-level info`
  - 29 passed
- All applicable pre-commit hooks passed, including Ruff, Ruff format, mypy and vulture.
- Local live verification passed against PostgreSQL, Elasticsearch and MinIO for:
  - a stored-reference duplicate proposal;
  - a self-exclusion/no-match canonical proposal;
  - multiple threshold-clearing candidates producing `NO_PROPOSAL`; and
  - a supplied reference built from the evaluation dataset and read from a MinIO blob.

Closes #874

### DIFF
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -1191,6 +1191,40 @@ class DeduplicationAssessment(BaseModel):
     unscorable_candidate_ids: list[UUID] = Field(default_factory=list)
 
 
+# The enhancement types the Deduper compares. Abstracts are excluded because its
+# scoring weights are configured without them, so sending one is wasted payload.
+SCORED_ENHANCEMENT_TYPES = (EnhancementType.BIBLIOGRAPHIC,)
+
+
+class DeduplicationPaper(ProjectedBaseModel):
+    """
+    The record shape the Deduper scores, built here rather than by the toolkit.
+
+    Mirrors the toolkit's ``Paper`` constructor so it can be handed over as
+    ``model_dump(exclude_none=True)``. Carries the fields the checked-in weights
+    score, plus identifiers and volume, which are unweighted today.
+    """
+
+    doi: destiny_sdk.identifiers.DOIIdentifier | None = None
+    openalex_id: destiny_sdk.identifiers.OpenAlexIdentifier | None = None
+    pubmed_id: destiny_sdk.identifiers.PubMedIdentifier | None = None
+    title: str | None = None
+    authors: list[destiny_sdk.enhancements.Authorship] | None = None
+    year: int | None = None
+    journal: str | None = None
+    pages: str | None = None
+    volume: str | None = None
+    issue: str | None = None
+
+    @model_validator(mode="after")
+    def validate_not_empty(self) -> Self:
+        """Reject an empty paper, which is the toolkit's own rejection too."""
+        if all(value is None for value in self.model_dump().values()):
+            msg = "A deduplication paper needs at least one populated field"
+            raise ValueError(msg)
+        return self
+
+
 class LinkedDataProjection(ProjectedBaseModel):
     """Result of projecting a LinkedDataEnhancement into flat searchable fields."""
 
--- a/app/domain/references/models/models.py
+++ b/app/domain/references/models/models.py
@@ -18,6 +18,7 @@ from pydantic import (
     ConfigDict,
     Field,
     HttpUrl,
+    JsonValue,
     PositiveInt,
     StringConstraints,
     TypeAdapter,
@@ -952,10 +953,20 @@ class CandidateSelectionInput(BaseModel):
         default_factory=list,
         description="Identifiers to union with the Elasticsearch candidates.",
     )
+    excluded_reference_id: UUID | None = Field(
+        default=None,
+        description="Reference ID to exclude from inline candidate selection.",
+    )
 
     @model_validator(mode="after")
     def validate_exactly_one_input_mode(self) -> Self:
         """Require exactly one of reference_id or inline candidate-search fields."""
+        if self.reference_id is not None and self.excluded_reference_id is not None:
+            msg = (
+                "excluded_reference_id applies to inline input; the reference_id "
+                "mode already excludes its own reference."
+            )
+            raise ValueError(msg)
         has_reference_id = self.reference_id is not None
         has_inline = bool(
             self.title or self.authors or self.publication_year or self.identifiers
@@ -1092,6 +1103,92 @@ class CandidateSelectionResult(BaseModel):
     input_searchability: InputSearchability
     diagnostics: CandidateSelectionDiagnostics
     candidates: list[Candidate] = Field(default_factory=list)
+
+
+class DeduplicationFieldStatus(StrEnum):
+    """How the values for one scored field relate across a reference pair."""
+
+    MATCH = auto()
+    """Both values are present and match."""
+    MISMATCH = auto()
+    """Both values are present and differ."""
+    MISSING_INCOMING = auto()
+    """Only the incoming value is missing."""
+    MISSING_CANDIDATE = auto()
+    """Only the candidate value is missing."""
+    MISSING_BOTH = auto()
+    """Both values are missing."""
+
+
+class DeduplicationFieldComparison(BaseModel):
+    """Compared values, availability and score for one Deduper field."""
+
+    incoming_value: JsonValue | None = None
+    candidate_value: JsonValue | None = None
+    status: DeduplicationFieldStatus
+    score: float | None = None
+
+
+class DeduplicationPairResult(BaseModel):
+    """Structured evidence returned by a Deduper adapter for one pair."""
+
+    probability: float | None = Field(default=None, ge=0, le=1)
+    field_comparisons: dict[str, DeduplicationFieldComparison] = Field(
+        default_factory=dict
+    )
+    early_stop_reason: str | None = None
+    doi_mismatch_adjusted: bool = False
+    suggested_label: str | None = None
+    unscorable_reason: str | None = None
+
+    @model_validator(mode="after")
+    def validate_scoring_state(self) -> Self:
+        """Require either a probability or an unscorable reason."""
+        if (self.probability is None) == (self.unscorable_reason is None):
+            msg = "Exactly one of probability or unscorable_reason is required"
+            raise ValueError(msg)
+        return self
+
+
+class DeduperMetadata(BaseModel):
+    """Identity and effective configuration of one loaded Deduper."""
+
+    package_version: str
+    configuration_hash: str
+    threshold: float = Field(ge=0, le=1)
+    effective_configuration: dict[str, JsonValue] = Field(default_factory=dict)
+
+
+class ScoredDeduplicationCandidate(BaseModel):
+    """One retrieved candidate and the Deduper evidence for its pair."""
+
+    candidate: Candidate
+    pair_result: DeduplicationPairResult
+    clears_threshold: bool | None
+
+
+class DeduplicationAssessmentOutcome(StrEnum):
+    """Reference-level proposal produced from all threshold-clearing pairs."""
+
+    PROPOSE_CANONICAL = auto()
+    """Every pair scored below threshold and the input was searchable."""
+    PROPOSE_DUPLICATE = auto()
+    """Every pair was scored and exactly one candidate cleared the threshold."""
+    NO_PROPOSAL = auto()
+    """The input was unsearchable, scoring was incomplete, or matches conflicted."""
+
+
+class DeduplicationAssessment(BaseModel):
+    """Read-only candidate retrieval and pair-scoring result for one reference."""
+
+    incoming_reference_id: UUID
+    candidate_selection: CandidateSelectionResult
+    deduper: DeduperMetadata
+    scored_candidates: list[ScoredDeduplicationCandidate] = Field(default_factory=list)
+    outcome: DeduplicationAssessmentOutcome
+    proposed_duplicate_of_id: UUID | None = None
+    threshold_clearing_candidate_ids: list[UUID] = Field(default_factory=list)
+    unscorable_candidate_ids: list[UUID] = Field(default_factory=list)
 
 
 class LinkedDataProjection(ProjectedBaseModel):
--- a/app/domain/references/services/anti_corruption_service.py
+++ b/app/domain/references/services/anti_corruption_service.py
@@ -95,17 +95,6 @@ class ReferenceAntiCorruptionService(GenericAntiCorruptionService):
         self, reference: RedactedReference
     ) -> destiny_sdk.references.Reference:
         """Convert the reference to a Reference SDK model."""
-        return await self._reference_to_sdk(reference)
-
-    async def internal_reference_to_sdk(
-        self, reference: Reference
-    ) -> destiny_sdk.references.Reference:
-        """Convert an unredacted reference for internal use, never an API response."""
-        return await self._reference_to_sdk(reference)
-
-    async def _reference_to_sdk(
-        self, reference: Reference
-    ) -> destiny_sdk.references.Reference:
         try:
             enhancements = [
                 await self.enhancement_to_sdk(enhancement)
--- a/app/domain/references/services/anti_corruption_service.py
+++ b/app/domain/references/services/anti_corruption_service.py
@@ -95,6 +95,17 @@ class ReferenceAntiCorruptionService(GenericAntiCorruptionService):
         self, reference: RedactedReference
     ) -> destiny_sdk.references.Reference:
         """Convert the reference to a Reference SDK model."""
+        return await self._reference_to_sdk(reference)
+
+    async def internal_reference_to_sdk(
+        self, reference: Reference
+    ) -> destiny_sdk.references.Reference:
+        """Convert an unredacted reference for internal use, never an API response."""
+        return await self._reference_to_sdk(reference)
+
+    async def _reference_to_sdk(
+        self, reference: Reference
+    ) -> destiny_sdk.references.Reference:
         try:
             enhancements = [
                 await self.enhancement_to_sdk(enhancement)
--- a/app/domain/references/services/deduplication_assessment_service.py
+++ b/app/domain/references/services/deduplication_assessment_service.py
@@ -4,7 +4,6 @@ from collections.abc import Awaitable, Callable
 from typing import Protocol
 from uuid import UUID
 
-import destiny_sdk
 from pydantic import ValidationError
 from sqlalchemy.exc import InterfaceError, OperationalError
 
@@ -29,10 +28,8 @@ from app.domain.references.models.models import (
 )
 from app.domain.references.models.projections import CandidateReferenceProjection
 from app.domain.references.services.access_control_service import (
+    RedactedReference,
     ReferenceAccessControlService,
-)
-from app.domain.references.services.anti_corruption_service import (
-    ReferenceAntiCorruptionService,
 )
 
 CandidateSelector = Callable[
@@ -71,8 +68,8 @@ class PairScorer(Protocol):
 
     async def score_pair(
         self,
-        incoming: destiny_sdk.references.Reference,
-        candidate: destiny_sdk.references.Reference,
+        incoming: RedactedReference,
+        candidate: RedactedReference,
     ) -> DeduplicationPairResult:
         """Score one incoming-reference and candidate pair."""
         ...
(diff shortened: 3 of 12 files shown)


jaybea:
Sorry, for the late commit/update, but while working on the evaluation runner #915, I realised the assessment code needed more nuanced handling of ES and DB failures: a transient blip aborted the whole caller, even though evaluate_supplied documents that it raises DeduplicationError per record. Trapped and remapped ESError and connectivity DBAPI errors to DeduplicationError, but took care to let malformed queries or runtime errors propogate.

app/domain/references/models/models.py
```diff
@@ -1094,6 +1105,92 @@ class CandidateSelectionResult(BaseModel):
     candidates: list[Candidate] = Field(default_factory=list)
 
 
+class DeduplicationFieldStatus(StrEnum):
```
jack-o-the-green, line 1108: Looking at the increasing size of this file, I feel like during this work would be a good time to bite the bullet and extract the deduplication models into a `domain/references/models/deduplication.py`

Happy if the change is made separate to this PR, or at another time if we want to move all deduplication models in one swoop.

app/domain/references/models/models.py
```diff
@@ -1094,6 +1105,92 @@ class CandidateSelectionResult(BaseModel):
...
+    candidate_value: JsonValue | None = None
+    status: DeduplicationFieldStatus
+    score: float | None = None
+
+
+class DeduplicationPairResult(BaseModel):
+    """Structured evidence returned by a Deduper adapter for one pair."""
+
+    probability: float | None = Field(default=None, ge=0, le=1)
+    field_comparisons: dict[str, DeduplicationFieldComparison] = Field(
```
jack-o-the-green, line 1136: `nitpick` Would it be beneficial to restrict the keys of the dictionary to deduper fields? If you wanted to get fancy you could generate a field enum from the relevant pydantic model fields, or just add an after validator.

app/domain/references/services/anti_corruption_service.py
```diff
@@ -95,6 +95,17 @@ async def reference_to_sdk(
         self, reference: RedactedReference
     ) -> destiny_sdk.references.Reference:
         """Convert the reference to a Reference SDK model."""
+        return await self._reference_to_sdk(reference)
+
+    async def internal_reference_to_sdk(
```
jack-o-the-green: There's something sitting wrong in my gut about adding an internal-only method to the anti-corruption service. I see the anti-corruption service as the translation layer for the API, so we know we're sending safe stuff at the boundary, whereas this should never be used at that boundary (but we can't really enforce that except through documentation)

Would it make sense to just use redacted sdk references for deduplication? It doesn't rely on full text enhancements anyway.

app/domain/references/services/deduplication_assessment_service.py
```diff
@@ -0,0 +1,262 @@
...
+
+    async def evaluate(
+        self,
+        reference_id: UUID,
+        *,
+        retrieval_policy: RetrievalPolicyName | None = None,
+        k: int | None = None,
+    ) -> DeduplicationAssessment:
+        """Assess a stored reference from one hydrated input snapshot."""
+        hydrated = await self._reference_reader.get_hydrated(
```
jack-o-the-green: `extreme nitpick` worth adding a hydration method to the protocol that works with a single reference id to avoid the list indexing in this function?

app/domain/references/services/deduplication_assessment_service.py
```diff
@@ -0,0 +1,262 @@
...
+    """Pair-scoring contract implemented by a stand-in or toolkit adapter."""
+
+    @property
+    def metadata(self) -> DeduperMetadata:
+        """Return the scorer identity and fixed configuration."""
+        ...
+
+    async def score_pair(
+        self,
+        incoming: destiny_sdk.references.Reference,
```
jack-o-the-green: I think I'm missing why we need to use sdk references here at all, linked to my comment about the anti-corruptions service earlier. 

Would it not be more appropriate to convert an incoming evaluation dataset of un-imported references to reference domain models using the anti-corruption service, then have the deduplication_assessment_service work entirely with domain objects? Converting to a domain object doesn't mean it will be imported into the database.

jack-o-the-green (commented):
This is looking really good, the deduplication assessment service is super clean and readable. A few comments around sdk model usage/sdk boundaries from me.

app/domain/references/services/deduplication_assessment_service.py
```diff
@@ -0,0 +1,262 @@
...
+            msg = f"Could not hydrate incoming deduplication reference: {reference_id}"
+            raise DeduplicationNotFoundError(msg)
+        reference = hydrated[0]
+        incoming = await self._anti_corruption_service.internal_reference_to_sdk(
+            reference
+        )
+        candidate_reference = CandidateReferenceProjection.get_from_reference(reference)
+        return await self._assess(
+            incoming,
+            CandidateSelectionInput(
```
jack-o-the-green: Technically (but not likely) I think title, authors, publication_year, and identifiers could all be empty, meaning we'd throw a validation error here when constructing, add a guard?

app/domain/references/services/deduplication_assessment_service.py
```diff
@@ -0,0 +1,262 @@
...
+    ) -> DeduplicationAssessment:
+        """
+        Assess a frozen SDK reference against a query the runner built.
+
+        The caller owns the query payload because it decides which fields and
+        identifiers the record presents, and sets ``excluded_reference_id`` when the
+        record is already held so it cannot be retrieved as its own candidate.
+        """
+        return await self._assess(
+            incoming,
```
jack-o-the-green, line 161: Is there value in filtering the incoming reference to 'Bibliographic' enhancements here also (to match the hydrated reference from above)? Thinking that a reference here could be quite a bit larger than one we hydrate from an id, and this results in different `incoming` coming in through this way to `_assess` for equivalent references between the two evaluation functions.

app/domain/references/models/models.py
jaybea: That sounds good, this PR is big enough already. I'll take a look and see if I can do it in #875.

app/domain/references/models/models.py
jaybea: I'd like to leave this open for now. Deduper's field set is configuration rhather than a fixed list, and it is not really stable yet, for instance abstract is now gone (thankfully!).

We have strong provenance for the version. There will be a Deduper adapter when the real thing lands, I think that would be a good place for the final design for this.

app/domain/references/services/anti_corruption_service.py
jaybea: You are right. I've removed it. Our path now redacts with an unentitled `ReferenceAccessControlService` and calls the existing `reference_to_sdk`.

app/domain/references/services/deduplication_assessment_service.py
jaybea: I'd like to leave this as-is rather than add a new method in the repository. This is the only array wrapped single element call I could find, if we get more, we should!

app/domain/references/services/deduplication_assessment_service.py
jaybea: I think I fixed the anti-corruption issue above, thanks for raising that one.

 For `evaluate_supplied()`, the Deduper expects SDK `Reference`s, so I’d prefer to pass the same shape the scorer is going to receive. At this point, I think converting an unimported evaluation row into a domain `Reference` would still end with converting it back to SDK for the Deduper.

app/domain/references/services/deduplication_assessment_service.py
jaybea: For OpenAlex ingestion, by definition there's always a W-ID identifier, BUT, this was a very nice catch, worse than a validation error. Nothing caught it, so a bare pydantic `ValidationError` escaped past the `DeduplicationError`. Grrr!!

I think I got it now, the input construction is wrapped and raises `DeduplicationValueError`.

app/domain/references/services/deduplication_assessment_service.py
jaybea: You got it! And using a single `SCORED_ENHANCEMENT_TYPES` constant (no drift!) and apply it inside `_assess()` now, so both `evaluate()` and `evaluate_supplied()` agree.

Fortunately, Deduper doesn't score abstracts so we're not sending those.

app/domain/references/models/models.py
jack-o-the-green: Makes sense to me, thanks for the additional context!

app/domain/references/services/deduplication_assessment_service.py
jack-o-the-green: Just confirming what we mean by "repository" in this case, do we mean destiny-repository or do we mean the [reference SQL repo here](https://github.com/destiny-evidence/destiny-repository/blob/14ca34db89520050ac6dad510189effec4fedc2e/app/domain/references/repository.py#L149). 

If the latter, I was thinking that you could just add a default method implementation on the Protocol itself which wraps the existing `get_hydrated` to return a single reference, so the change would be this file only. 

However, still a nitpick, up to you!

app/domain/references/services/deduplication_assessment_service.py
jack-o-the-green: This raises a broader concern from me about having the deduper accept those SDK references as inputs. Given it's an external service/toolkit for us, I think that the repository should own the conversion from our format to their format before sending requests to them - so we handle conversion of either domain references or sdk references into the format the Deduper defines. 

In this case that would mean we adopt this conversion into the repository https://github.com/destiny-evidence/deduplication-toolkit/blob/main/destiny_dedupe/data_models.py#L108-L193 and then the deduplication toolkit would accept `Paper` as it's input https://github.com/destiny-evidence/deduplication-toolkit/blob/main/destiny_dedupe/data_models.py#L27-L43. That gives the toolkit ownership of it's input format, which feels like where that ownership belongs and reduces coupling. 

I know I wasn't at the walkthrough where it was decided that the deduper would own the conversion (https://github.com/destiny-evidence/deduplication-toolkit/issues/17), am sorry I am raising concerns so late!

app/domain/references/services/deduplication_assessment_service.py
jaybea: Too many repositories! I should have said `ReferenceSQLRepository`.

I tried the Protocol-default version, but `ReferenceSQLRepository` and the test fakes satisfy `ReferenceReader` by duck typing rather than inheritance. So the default body wouldn't exist on the runtime object.

And a defaulted member still counts as a protocol member, so `ReferenceSQLRepository` would have to provide it too, which is back at the repository change.

The unit tests would hide that, because their readers are mocks: the spec auto-creates the method as a mock rather than running the Protocol's default.

I've added a private `_hydrate_one` on the assessment service instead. If this solution still bugs you, happy to give it another go.

app/domain/references/services/deduplication_assessment_service.py
jaybea: This was a great callout, the code is much improved as a result!

I took it on and changed `PairScorer` to take our redacted domain `Reference`, so the assessment service no longer deals in SDK shapes at all. Anti-corruption dropped out entirely and the PR came down by almost 70 lines.

On moving the `Paper` mapping here, I'd rather not yet. You're right that it would eventually get the toolkit off our SDK, but it wouldn't today: `Paper`'s fields and `normalise_authors` are destiny-sdk types, so they'd still import the SDK and we'd be building SDK values to populate `Paper`. The coupling just changes location rather than is removed.

So the boundary I'd keep for now is the one `openalex-incremental-updater` uses with us: external code speaks SDK. Our own concrete `PairScorer` does the domain-to-SDK step, and the toolkit keeps the SDK-to-`Paper` step it already has.

Getting your version to pay off means de-SDK-ing `Paper` and the normalisers first, and I'd rather not put that in front of the toolkit team right now.

app/domain/references/services/deduplication_assessment_service.py
jack-o-the-green: That's great, thanks for indulging my nitpick so much!

jaybea: Happy to!

app/domain/references/services/deduplication_assessment_service.py
jack-o-the-green: Those changes you made are great, nice to see the internal service not needing to worry about SDK models. 

> Getting your version to pay off means de-SDK-ing Paper and the normalisers first, and I'd rather not put that in front of the toolkit team right now.

I don't agree with this one - I don't think removing the SDK from the deduplication toolkit is the major benefit here (though still something that I'd consider ideal long term). For me it's about the ownership of input formats and where we're drawing responsibility boundaries. Moving the mapping to Paper into destiny-repository establishes the first step of "the deduplication toolkit owns it's own domain models, _we_ are responsible for giving it something it can use" 

I guess I'm arguing that I do not think we should do this ticket here https://github.com/destiny-evidence/deduplication-toolkit/issues/17 and that we should take the first step down the path to removing destiny sdk from the deduplication toolkit because confirming that ownership boundary is valuable in of itself. 

Happy to chat about it in tech meet if going back and forward on the thread isn't helpful!

jaybea:
@jack-o-the-green

This answers your ownership-boundary thread directly. Two commits:

**`Project deduplication scorer inputs to Paper`** adds `DeduplicationPaper`, a repo-owned model, and a projection that builds it from a domain `Reference`. `PairScorer.score_pair()` now takes two of those. The scorer receives no `Reference` of any kind.

**`Score deduplication pairs without a redacted reference view`** drops the `RedactedReference` step from this path. It only ever removed full-text enhancements, a denylist that the paper's fixed field list already dominates, so it was doing no work here.

**The type story, because it looks cool in a table:**

| layer | type | who translates |
|---|---|---|
| wire and dataset input | SDK models | the anti-corruption layer |
| domain services | domain `Reference` | nothing, this is the internal model |
| scorer boundary | `DeduplicationPaper` | this projection |

`DeduplicationAssessmentService` still takes a domain `Reference` on both entrypoints, unchanged.

As discussed, `DeduplicationPaper` still uses SDK identifier and authorship types for its fields, because their `Paper` does and the handover is `model_dump(exclude_none=True)`. We'll work on fixing that. What has gone is the SDK `Reference`, not every SDK type. Owning plain types with SDK-shaped construction pushed into the adapter is cleaner layering and a bigger change than this one.

**Toolkit issue 17 should be closed or re-scoped.** It was written expecting the toolkit to do the conversion. Projecting here answers it, so I'll close it.

**I've updated the description**, three changes:

- the supplied bullet said SDK references, now domain references;
- a new bullet for the Paper projection; and
- current test counts.

Tests: 1,108 unit, 29 e2e, pre-commit including mypy and vulture. Plus a 6,000-reference check against the local corpus confirming the redaction removal changes no projected output.

app/domain/references/services/deduplication_assessment_service.py
```diff
@@ -93,15 +90,14 @@ def __init__(
...
-        # fields, so nothing here may reach full text.
-        self._access_control_service = ReferenceAccessControlService()
 
-    def _to_scorer_paper(self, reference: Reference) -> DeduplicationPaper:
-        """Reduce a reference to the least-privileged view the scorer sees."""
-        redacted = self._access_control_service.redact_reference(reference)
+    @staticmethod
+    def _to_scorer_paper(reference: Reference) -> DeduplicationPaper:
+        """Project a reference into the record the scorer compares."""
+        # No access-control redaction: a paper's fields are an allowlist, so it
```
jack-o-the-green: `nitpick` is this comment about removed redaction valuable to keep?

jack-o-the-green (approved):
LGTM - thank you for your work on this!

app/domain/references/services/deduplication_assessment_service.py
jaybea: Oops. Slipped through, thanks.

PR #919, opened 2026-08-11
Build the deduplication evaluation runner

### LINKED ISSUE #915: Build the evaluation runner

jaybea:
Closes #915. Assess a supplied JSONL set of records against the corpus and write a reviewable artifact bundle, plus a driver to run one.

Base `main`. Seven source files, five test files.

Four implementation notes:

- A record that contradicts the run's configuration aborts with `EvaluationConfigurationMismatchError`, so a manifest never misdescribes what ran. The completed records are still written, and the missing manifest marks the bundle incomplete. That error deliberately does not subclass `DeduplicationError`, which the per-record handler catches and would turn into an envelope.
- A record declaring a different `dataset_version` than the configuration aborts the same way, but is checked before the assessment rather than after it. The input can contradict the manifest on its face, so there is no reason to spend retrieval on it first.
- The manifest records a SHA-256 of the input, and it has to hash the exact bytes held on the blob. The line reader hands back decoded strings, so it cannot produce that hash, and the input is streamed a second time as raw bytes. That pass is `BlobRepository.digest()`, which shares `copy()`'s chunk handling, and it carries the input size cap so an oversized input aborts before any record is assessed.
- The driver holds a read-only session per record rather than one for the run. That costs connection churn and buys interruptibility: no transaction spans records, so a stopped run still uploads what it completed. Postgres refuses a write from those transactions; on Elasticsearch the guarantee is only that the assembled services take read paths.

## Validation performed

- Ran the full `retrieval-query-set/v1` known-match set against local Postgres, Elasticsearch and MinIO: about 6,000 searchable records, about 41,000 scored pairs, complete bundle in around 90 seconds, no failed records.
- Ran the same set a second time through the released deduper wheel via a local adapter, exercising the real scorer path end to end. Both runs retrieved identically, which is what should happen when only the scorer changes.
- Verified bundles by reading them back out of MinIO: input SHA-256 against the stored bytes, one envelope per non-blank line in order, manifest counts and artifact checksums, pair rows against the scored candidates.
- The set carries a record with a raw U+2028. The runner read 6,042 lines where `splitlines()` on the same bytes gives 6,043, so line numbering held.
- Declared `reference_v9` against a live `reference_v1` and got the abort, the completed envelopes and no manifest.
- Declared a record from another dataset at line 3 of four and got the abort, the two completed envelopes and no manifest. The assessor was called exactly twice, so the guard ran before the offending record reached it.
- Yearless rows returned 0 candidates under `candidate_selection_v1`, which requires publication year, so the searchability gate rejects them before any query is issued. The same rows returned 7 to 10 each under `soft_year_decay_year_optional_v1`.
- Checked non-mutation by counting: nine tables, the latest decision timestamp and the Elasticsearch doc count unchanged either side of a run.
- Suite 1176 passed / 1 skipped, `tests/e2e` 29 passed, mypy and pre-commit clean.

Not covered: the deduper run above used a local wheel and a throwaway adapter, so wiring the released deduper into the repository is still #887.

### DIFF
--- a/app/core/exceptions.py
+++ b/app/core/exceptions.py
@@ -666,6 +666,10 @@ class DeduplicationValueError(DeduplicationError, ValueError):
         super().__init__(detail)
 
 
+class EvaluationConfigurationMismatchError(DestinyRepositoryError):
+    """An evaluation run contradicted the configuration its manifest publishes."""
+
+
 class ParseError(DestinyRepositoryError):
     """An exception for when we fail to parse some input."""
 
--- a/app/domain/references/services/deduplication_assessment_service.py
+++ b/app/domain/references/services/deduplication_assessment_service.py
@@ -49,12 +49,14 @@ UNPROJECTABLE_CANDIDATE_REASON = "Candidate could not be projected for scoring."
 class ReferenceReader(Protocol):
     """Reference reads required by assessment orchestration."""
 
-    async def get_hydrated(
+    # Awaitable rather than ``async def``: the latter demands a coroutine, which the
+    # traced repository methods do not declare, and this only ever awaits the result.
+    def get_hydrated(
         self,
         reference_ids: list[UUID],
         enhancement_types: list[str] | None = None,
         external_identifier_types: list[str] | None = None,
-    ) -> list[Reference]:
+    ) -> Awaitable[list[Reference]]:
         """Read full references with optional relationship filters."""
         ...
 
--- a/app/domain/references/services/deduplication_assessment_service.py
+++ b/app/domain/references/services/deduplication_assessment_service.py
@@ -49,8 +49,7 @@ UNPROJECTABLE_CANDIDATE_REASON = "Candidate could not be projected for scoring."
 class ReferenceReader(Protocol):
     """Reference reads required by assessment orchestration."""
 
-    # Awaitable rather than ``async def``: the latter demands a coroutine, which the
-    # traced repository methods do not declare, and this only ever awaits the result.
+    # Awaitable, not ``async def``: traced repository methods don't declare one.
     def get_hydrated(
         self,
         reference_ids: list[UUID],
@@ -111,8 +110,6 @@ class DeduplicationAssessmentService:
         try:
             candidate_paper = self._to_scorer_paper(candidate)
         except DeduplicationValueError:
-            # One unprojectable candidate is not a failed assessment; it is a
-            # candidate the Deduper was never given a chance to score.
             return DeduplicationPairResult(
                 unscorable_reason=UNPROJECTABLE_CANDIDATE_REASON
             )
@@ -121,8 +118,7 @@ class DeduplicationAssessmentService:
                 incoming=incoming, candidate=candidate_paper
             )
         except Exception as exc:
-            # Deliberately broad, unlike the narrow retrieval and hydration
-            # handlers: a third-party scorer's defect is one record's, not the run's.
+            # Broad by design: a scorer's defect fails one record, not the run.
             msg = f"Candidate scoring failed ({type(exc).__name__}): {exc}"
             raise DeduplicationError(msg) from exc
 
--- a/app/domain/references/services/deduplication_assessment_service.py
+++ b/app/domain/references/services/deduplication_assessment_service.py
@@ -104,6 +104,28 @@ class DeduplicationAssessmentService:
             msg = f"Cannot build a deduplication paper for {reference.id}"
             raise DeduplicationValueError(msg) from exc
 
+    async def _score_candidate(
+        self, incoming: DeduplicationPaper, candidate: Reference
+    ) -> DeduplicationPairResult:
+        """Score one pair, or say why the Deduper could not be asked."""
+        try:
+            candidate_paper = self._to_scorer_paper(candidate)
+        except DeduplicationValueError:
+            # One unprojectable candidate is not a failed assessment; it is a
+            # candidate the Deduper was never given a chance to score.
+            return DeduplicationPairResult(
+                unscorable_reason=UNPROJECTABLE_CANDIDATE_REASON
+            )
+        try:
+            return await self._pair_scorer.score_pair(
+                incoming=incoming, candidate=candidate_paper
+            )
+        except Exception as exc:
+            # Deliberately broad, unlike the narrow retrieval and hydration
+            # handlers: a third-party scorer's defect is one record's, not the run's.
+            msg = f"Candidate scoring failed ({type(exc).__name__}): {exc}"
+            raise DeduplicationError(msg) from exc
+
     async def _hydrate_one(self, reference_id: UUID) -> Reference:
         """Read one stored reference, treating an empty result as not-found."""
         hydrated = await self._reference_reader.get_hydrated(
@@ -229,20 +251,9 @@ class DeduplicationAssessmentService:
         scorer_metadata = self._pair_scorer.metadata.model_copy(deep=True)
         threshold = scorer_metadata.threshold
         for candidate in candidate_selection.candidates:
-            try:
-                candidate_paper = self._to_scorer_paper(
-                    candidates_by_id[candidate.reference_id]
-                )
-            except DeduplicationValueError:
-                # One unprojectable candidate is not a failed assessment; it is a
-                # candidate the Deduper was never given a chance to score.
-                pair_result = DeduplicationPairResult(
-                    unscorable_reason=UNPROJECTABLE_CANDIDATE_REASON
-                )
-            else:
-                pair_result = await self._pair_scorer.score_pair(
-                    incoming=scored_incoming, candidate=candidate_paper
-                )
+            pair_result = await self._score_candidate(
+                scored_incoming, candidates_by_id[candidate.reference_id]
+            )
             clears_threshold = (
                 pair_result.probability >= threshold
                 if pair_result.probability is not None
--- a/app/domain/references/services/deduplication_evaluation_runner.py
+++ b/app/domain/references/services/deduplication_evaluation_runner.py
@@ -4,7 +4,7 @@ import json
 from collections.abc import AsyncIterable, AsyncIterator, Sequence
 from enum import StrEnum, auto
 from json import JSONDecodeError
-from typing import Literal, Protocol
+from typing import Literal, Protocol, cast
 from uuid import UUID, uuid7
 
 import destiny_sdk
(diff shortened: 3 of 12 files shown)


app/domain/references/services/deduplication_assessment_service.py
```diff
@@ -49,12 +49,14 @@
 class ReferenceReader(Protocol):
     """Reference reads required by assessment orchestration."""
 
-    async def get_hydrated(
+    # Awaitable rather than ``async def``: the latter demands a coroutine, which the
+    # traced repository methods do not declare, and this only ever awaits the result.
```
Adam-Hammo:
```suggestion
    # Awaitable rather than ``async def``; not strictly a coroutine
```

app/domain/references/services/deduplication_assessment_service.py
```diff
@@ -102,6 +104,28 @@ def _to_scorer_paper(reference: Reference) -> DeduplicationPaper:
...
 
+    async def _score_candidate(
+        self, incoming: DeduplicationPaper, candidate: Reference
+    ) -> DeduplicationPairResult:
+        """Score one pair, or say why the Deduper could not be asked."""
+        try:
+            candidate_paper = self._to_scorer_paper(candidate)
+        except DeduplicationValueError:
+            # One unprojectable candidate is not a failed assessment; it is a
+            # candidate the Deduper was never given a chance to score.
```
Adam-Hammo: Nit: superfluous comment

app/domain/references/services/deduplication_assessment_service.py
```diff
@@ -102,6 +104,28 @@ def _to_scorer_paper(reference: Reference) -> DeduplicationPaper:
...
+            return DeduplicationPairResult(
+                unscorable_reason=UNPROJECTABLE_CANDIDATE_REASON
+            )
+        try:
+            return await self._pair_scorer.score_pair(
+                incoming=incoming, candidate=candidate_paper
+            )
+        except Exception as exc:
+            # Deliberately broad, unlike the narrow retrieval and hydration
+            # handlers: a third-party scorer's defect is one record's, not the run's.
```
Adam-Hammo: Nit: superfluous comment

app/core/exceptions.py
```diff
@@ -666,6 +666,10 @@ def __init__(self, detail: str) -> None:
         super().__init__(detail)
 
 
+class EvaluationConfigurationMismatchError(DestinyRepositoryError):
```
Adam-Hammo: Should this inherit `DeduplicationError`?

app/domain/references/services/deduplication_evaluation_runner.py
```diff
@@ -0,0 +1,584 @@
...
+        blob_repository: BlobRepository,
+        configuration: EvaluationRunConfiguration,
+    ) -> EvaluationRunArtifacts:
+        """Evaluate a JSONL blob and write its run-scoped artifact bundle."""
+        # Two passes over the immutable input: the manifest digest must cover the
+        # exact bytes, which the line reader cannot yield once it has decoded them.
+        digest = hashlib.sha256()
+        byte_size = 0
+        async for chunk in blob_repository.stream_chunks_from_blob_storage(input_file):
+            digest.update(chunk)
```
Adam-Hammo: We already have some similar logic to this inside of `BlobRepository.copy()`, is it worth extracting out into the repository?

app/domain/references/services/deduplication_evaluation_runner.py
```diff
@@ -0,0 +1,584 @@
...
+            f"{outcomes[DeduplicationAssessmentOutcome.PROPOSE_CANONICAL]}\n"
+            "- `propose_duplicate`: "
+            f"{outcomes[DeduplicationAssessmentOutcome.PROPOSE_DUPLICATE]}\n"
+            "- `no_proposal`: "
+            f"{outcomes[DeduplicationAssessmentOutcome.NO_PROPOSAL]}\n\n"
+            "## Pair evidence\n\n"
+            f"- Clears threshold: {clears}\n"
+            f"- Below threshold: {below}\n"
+            f"- Unscorable: {unscorable}\n"
+        )
```
Adam-Hammo, line 613: Would we consider making this a part of the API response model once retrieving the result, instead of a markdown?

app/domain/references/services/deduplication_evaluation_runner.py
```diff
@@ -0,0 +1,584 @@
...
+            assessment=assessment,
+        )
+
+    @staticmethod
+    def _verify_configuration(
+        assessment: DeduplicationAssessment,
+        configuration: EvaluationRunConfiguration,
+        line_number: int,
+    ) -> None:
+        """Fail the run rather than publish a manifest the run contradicts."""
```
Adam-Hammo, line 384: What mechanisms would actually cause this to fire?

app/domain/references/services/deduplication_evaluation_runner.py
```diff
@@ -0,0 +1,584 @@
...
+
+class EvaluationInputReference(BaseModel):
+    """Bibliographic fields frozen into one evaluation row."""
+
+    title: str | None
+    authors: list[str]
+    year: int | None
+
+
+class EvaluationInputRecord(BaseModel):
```
Adam-Hammo, line 63: I presume we plan to make SDK models, anti-corruption layers, etc for these models when we add the routes?

app/domain/references/services/deduplication_evaluation_runner.py
```diff
@@ -0,0 +1,584 @@
...
+                    )
+                    record_results.append(result)
+                    # Checked after the record is kept, so an aborting bundle still
+                    # carries the assessment that disagrees with the configuration.
+                    if result.assessment is not None:
+                        self._verify_configuration(
+                            result.assessment, configuration, line_number
+                        )
+        except BaseException:
+            # A run costs hours of retrieval, so keep whatever completed, a
```
Adam-Hammo: Long-running tasks are a risk for us and I think I'd like to avoid them where we can. It also opens up questions of how a user can poll status etc which might lead to scope creep. Can we propose a sensible cap on the size of the input to keep unit size reasonable? There'd be no problem with them queuing multiple tasks of course.

app/domain/references/services/deduplication_evaluation_runner.py
```diff
@@ -0,0 +1,584 @@
...
+            # cancelling shutdown included. No manifest marks the bundle incomplete.
+            await self._upload(
```
Adam-Hammo, line 211: I think this may want to be wrapped in `asyncio.shield()`

app/domain/references/services/deduplication_evaluation_runner.py
Adam-Hammo: Similarly everything is buffered in memory which would be another reason to cap

app/domain/references/services/deduplication_evaluation_runner.py
Adam-Hammo: Also, in a crash or container exit, I think the task will attempt to rerun, which will just overwrite these again.

app/persistence/blob/repository.py
```diff
@@ -226,6 +226,15 @@ async def stream_file_from_blob_storage(
         client = await self._preload_config(file)
         yield client.stream_file(file)
 
+    async def stream_chunks_from_blob_storage(
```
Adam-Hammo: This possibly deserves its own unit test

Adam-Hammo (commented):
Broader question - how do we intend to expose this feature? The implementation feels task-native, but #915 explicitly denies queuing background work.

PR #921, opened 2026-08-11
Add prod release notifications

### LINKED ISSUE #889: Notify #destiny-ci when a production deploy runs
## Context

Production deploys are manual and silent, so you have to open the Actions tab to find out. We want to be notified when they happen.

Terraform applies already post to `#destiny-ci`, so there is a pattern to follow and possibly a webhook we can reuse.

The same work is tracked for the other two deployed services, and whoever picks one up should probably do all three. Start here; this repo has the most moving parts.

- https://github.com/destiny-evidence/evidence-repo-ui/issues/224
- https://github.com/destiny-evidence/taxonomy-builder/issues/259

## Agreed at tech-debt prioritisation

- All three services post to `#destiny-ci`, giving one deploy timeline across the apps. If we ever split them, split by environment rather than by application.
- Production only — staging and development deploy far too often to be worth announcing.
- The robots and the AI evidence summariser are excluded; they deploy too rarely to be worth it.

## Acceptance criteria

- [ ] One message per production deploy in `#destiny-ci`, giving the service, the image tag, who triggered it, and a link to the run.
- [ ] Failed deploys notify too, and read unmistakably as failures.
- [ ] Rollbacks notify — `rollback-production.yml` puts a different image into production.
- [ ] Development and staging deploys stay silent.
- [ ] The webhook URL lives in a secret.

Adam-Hammo:
Closes #889

Delivers notifications to slack on successful/failed/cancelled production deploys.

Example messages are in `#testiny`.

The calling trigger (eg promote or rollback) is in the footer of the message.

### DIFF
--- a/.github/workflows/promote-to-production.yml
+++ b/.github/workflows/promote-to-production.yml
@@ -32,6 +32,7 @@ jobs:
     if: github.ref == 'refs/heads/main' # Only run on main, not strictly necessary but for safety
     uses: ./.github/workflows/deploy.yml
     needs: get-image-tag
+    secrets: inherit
     with:
       environment: production
       image_tag: ${{ needs.get-image-tag.outputs.image_tag }}
--- a/.github/workflows/deploy-to-development.yml
+++ b/.github/workflows/deploy-to-development.yml
@@ -22,7 +22,6 @@ jobs:
   deploy:
     needs: build
     uses: ./.github/workflows/deploy.yml
-    secrets: inherit
     with:
       environment: development
       image_tag: ${{ needs.build.outputs.image_tag }}
--- a/.github/workflows/deploy-to-development.yml
+++ b/.github/workflows/deploy-to-development.yml
@@ -22,6 +22,7 @@ jobs:
   deploy:
     needs: build
     uses: ./.github/workflows/deploy.yml
+    secrets: inherit
     with:
       environment: development
       image_tag: ${{ needs.build.outputs.image_tag }}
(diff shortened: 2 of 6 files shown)


.github/workflows/promote-to-production.yml
```diff
@@ -32,6 +32,7 @@ jobs:
     if: github.ref == 'refs/heads/main' # Only run on main, not strictly necessary but for safety
     uses: ./.github/workflows/deploy.yml
     needs: get-image-tag
+    secrets: inherit
```
Adam-Hammo, line 35: Shoutout https://github.com/actions/runner/issues/4453

jack-o-the-green (commented):
Do we also want to be able to give deployment failed notifications if the replicas of any of the container apps fail to start, or do we feel the restart alerts here https://github.com/destiny-evidence/destiny-repository/blob/main/infra/app/alerts.tf#L314 give enough coverage. 

iirc the deployment flow is green even if the replicas don't start up, because it's delivered all the changes.

Adam-Hammo:
Good call-out. I think the only way to confidently alert on this is a bit of `az api` polling in the github action, will investigate that.

jack-o-the-green:
> Good call-out. I think the only way to confidently alert on this is a bit of `az api` polling in the github action, will investigate that.

Also happy to approve this one with the investigation as a potential follow up, this is still better than what we have now!

Adam-Hammo:
@jack-o-the-green ready for review again. I considered an azure `alerts.tf` approach but it was hard (perhaps impossible) for those alerts to be sure what to expect - you can't alert on nothing happening! So unfortunately we're in a polling landscape now.

jack-o-the-green (approved):
LGTM - might need to keep an eye on how much of our github actions usage the resource wait chews. Thanks for updating!

PR #924, opened 2026-08-14
Authenticate users with client roles

jack-o-the-green:
### Context

Cut over the repository to per-env authentication #67 

Will want to deploy this to development and test once https://github.com/destiny-evidence/destiny-shared-infra/pull/69 has landed

### What's changed

- Access is no longer gated on realm roles, but on client roles which are assigned per-environment. 
- Local and test keycloak realms use client roles 
- Procedure documentation updated 
- Scope requests removed from SDK and SDK auth supports client role auth
    - This is a breaking change for clients, we will need to update the bulk-abstract-ingest-robot as it is the only one using the keycloak flow. 

### Testing 

Unit tests/e2e tests updated

Tried this out locally with the client roles deployed from https://github.com/destiny-evidence/destiny-shared-infra/pull/69 after assigning myself to the groups. Also updated the keycloak realms to hold the new way of doing things for local testing. 

Deployed this branch to development after merging and applying https://github.com/destiny-evidence/destiny-shared-infra/pull/69 under the following scenarios 

* User with no `consumers-development` or `developers-development` group membership 
    * User was unable to use https://data.dev.evidence-repository.org/hpv to execute a search against dev destiny repository, receiving a 403 forbidden [expected] despite having realm roles which previously conveyed reference reader access. 
* User with `consumers-development` group membership (allows reference reads) 
    * User was able to use https://data.dev.evidence-repository.org/hpv to execute a search against dev destiny-repository
    * User was unable to get the status of an import batch [expected] despite having realm roles which previously conveyed import writer access. 
* User with `developers-development` but no `consumers-development` group membership (allows import writer and reference reads through the developers-development group)
    * User was able to use https://data.dev.evidence-repository.org/hpv to execute a search against dev destiny-repository
    * User was able to get the status of an import batch from dev destiny repository. 
* User with `developers-development` but no `developers` (old group) group membership was able to authenticate against dev-destiny repository with expected access.

Realm role access for AI summaries and vocabulary builder/reader behavior is unchanged.  
    
### What must we do when this is merged and deployed?

- Release the new SDK version and update https://github.com/destiny-evidence/bulk-abstract-ingest-robot/blob/eb2aa53fc9b37fd56c032d648f8af779561f3327/app/util/repository.py#L7
- Clean up the old realm-roles from destiny shared infra

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ To get a token for use in a development environment, there is a utility module:
 uv run python -m app.utils.get_token
 ```
 
-### Keycloak Authentication (Experimental)
+### Keycloak Authentication
 
 Keycloak support is experimental as we transition from Azure AD to Keycloak. It is controlled by the `AUTH_PROVIDER` setting in `.env`:
 
@@ -205,8 +205,12 @@ Keycloak support is experimental as we transition from Azure AD to Keycloak. It 
 AUTH_PROVIDER=keycloak
 KEYCLOAK_URL=http://localhost:8080
 KEYCLOAK_REALM=destiny
-KEYCLOAK_CLIENT_ID=destiny-repository-client
+KEYCLOAK_CLIENT_ID=destiny-repository-client-local
 ```
+
+The client is named after the environment, and authorization is the set of client
+roles the principal holds on it (`resource_access["destiny-repository-client-local"]`
+in local development). Realm roles and the `scope` claim confer no access.
 
 When `ENV=local` (the default in `.env.example`), authentication is bypassed entirely regardless of the provider. To test with Keycloak auth enforced locally, start Keycloak and run the app via docker compose:
 
--- a/README.md
+++ b/README.md
@@ -212,13 +212,21 @@ The authorization is the set of client
 roles the principal holds on it (`resource_access["destiny-repository-client-local"]`
 in local development). Realm roles and the `scope` claim confer no access.
 
-When `ENV=local` (the default in `.env.example`), authentication is bypassed entirely regardless of the provider. To test with Keycloak auth enforced locally, start Keycloak and run the app via docker compose:
+When `ENV=local` (the default in `.env.example`), authentication is bypassed entirely regardless of the provider. To test with Keycloak auth enforced locally, set `BYPASS_AUTH=false` and configure the internal Keycloak URL (`http://keycloak:8080`) for token validation, while using `http://localhost:8080` as the issuer URL to match tokens obtained from the browser.
+
+```dotenv
+KEYCLOAK_URL=http://keycloak:8080
+KEYCLOAK_REALM=destiny
+KEYCLOAK_CLIENT_ID=destiny-repository-client-local
+KEYCLOAK_ISSUER_URL=http://localhost:8080
+BYPASS_AUTH=false
+```
+
+Then start Keycloak and run the app via docker compose:
 
 ```sh
 docker compose --profile keycloak --profile app up
 ```
-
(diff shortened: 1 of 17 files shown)


docs/procedures/oauth.rst
```diff
@@ -164,4 +166,6 @@ The tokens will expire after a certain period (usually two hours). After expirat
 Troubleshooting
 ---------------
 
+A ``401`` means the token was rejected - most often expired, or obtained for a different environment than the API you called. A ``403`` means you lack the permission that resource requires.
```
Adam-Hammo, line 169: Nice 👌

app/api/auth.py
```diff
@@ -477,7 +480,7 @@ def auth_strategy():
                     keycloak_url="http://localhost:8080",
                     realm="destiny",
                     client_id="destiny-repository-client",
```
Adam-Hammo:
```suggestion
                    client_id="destiny-repository-client-staging",
```

libs/sdk/src/destiny_sdk/keycloak_auth.py
```diff
@@ -109,22 +101,23 @@ class KeycloakAuthCodeFlow:
...
-    Uses the same `destiny-auth-client` as all other human users.
-    Scopes are determined by the user's group membership (consumers vs developers).
+    Authorization is the set of client roles the user's group membership grants;
+    there is nothing to request.
 
     Example usage:
         flow = KeycloakAuthCodeFlow(
             keycloak_url="http://localhost:8080",
             realm="destiny",
+            client_id="destiny-auth-client",
```
Adam-Hammo:
```suggestion
            keycloak_url="https://auth.evidence-repository.org",
            realm="destiny",
            client_id="destiny-auth-client-staging",
```

README.md
```diff
@@ -197,17 +197,21 @@ To get a token for use in a development environment, there is a utility module:
 uv run python -m app.utils.get_token
 ```
 
-### Keycloak Authentication (Experimental)
+### Keycloak Authentication
 
 Keycloak support is experimental as we transition from Azure AD to Keycloak. It is controlled by the `AUTH_PROVIDER` setting in `.env`:
```
Adam-Hammo:
```suggestion
Keycloak support is controlled by the `AUTH_PROVIDER` setting in `.env`:
```

Adam-Hammo: (This README needs a wider refresh methinks!)

Adam-Hammo (approved):
Looks good, just docs nits!

We should maybe also default `ui/lib/authConfig.ts` to `destiny-auth-client-local` or similar (or remove the default entirely to make the failure mode more obvious)

app/api/auth.py
jack-o-the-green: Pedantically, the client_ids are defined in destiny-shared-infra and could be anything, so I don't think this example should reflect the name of an actual client id. I could make it the following instead.

```
destiny-repository-client-id
```

README.md
jack-o-the-green: 100%

libs/sdk/src/destiny_sdk/keycloak_auth.py
jack-o-the-green: Same as above, I'm unsure we want actual values in the example docs. Be happy to do a 

```
keycloak_url="https://keycloak-deployment.org",
realm="destiny",
client_id="destiny-auth-client-id",
```

Adam-Hammo: Nice, happy with both of those

Issue #922, opened 2026-08-12
Dark launch candidate retrieval against production ingestion


jaybea:
## Outcome

Run candidate retrieval alongside production ingestion, making no changes to references relationships (duplicate, canonical). Determine whether it keeps pace with steady-state ingestion without disrupting imports, and also candidate retrieval statistics on real arrivals. Report latency, memory, throughput, candidate yield and failures, without changing decisions or load above normal ingestion.

Shadow retrieval can be built and run before the Deduper adapter and scoring telemetry land. It is where the cost and the disruption risk sit, so measure it on its own first.

## Context

- **Prerequisite:** deploy #937 before enabling shadow retrieval. Verify its Honeycomb measurements are being collected on initial shadow traffic.
- The shortcut returns before candidate retrieval, and makes the real decision, which on the steady-state feed will always be canonical: OpenAlex W-IDs are globally unique, so the shortcut finds no candidates and writes no side-effect decisions. Shadow retrieval only observes. Observe the corpus in the state deep deduplication will see, which means before this task has written anything. On the steady-state feed that is equivalent either way, since the only row the shortcut touches is the incoming reference and it is excluded from its own candidate search. It matters on re-imports and backfills, where the shortcut does find siblings and collapses them onto one canonical, changing what the identifier arm returns.
- The normal feed was recently about ~70k works/day, but 8th August queued 750k. For initial tests, run during steady state, not catch-up.
- This gives us a feel for normal queueing and autoscaling. It is not a controlled capacity test or concurrency sweep.
- We expect Elasticsearch to dominate per-record cost. Pair scoring and hydration are likely under 1% of retrieval time even at k=100. So this arm answers most of whether deduplication keeps pace, and the scoring arm adds little to the duration measured here.

## Scope

- #937 implements candidate-search telemetry. Add worker memory and in-flight concurrency in Honeycomb. Memory is a peak reading, a high-water mark for the process rather than an instantaneous one, so it answers how close a worker came to its limit and cannot show memory being reclaimed. Both readings are per worker **process**, and a replica runs more than one.
- Piggy-back off normal ingestion.
[...]

Issue #877, opened 2026-07-30
Confirm the first automatic deduplication policy for OpenAlex


jaybea:
Part of #791. Needs the evaluation run in #887, the Deduper evaluation pairs from #822 and the retrieval evaluation set from #823.

## Outcome

Name the policy that may merge new OpenAlex references automatically, and which references are eligible, on published evidence.

## The policy must state

The acceptable false-merge rate for the first automatic policy. Precision and false-merge evidence per score band, with counts, reported separately by adjudication method and by population rather than pooled. The Deduper version and configuration, the retrieval policy and the number of candidates it returns, the score rule, and the eligible inputs.

Whether identifier-route matches need corroboration beyond identifier equality, or a threshold of their own. #808 holds the constraints that bear on this.

How missed matches were measured: labels, later duplicate discovery, or an audited sample. A below-threshold count is not a missed-match rate.

Scores are not comparable across records with different field availability, because an absent field behaves like a contradictory one. A single threshold covering sparse and complete records is not safe, and the policy has to say how it handles that.

## Acting policy

Apply this rule to the candidates that clear the Deduper threshold:

| Candidates clearing the threshold | Incoming reference | Result |
|---|---|---|
| 0 | Any eligible reference | Write `CANONICAL` |
| 1 | Not community-bound | Write `DUPLICATE` of that candidate |
| 1 | Community-bound | Retain the assessment for review; write no duplicate decision |
| 2 or more | Any eligible reference | Retain the assessment with all threshold-clearing candidates for review; write no duplicate decision |

The candidate's community membership does not block a merge. A new OpenAlex reference may become a duplicate of an existing community canonical because that does not change the canonical's decision. Do not exclude community canonicals from the candidate union defined in #876, and do not search behind them for a different target. Candidate community membership does not alter the #874 assessment.

For the first policy, the incoming reference is community-bound when its current annotations contain one of a versioned, configured list of scheme/label pairs. This is the temporary definition until #658 provides a community registry. Validate the list against production before turn-on: it must identify the existing ESEA and HPV populations, including references carrying linked-data enhancements. Record the list version with the policy generation.

[...]

PR #927, opened 2026-08-17
Record the evidence behind each deduplication assessment

jaybea:
## Summary

Persist the evidence behind a deduplication assessment, so an automatic decision can
later explain itself without re-running retrieval. The proposal is recorded, never
applied: no `ReferenceDuplicateDecision` is written, so no reference's duplicate
determination changes.

A new `deduplication_assessment` table holds one row per assessment:

- identity and cohort: the delivery that produced it, the incoming reference, the
  purpose and the policy generation;
- retrieval provenance: the policy name, configured `k`, actual candidate count,
  whether the Elasticsearch route ran, the concrete index when it did, and why the
  input did or did not meet the searchability gate;
- scorer provenance: version, configuration hash and threshold;
- outcome: the proposal, the best score, the best score that did not win, and a
  bounded per-candidate summary as JSONB.

Compared values and per-field scores stay out of the row. They go to the operations
blob container.

## Retention

Every assessment gets a summary row. The full payload is kept when a later reader
would need it, meaning it proposed a duplicate or failed to score a candidate, plus a
configurable share of the rest. The rate is a base-2 exponent: 0 keeps everything,
null keeps none. It defaults to keeping everything, because the initial probe runs on
full provenance. We expect to narrow it to a sample longer term, once the question
becomes monitoring drift rather than measuring the run. Each environment sets its own.

The sample is keyed on the reference alone, so every policy generation keeps evidence
for the same references and two generations can be compared like for like.

## Boundaries

Nothing in the application calls this yet. The recorder, its blob writer and its
configuration are reached only by tests and local verification. The first production
caller is the measurement run in #922.

It records proposals but cannot apply one. It has no decision-writer dependency and
does not touch `reference_duplicate_decision`, Elasticsearch or background tasks, so
a recorded `propose_duplicate` leaves the reference exactly as it was.

#875 stays open after this merges. Applying an assessment as a decision, the
acting-policy boundary from #877 and the automatic-decision flag are not here.

## Implementation notes

**The payload is written before the row, and named for the delivery.** `add()` only
flushes, so a row written first could not survive a rollback of its blob, and the
compensating update that would have corrected it afterwards cannot be relied on.
Naming the object for the caller's idempotency key rather than the record id means a
retry after a rolled-back insert overwrites one object instead of stranding another,
since the find-first guard only sees committed rows.

**`idempotency_key` is an attempt identity, caller-supplied and required.** A semantic
key of reference, generation and purpose was rejected because it would block a
legitimate reassessment of the same reference as the corpus moves beneath it. It is
required rather than defaulted so the first caller has to decide where it comes from.

**Field status records that a field was compared, not whether the values agreed.** The
scorer returns a similarity rather than a verdict, and there is no honest cut to make
one: journal scores sit between 0.3 and 0.8 on 94% of duplicate pairs in the OpenAlex
corpus we score. The score and the normalised values carry that evidence instead.

**`incoming_reference_id` is not a foreign key.** Supplied records under evaluation
need not be held.

**A check constraint holds the payload columns coherent.** The generic `update_by_pk`
writes attributes without model validation, so the database has to hold that
invariant rather than the model.

## Validation performed

- Live verification against PostgreSQL, Elasticsearch and MinIO, 15 checks: an
  assessment proposing a duplicate keeping its evidence; a redelivery returning the
  original record with no second upload; a fresh delivery recording the same
  reference again; an uninteresting assessment keeping no payload; an unsearchable
  input recording why and running no route; an unscorable pair; and a delivery whose
  insert was rolled back through a savepoint, whose retry wrote back over the same
  object rather than stranding one.
- Blob storage counted either side of one verification run: four new objects for the
  four records that retained a payload, all named for a delivery, none unreferenced.
- Single migration head, round-tripped down and up against a local database.
- Unit and integration 1,197 passed with 1 skipped, e2e 29 passed, all pre-commit
  hooks clean.

Relates to #875. Builds on the assessment contract from #874.


tests/unit/domain/references/deduplication/test_assessment_record.py
Adam-Hammo, line 1: This is essentially 835 lines of testing for a 60 line method - is there some consolidation we could do here?

app/domain/references/services/deduplication_assessment_recorder.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+
+    async def record(
+        self,
+        assessment: DeduplicationAssessment,
+        *,
+        purpose: DeduplicationAssessmentPurpose,
+        policy_generation: str,
+        idempotency_key: UUID,
+    ) -> DeduplicationAssessmentRecord:
```
Adam-Hammo, line 176: Is this exposed to TOCTOU like how we handle it in `tasks.process_reference_duplicate_decision`?

app/domain/references/models/sql.py
```diff
@@ -778,6 +788,157 @@ def to_domain(
...
+            "policy_generation",
+        ),
+        Index(
+            "ix_deduplication_assessment_incoming_reference_id",
+            "incoming_reference_id",
+        ),
+        Index(
+            "ix_deduplication_assessment_created_at",
+            "created_at",
+        ),
```
Adam-Hammo: Is this made redundant by the UUID7 PK?

app/domain/references/models/sql.py
```diff
@@ -778,6 +788,157 @@ def to_domain(
...
+        Boolean, nullable=False, default=False
+    )
+    payload_bytes: Mapped[int | None] = mapped_column(Integer, nullable=True)
+
+    __table_args__ = (
+        # For listing the population an acting policy generation produced.
+        Index(
+            "ix_deduplication_assessment_policy_generation",
+            "policy_generation",
+        ),
```
Adam-Hammo: I fear the cardinality of this could make the index pretty inconsequential. Most/all rows will share the same `policy_generation` so this would be little better than a sequential scan.

Adam-Hammo: Maybe a composite key by adding `outcome` or similar could help give this use? What sort of query shape do you anticipate?

app/domain/references/models/models.py
```diff
@@ -1226,6 +1234,149 @@ def validate_not_empty(self) -> Self:
         return self
 
 
+class FieldAvailability(StrEnum):
```
Adam-Hammo: This feels prone to drift against `DeduplicationFieldStatus`. Could we instead invert by placing a validator on `AssessmentCandidateSummary` that just asserts `field_availability != DeduplicationFieldStatus.compared`?

app/domain/references/models/sql.py
```diff
@@ -778,6 +788,157 @@ def to_domain(
...
+    )
+    payload_state: Mapped[AssessmentPayloadState] = mapped_column(
+        String, nullable=False
+    )
+    payload_blob_url: Mapped[str | None] = mapped_column(String, nullable=True)
+    payload_reason: Mapped[str | None] = mapped_column(String, nullable=True)
+    payload_sampled: Mapped[bool] = mapped_column(
+        Boolean, nullable=False, default=False
+    )
+    payload_bytes: Mapped[int | None] = mapped_column(Integer, nullable=True)
```
Adam-Hammo: This looks like it'll come out to 40-50Kib per row. If we import a single 1mio reference block from OpenAlex, that's 40-50GiB, before indices etc. We of course have room for this, but we'll need to keep a close eye on it.

app/domain/references/models/sql.py
```diff
@@ -778,6 +788,157 @@ def to_domain(
...
+    )
+    best_score: Mapped[float | None] = mapped_column(Float, nullable=True)
+    best_non_winning_score: Mapped[float | None] = mapped_column(Float, nullable=True)
+    scored_candidates: Mapped[list[dict[str, Any]]] = mapped_column(
+        JSONB, nullable=False
+    )
+    payload_state: Mapped[AssessmentPayloadState] = mapped_column(
+        String, nullable=False
+    )
+    payload_blob_url: Mapped[str | None] = mapped_column(String, nullable=True)
```
Adam-Hammo, line 835: Tying into our conversation offline - it feels slightly off having two different blob files across #919 and this PR.

Adam-Hammo (commented):
Added a few notes, looking good.

This made me realise we don't have much telemetry on the dedup process - might be something to keep in mind.

tests/unit/domain/references/deduplication/test_assessment_record.py
jaybea: Agreed. I've consolidated from 30 tests / 835 lines to 12 tests / 510 lines. Moved some tests to config validation rather than at use. Tried to leave the remaining tests focus being the recorder's public behaviour.

app/domain/references/services/deduplication_assessment_recorder.py
jaybea: Great pointer. This should be uncommon, but I added insert-or-find using a savepoint.

app/domain/references/models/sql.py
jaybea: Yes it does! temporal UUID7 has so many uses! And there's no query currently using this index so I chopped it. Can add later if a retention or timestamp query needs it actual created at time.

app/domain/references/models/sql.py
jaybea: These were premature, I’ve removed the speculative index and retained only the unique idempotency key and the incoming_reference_id index, which support current behaviour.

app/domain/references/models/models.py
jaybea: Good point, I changed to using the `DeduplicationFieldStatus` directly.

app/domain/references/models/sql.py
jaybea: Maybe we can reconcile, but it feels like they fill 2 different needs. This one stores sampled assessment evidence, and hopefully helps us monitor and analyse for improvement possibilities longer term. #919 just does a point in time artifact. But maybe we can factor out common behaviours somehow.

jaybea:
On the telemetry, your spidey-sense is spot on. in `deduplication_service.py`, almost 1000 lines of code in 17 methods, 2 log calls and 0 trace attributes. Independent of the assessment recording we need to instrument "is live deduplication behaving right now?" I'll file an issue and I'd really appreciate your feedback.

app/domain/references/models/sql.py
Adam-Hammo: Fair enough!

tests/integration/test_deduplication_assessment_repository.py
```diff
@@ -266,6 +267,24 @@ async def test_one_attempt_cannot_write_two_records(
         await repository.add(second)
 
 
+async def test_idempotent_add_returns_the_existing_record(
```
Adam-Hammo, line 302: Could we add a test with two sessions that tests the concurrency behaviour?

app/domain/references/repository.py
```diff
@@ -1258,6 +1259,36 @@ def __init__(self, session: AsyncSession) -> None:
...
+        self, record: DomainDeduplicationAssessmentRecord
+    ) -> DomainDeduplicationAssessmentRecord:
+        """Insert a record, or return the row written by a racing delivery."""
+        trace_attribute(Attributes.DB_PK, str(record.id))
+        self.trace_domain_object_id(record)
+        persistence = self._persistence_cls.from_domain(record)
+        try:
+            # A savepoint, so a losing insert does not abort the enclosing unit of
+            # work and leaves the session usable for the lookup below.
+            async with self._session.begin_nested():
```
Adam-Hammo, line 1273: This is a nice pattern!

tests/integration/test_deduplication_assessment_repository.py
Adam-Hammo:
```
Session A inserts key K, flushes, does not commit.
Session B inserts key K concurrently and blocks on the unique index.
A commits.
B's insert raises, B runs find.
```

app/domain/references/services/deduplication_assessment_recorder.py
```diff
@@ -199,7 +197,7 @@ async def record(
             else _PayloadOutcome(AssessmentPayloadState.NOT_RETAINED)
         )
 
-        return await self._record_store.add(
+        return await self._record_store.add_or_find_by_idempotency_key(
```
Adam-Hammo, line 200: I've realised we have the same gap on the blob files - I'm willing to overlook it given the filename ensures there's only one, and a racey call will be identical "almost always"™️.

I guess the solution if you want it is to do the postgres row first, detect if inserted, and use that to decide whether to write the blob.

Adam-Hammo (commented):
Looking good, a couple more things then this should be sweet!

jaybea:
FYI, telemetry issues added: #937 and #938.

tests/integration/test_deduplication_assessment_repository.py
jaybea: Added it, should have added it before, that was cool to really prove it.

Was tricky though, it polls a fresh connection per attempt, and matches the waiting query to the insert, so the test cannot pass by commiting before the 2nd one has even started.

app/domain/references/services/deduplication_assessment_recorder.py
jaybea: Yeah, I think it’s going to be pretty rare unless I’m missing something, and the repercussions seem limited:

1. Executions with the same idempotency key must overlap. That’s possible, and concurrent redelivery has previously occurred with dedup tasks.
2. Both must pass the recorder’s initial lookup before the first transaction commits. Assuming recording is immediately followed by the unit-of-work commit, that should be a fairly short interval.
3. For the overwrite to matter, the assessments must differ materially. Per-run diagnostics such as Elasticsearch timing may differ routinely, but an inconsistent summary and payload requires the live database or Elasticsearch results to change the candidates, scores, outcome, or other recorded evidence.

Even then, this affects assessment evidence rather than production decision state. Yes, I'd like to leave it as-is.

Adam-Hammo: Agreed.

Adam-Hammo (approved):
Looks good!

PR #939, opened 2026-08-20
Add managed identity auth for research mapper

Adam-Hammo:
Does what it says on the tin

Partner PR to https://github.com/destiny-evidence/research-mapper/pull/27

### DIFF
--- a/infra/app/auth.tf
+++ b/infra/app/auth.tf
@@ -175,3 +175,17 @@ resource "azuread_app_role_assignment" "ai_evidence_summariser_to_reference_full
   principal_object_id = data.azurerm_user_assigned_identity.ai_evidence_summariser[0].principal_id
   resource_object_id  = azuread_service_principal.destiny_repository.object_id
 }
+
+# Agentic mapping (research-mapper) role assignments
+data "azurerm_user_assigned_identity" "research_mapper" {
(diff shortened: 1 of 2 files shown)


infra/app/auth.tf
```diff
@@ -175,3 +175,17 @@ resource "azuread_app_role_assignment" "ai_evidence_summariser_to_reference_full
   principal_object_id = data.azurerm_user_assigned_identity.ai_evidence_summariser[0].principal_id
   resource_object_id  = azuread_service_principal.destiny_repository.object_id
 }
+
+# Agentic mapping (research-mapper) role assignments
+data "azurerm_user_assigned_identity" "research_mapper" {
+  count               = var.research_mapper_app_name != null ? 1 : 0
+  name                = "${var.research_mapper_app_name}-${var.environment}"
```
jack-o-the-green: `nitpick` if we're not defaulting the name of the research_mapper_app_name to something in the variable that we want to automatically use in most envs, we're not really getting anything out of assuming the `-${var.environment}` suffix on the name. And the cost of having the suffix on there is that we are restricted on naming the research mapper resources in alignment with the deployed destiny environments. 

Prefer just passing in the full name?

jack-o-the-green (approved):
LGTM - only a nitpick

